PR #13, opened 2019-10-07
Feature/#5 monorepo code port

foxleigh81:
## 🧰 Ticket
#6 

## 🚀 Overview: 
Moved codebase from serge repo to here and got it working with the monorepo archictecture

## 🤔 Reason: 
Not much use in having a monorepo if there's no code in it!

## 🔨Work carried out:

- [x] imported `client` code
- [x] imported `server` code
- [x] Added some config files to the root of the repo
- [x] Installed and configured Lerna
- [x] Moved the styles from the client folder to their own `themes` package
- [x] Added documentation

## 📝 Developer Notes:
- Lerna (https://github.com/lerna/lerna) is a monorepo management tool. For now it will just sit there in the background but as the monorepo grows more complex it will make sense to use some of its commands for things like releases
- This is a very bare-bones start to the monorepo approach and it will evolve a lot as thie proejct scales and when I carry out the converting components to modular and especially when the microservices architecture is added. This is why I split the styles out into their own package so you can get an idea of how it will work, down the line.


IanMayo (approved):
Great work, thanks @foxleigh81

Issue #3, opened 2019-10-01
Introduce E2E testing


IanMayo:
## User story

As a _maintainer_ I wish to be able to _run e2e testing_ to _avoid the introduction of regression tests while adding new code or refactoring_

## Overview

The inherent complexity of software means it's quite frequent that a change in one subsystem can affect apparently unrelated code, either in that sub-system or another. 

Frequently, the only way to be sure an application isn't broken is to perform a full manual run-through, testing as many execution branches as possible.  This is inherently time consuming, so may only be performed infrequently.

Automated e2e testing is able to perform this manual run-through without user intervention.  It can even perform it headless within a CI pipeline.

As new features are added, they can be covered by new e2e tests, when necessary.

## Strategy

(see below in Task List)

## MVP test
* Create wargame
* Name as `e2e_test_one`
* Set passwords to visible
* Create Blue force
* Create channel titled `All chat`, Blue and White can send `chat` messages
* Create player, login as White
* Create player, login as Blue
* Send message as white
* Verify message shown

## Wider testing
* more forces
* more roles
* more channels, including specific message types
* test correct channels displayed
* test sending formatted messages
* Test open/close messages
* Test marking all messages as read
* test sending Game Admin messages
* test sending feedback
* test turn time warnings
* test Control making turn steps
* test leaving/re-joining game
* test exporting messages

IanMayo:
Note: @Heri has already started this work.

IanMayo:
@heridehero is working on this issue, and has produced a first set of testing.

Once we've got Serge running in the new repo, we should _tie in_ the e2e testing for `yarn`, so we can run the tests more easily.

We should also have some peer review of the e2e implementation, for us as a team to agree on the strategy.

Then,  it seems a good idea for @heridehero to keep this as a _background task_, to steadily address technical debt when there are no other tasks pending.

zukilover:
well noted.

IanMayo:
@foxleigh81 is working on integrating the e2e testing into the build

IanMayo:
Ok. We're going to defer this issue, until @foxleigh81 has a larger budget of time for the project, and can provide a model e2e test case, built into our env, that we can build upon.

IanMayo:
Initial solution provided by @toan301

IanMayo:
Initial solution is satisfactory.

PR #39, opened 2019-10-28
Feature/unit tests

foxleigh81:
## 🧰 Ticket
continues #3 

## 🚀 Overview: 
Moves the existing e2e tests out of the `client` package and into a temp folder in the `e2e` package ready for future reintegration. Also re-establishes tests (limited to unit tests right now) in the `client` package.

## 🤔 Reason: 
`react-scripts` isn't really well suited to running e2e tests, also e2e tests should really run independently of the `client` or `server` applications. So it makes sense to run them as their own package.

## 🔨Work carried out:

- [x] Added a `serge` script so clients can start the server with `yarn serge`
- [x] Moved e2e tests to `e2e` package
- [x] Updated package.json to run the unit tests from the root of the monorepo

## 📝 Developer Notes:
- The unit tests are currently not passing, this appears to be because of a recent change in the shape of the data but it could also be beacuse I perhaps used the wrong `REACT_APP_SERVER_PATH` in `setupTests.js`. Either way, this needs to be investigated and amended.


IanMayo (approved):
Quite happy.  Will try to get the broken ones fixes ASAP.

PR #42, opened 2019-10-29
Fix demo password list overflow

### LINKED ISSUE #35: List of roles can get too tall for Player Screen
For a wargame, game control has created lots of forces, and lots of roles in those forces.

At the player welcome screen, eventually the list on the left-hand side becomes taller than the screen.  But, no scroll bars are offered - making it impossible to select some roles.

Either the whole screen should have a scroll bar, or the left-hand column of forces/roles should have a scroll bar when necessary.

I prefer the second option, since it keeps the login/entry-code in the centre of the screen.

zukilover:
## 🧰 Ticket
Closes #35 

## 🚀 Overview: 
Update styling for demo password list section

## 🤔 Reason: 
For a wargame, game control has created lots of forces, and lots of roles in those forces.

At the player welcome screen, eventually the list on the left-hand side becomes taller than the screen. But, no scroll bars are offered - making it impossible to select some roles.

## 🔨Work carried out:

- [x] Add new class selector
- [x] Add new CSS rules
- [x] Tests pass
[List of roles can get too tall for Player Screen](https://app.gitkraken.com/glo/board/XZIuhoko5QAPaF0f/card/XbL5EtnTpQAP1k-J)

### DIFF
--- a/packages/client/src/Views/PlayerUiLobby.jsx
+++ b/packages/client/src/Views/PlayerUiLobby.jsx
@@ -53,7 +53,7 @@ export default function PlayerUiLobby ({ wargameList, roleOptions, checkPassword
           {selectedWargame && state.showAccessCodes &&
           <div className="demo-passwords">
             <h3>Not visible in production</h3>
-            <ul>
+            <ul className="demo-list-forces">
               {roleOptions.map((force) => {
                 return (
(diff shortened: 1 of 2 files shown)


IanMayo (approved):
Yes, working perfectly:
image

PR #44, opened 2019-10-30
serge as executable

### LINKED ISSUE #16: Updates to support "Serge Executable"
The structure of the repo has changed, as we moved to a monorepo.

We should check if/how "Serge as executable" works.  We're no longer using `npm`, so the script will probably need some re-working.

lilitkarapetyan:
## 🧰 Ticket
`closes #16`

## 🚀 Overview: 
Serge as executable reworked.
Added .eslintrc and updated syntax.
Updated .gitignore file.


packages/executable/.eslintrc
```diff
@@ -0,0 +1,4 @@
+{
```
IanMayo, line 1: @lilitkarapetyan - @foxleigh81 has taken responsibility for build & linting. Can you explain why this lint resource file has been introduced?

thanks :-)

IanMayo:
@foxleigh81  - I've added you as reviewer, since this relates to build.

packages/executable/.eslintrc
foxleigh81: It looks like it's just a copy of one of my eslint files, either way, it's extending the stuff I've already established so I'm fine with this. Each package should have its own linting at the moment.

packages/executable/helpers/pkgResolveErrors.js
```diff
@@ -0,0 +1,15 @@
...
+  C:\Users\User\Documents\serge\node_modules\pouchdb-fauxton\www\index.html
+*/
+
+const file = path.resolve(__dirname, '../../../node_modules/pouchdb-fauxton/www/index.html')
+console.log(`error correction [pouchdb-fauxton]: ${file}`)
+fs.writeFile(file, '', 'utf8', err => {
+  if (err) return console.log(err)
+  else console.log('error correction [pouchdb-fauxton]: success')
+})
+console.log('')
```
foxleigh81, line 15: What is this blank console log for?

packages/executable/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "executable",
+  "version": "1.0.0",
```
foxleigh81: We are going to try to keep our package versions in-line for the time being, please make this `0.2.0`

IanMayo:
@lilitkarapetyan - this will be a great way of delivering the application.

But, I think we need some "fiddling".  I'm concerned that the IP-Address the app is running from may not match the one in `.env`.   Once we've done the build, we can't edit the IP-Address.

We need a downloaded copy of `serge.exe` to be able to run out of the box.

Having the IP-Address as `localhost` won't work, since we have to use the IP-Address for other players to connect to.

I _think_ the above situation could be resolved by the client connecting to the full URL/IP-Address of the server.  No, I don't know how we communicate this to the client.

For devs, they will still probably want to be able to run the client and server separately (`yarn start:client` and `yarn start:server`).

packages/executable/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "main": "index.js",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "find": "^0.3.0",
+    "pkg": "^4.4.0"
+  },
+  "bin": "../server/bin.js",
+  "pkg": {
+    "assets": [
+      "../client/build/**/*",
```
foxleigh81, line 14: @IanMayo We can leave this in for now but I'd rather us not reference other packages in the monorepo as relative paths. I had to do the same thing in the `server` package so like I say, it's fine for now but we should add refactoring these references out to our tech debt.

foxleigh81 (changes requested):
Great use case for a package. Nicely done :) 

One change request (version number in package.json), otherwise I'm happy to approve this.

.gitignore
```diff
@@ -62,4 +62,14 @@ typings/
 
 # Generated files and folders
 packages/client/build/**/*
-log.txt
\ No newline at end of file
+log.txt
+
+# Ignore local and special envs
+
+# loaded in all cases
```
IanMayo, line 69: @lilitkarapetyan - if we're ignoring these environment files, they're not present in the repo. That means they're not available to developers, to get them started.

Or, am I missing something?

packages/executable/helpers/pkgResolveErrors.js
lilitkarapetyan: Just for break

.gitignore
lilitkarapetyan: All default settings must be saved in the ".env" file and must be overridden by special envs (.env.local, .env.development.local ...)

For example, for development, we need to create .env.development and put all the dev configs inside, and if some developer wants to change the settings only for his computer, he can add the .env.development.local (gitignored) file and override the settings. env.development, and there is no need to edit .env files before each pull/push request

lilitkarapetyan: Here's the priority of the files for the development build and the production build:

Dev.: (npm start): .env.development.local, .env.development, .env.local, .env

Prod.: (npm run build): .env.production.local, .env.production, .env.local, .env

If you ever want to use something in your local environment without being specific to the development build or the production build, you can add some variables to your .env.local file.

lilitkarapetyan: local means specific configs for current server or workstation so it shouldn't be deployed to git. It can be also content server database passwords (just example not for our case)

lilitkarapetyan: Just checked we are using .env.local as main env file it need to be renamed as .env
изображение

IanMayo: That's great @lilitkarapetyan - thanks for explaining.  

(sorry if I forget what you've taught me, probably in a few days :-( )

IanMayo:
Have merged `develop` into this branch.  Now the client in the executable can connect directly to the server (via the window.location.origin parameter).

The application has built & run correctly.

PR #45, opened 2019-11-04
Fix incorrect templates assignment

### LINKED ISSUE #33: Logic bug in providing templates for player
A game designer has created a channel.  In it, Blue Role_A can send `Weekly Orders` and `Chat`.  Also in that channel, Blue Role B can send `Request for Information`. One the play starts, Blue Role B can only send `Weekly Orders` and `Chat`.

It looks like...  when we retrieve the message types for a force-role, we just retrieve the first message-type description for that force.

Instead, we should show the "union" of all message types allocated to that force-role in that channel.

zukilover:
## 🧰 Ticket
Closes #33 

## 🚀 Overview: 
Select templates based on current role instead of current  force

## 🤔 Reason: 
when we retrieve the message types for a force-role, we just retrieve the first message-type description for that force. Instead, we should show the "union" of all message types allocated to that force-role in that channel.

## 🔨Work carried out:

- [x] Update templates assignment on `SET_ALL_MESSAGES` action
- [x] Update templates assignment on `SET_LATEST_WARGAME_MESSAGE` action
- [x] Tests pass
[Logic bug in providing templates for player](https://app.gitkraken.com/glo/board/XZIuhoko5QAPaF0f/card/XbL4AdnTpQAP1k_h)

### DIFF
--- a/packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.js
+++ b/packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.js
@@ -173,14 +173,14 @@ export const playerUiReducer = (state = initialState, action) => {
             (channelActive || allRoles) &&
             !newState.channels[channel.uniqid]
           ) {
-            const participatingRole = channel.participants.some((p) => p.forceUniqid === newState.selectedForce && p.roles.some((role) => role.value === newState.selectedRole))
+            const participatingRole = channel.participants.find((p) => p.forceUniqid === newState.selectedForce && p.roles.some((role) => role.value === newState.selectedRole))
             const participatingForce = channel.participants.find((p) => p.forceUniqid === newState.selectedForce)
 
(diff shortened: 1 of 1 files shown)


IanMayo:
@zukilover - were you able to reproduce the problem from the current develop branch, and did you verify that this PR fixed it?

IanMayo:
I'm getting this error intermittently.  It _may_ be due to some inconsistency in the testing backend database - but it would be good to robustly handle it:

image


Hey, @zukilover -  I think I've fixed the bug. I've pushed it to this branch in this commit: https://github.com/serge-web/serge-web/pull/45/commits/3bb9396866e8732d7fba3e5fc97fcadd8b2ab97b

IanMayo (commented):
I've had a play, and saw the underlying bug was fixed.  But, on some occasions, I had a logic bug.  I think I've solved it - and pushed a commit.

Would you mind checking my commit please @zukilover ?

PR #58, opened 2019-11-18
try using "/" to force connection to current server

IanMayo:
## 🚀 Overview: 
@lilitkarapetyan has explained that we could use "/" as the REACT_APP_SERVER_PATH to force a connection to the current host.

This PR tests that suggestion.

## 🔗 Link to preview

## 🤔 Reason: 
Lilit's strategy is almost certainly better than Ian's.

## 🔨Work carried out:
Just trying "/" as REACT_APP_SERVER_PATH

## 🖥️ Screenshot
[If the work is UI related then paste a screenshot of the update here.]

### DIFF
--- a/packages/client/.env
+++ b/packages/client/.env
@@ -4,4 +4,4 @@ REACT_APP_VERSION=$npm_package_version
 REACT_APP_NAME=$npm_package_name
 # uncomment the next line to connect Serge to a server other than the one this
 # page was served from
-REACT_APP_SERVER_PATH='/'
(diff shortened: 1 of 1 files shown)


IanMayo:
I don't think we're ready for this.  

If we adopt this strategy, we need to remove the previous workaround (use current window URL if `REACT_APP_SERVER` not present.

foxleigh81:
@IanMayo if that's the case then let's reject the PR and close this branch down. The single-line change can be re-implemented at a later date and it keeps the repo clean.

Issue #62, opened 2019-11-19
Issue: Wrong channels displayed for player


IanMayo:
I have a game.  I've given a two players some templates to use:

* **White**: `State of the World`, `Chat`
* **Blue**: `Link`, `Chat`

But, when the Blue player logs in, he is only offered `Chat` templates.

If I change the templates from Admin page, blue player has valid set.  But, if he logs out/in again, he only gets the Chat template.  _Maybe_ he just gets the template that's common to all roles.

Ok. Whatever templates I offer to **Blue** **CO**, he is only offered `Chat` at initial login.  But, **White** gets the correct set of templates each time.  No, hold on. **White** sometimes gets the correct channels.

[![Image from Gyazo](https://i.gyazo.com/0ebc015f59938b118207ae7b59f22b14.gif)](https://gyazo.com/0ebc015f59938b118207ae7b59f22b14)

Screen-cast here:
https://i.gyazo.com/0ebc015f59938b118207ae7b59f22b14.mp4

IanMayo:
Closed in #66

PR #66, opened 2019-11-20
Live player additional bug fix

### LINKED ISSUE #61: Issue: Player UI falls over when live player added to a channel
I have a game configured.

I have players in a game.  

From the admin page, I add a force to a new channel.

The UI for that player then falls over:
image

Note:  if a specific role is provided for that channel participation, things work.  The issue is only if I add all Blue force to the channel.

[Image from Gyazo](https://gyazo.com/fd0a03c9c33283967cab70b7018b5d2b)

Steps to reproduce below:
https://i.gyazo.com/fd0a03c9c33283967cab70b7018b5d2b.mp4

zukilover:
## 🧰 Ticket
Closes #61 and #62 

## 🚀 Overview: 
Add conditional for all roles template selections

## 🤔 Reason: 
Adding new force with all roles selected, to a channel caused template to break

## 🔨Work carried out:
- [x] Add conditional template selection for all roles
- [x] Refactor template selection logic
- [x] Tests pass

### DIFF
--- a/packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.js
+++ b/packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.js
@@ -45,6 +45,46 @@ export const playerUiReducer = (state = initialState, action) => {
   const newState = copyState(state)
   let messages
   const channels = {}
+  const getParticipantStates = (channel, newState) => {
+    let chosenTemplates
+    let templates
+    let observing = false
+    const matchedForce = (p) => p.forceUniqid === newState.selectedForce
+    const matchedRole = (role) => role.value === newState.selectedRole
+    const participatingRole = channel.participants.find((p) => matchedForce(p) && p.roles.some(matchedRole))
+    const participatingForce = channel.participants.find(matchedForce)
+
+    if (!participatingForce && !newState.isObserver) return
+
+    const isParticipant = !!participatingRole
+    const allRolesIncluded = channel.participants.find((p) => {
+      return matchedForce(p) && p.roles.length === 0
+    })
+    if (participatingRole) {
+      chosenTemplates = participatingRole.templates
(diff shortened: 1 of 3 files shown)


IanMayo:
@zukilover - it looks like the e2e tests did their job.   The player UI crashes when logging in as a player.

While I can see the e2e results in Cypress, I don't know if you can.  I've made the Cypress project public. Hopefully you can click on the Cypress link, login via GitHub, and watch the video replay of the failed test.

Alternatively, you should be able to run **yarn test:e2e** - and see Cypress running locally.

IanMayo:
Serge does have quite complex logic when it comes to forces. roles & templates.

When things quieten down, I guess it may be worth re-factoring the logic.  This may make it more reliable.  It may also make it easier to test - so we can use our Unit Tests to ensure the logic carries on working.

I look forward to that.  But, in the near term we could really do with this branch being fixed.  If you think the above refactoring should be part of the fix, I'm quite happy to sponsor it.

IanMayo:
@zukilover  - here is the stack trace:
image

The e2e test is a walkthrough of this web-page:
https://sites.google.com/deepbluec.com/serge/introduction?authuser=0

But, I appreciate it's a lot of effort to get to that point.

I've just reproduced the problem locally by running `yarn build` in your branch, then `yarn serge` in one terminal, and `yarn test:e2e` in another.  The test fails, but it leaves an instance of the breaking wargame in the local server.

So, it's then possible to connect to the local server and go into the game. When I did that, any player (other than Game Control) caused the UI to fail.

Looking at the stack-trace, I think I've got it. Here is playerUI_Reducer:
image

If the method drops out at line 60 then there isn't a structure to decompose at line 75.

IanMayo (changes requested):
The player UI is crashing for this branch. I've tried to highlight the issue.

PR #69, opened 2019-11-21
E2e fix show access clicking problem

toan301:
It was that the state of checkbox refreshed after clicking, it is weird but might because we have another React event there

so what I did is to repeat the action click for several time (5 times) if it is not working then it is an error, but if it works, it will be sure to be checked


IanMayo (approved):
Working, thanks @toan301

PR #79, opened 2019-11-24
Message to multiple channels

### LINKED ISSUE #49: Send message to multiple channels
As a ***Serge Player*** I wish to be able to ***Submit some formatted messages for review prior to sending*** in order to ***ensure my stakeholders agree to my intention***.

As a ***Serge Player*** I wish to be able to ***Send some messages to other channels*** in order to ***share a message to roles in other channels so that they see the message but don't have to see the process of developing that message***.

We will implement this through combing the ability to send an open "Live" message as draft and to receive/handle feedback on that message, with the ability to send a message to multiple channels.

We should introduce two checkboxes in the `Admin` page that allows a player to send a message to multiple channels. The checkboxes would be present on each row of the players list for a channel in the `Channels` tab, and be titled "Multi-channel" and "Support Draft Messages".

If `Support Draft Messages` is set for a role/channel, the UI would change for that role/channel so that:
* In the `NEW MESSAGE` form,  the `SEND` button would switch to two buttons: `Send as Draft` and `Send`. 
* `Send as Draft`. When this button is clicked, the message is sent, but the edit form remains open - with the existing content present, so it can be edited and re-sent.
* Players can read messages coming back into that channel, and make further edits to the message.  Once they are happy with the message they can `Send`
* `Send`. When this button is clicked, the message is sent and the form closes (as at present).

If `Multi-channel` is set for a role/channel, then the UI would change so that
* Each of `Send` and `Send as draft` (if present) would switch to be a split button.  This button would have a drop-down list of other channels that this role is a member of.  There would also be commands to select `All` or `None` of the channels.  Ticking one or more of those channels, then clicking `Send` or `Send as draft` (respectively) would send that message to those channels.

The new control would behave roughly like this control (though ideally the child menu items would be represented with check-boxes rather than toggle state:
https://codesandbox.io/s/material-demo-m1kcz

Here is a mock-up of the new admin page:
image

With only "Multi-channel" ticked, the send button would become a split button like this:
image

With only "Support Draft Messages" ticked, a new button would appear next to the existing Save button:
image

With both boxes ticked for a Role, the message editor would have this at the bottom:
image

lilitkarapetyan:
## 🧰 Ticket
closes #49

## 🚀 Overview: 
Added the button for draft message sending
Added the multiselect dropdown button for  multi channel messages share 
Added the multiselect dropdown button for multi channel  draft messages share
Updated message list component to show message feedbacks and added the form for writing feedbacks 

## 🤔 Reason: 
Used material ui dropdown but I think its better to remove it and use some alternative form bootstrap (witch one already used in project)

## 📝 Developer Notes:
I want to test it and done few style fixes, it's working so can be deployed if necessary.


IanMayo (changes requested):
Thanks @lilitkarapetyan - I'm having a play. 

I've noticed that in the Admin screen I think we're missing some elements from the "view" version of a channel, which pulls the edit/delete buttons to the left:
image

IanMayo:
Another item of feedback @lilitkarapetyan - when a message has been "Sent as Draft" - the contents of the message have to stay in the form.  Then when the "reviewer" asks for changes in their feedback, the player can make the changes, and either send another draft, or send the message for real.

IanMayo:
Hmm, in some circumstances the "Send Draft" (with multiple recipients) is shown at the bottom-left with a drop-down, on other occasions it's shown at the bottom right, near the existing send message.

image

IanMayo:
Thanks for the update @lilitkarapetyan 

The form should _only_ be kept on "Send draft".  For "Send" it should be cleared (and collapsed)

IanMayo:
We're going to have to drop this work. It was defined/implemented at haste, trying to make it into the current wargame. But, It wasn't available in sufficient time to be incorporated in the game.  When we get further resources we should re-consider it, with UX, and re-implement. The code in this branch will remain available.

PR #86, opened 2019-12-04
Export messages game turn and tests

### LINKED ISSUE #72: Game turn marker not going to export
When we export a wargame to CSV, we aren't exporting the current turn number.

I suspect this isn't actually a function of the export process, it's more about injecting the current turn into the message when it's sent.

We can verify implementation of this by running a wargame, sending a few messages, then go back to the Admin page and exporting the wargame to XLS.  The exported file should contain the turn number.

Note: the turn number should go into messages, game admin messages, and insights/feedback.

Ideally this feature would also be covered in a unit test.  We should just send a message, then check that the turn id gets inserted.

lilitkarapetyan:
closes #72

## 🚀 Overview: 
Add game turn number for export messages, redesigned export messages Component code and added tests


IanMayo:
@foxleigh81  - I'd like you to have a look at this please.

My (incorrectly expressed) original requirement was that when we export messages to XLS, that each message includes the turn when it happened.

@lilitkarapetyan has interpreted this as wishing to export the turn markers.  This is valid, since it will improve the data available for analysis, but I still think it doesn't add the turn number to the message export.

I'm happy for any missing functionality to be covered in a separate PR - implemented under _Foxy's new ways of working_.

foxleigh81:
Yup. Let's address it in a separate PR. If this technically covers the tickets specification then we can close this one down and raise a new ticket for that additional functionality.

foxleigh81:
That said, I'd like to see those Code Climate issues addressed before the PR merges.

IanMayo:
@lilitkarapetyan  - could you have a go at fixing the issues identified by Code Climate?

lilitkarapetyan:
Working on it.

IanMayo (approved):
If code-climate is happy, then I'm happy :-)

PR #87, opened 2019-12-05
Pull row buttons to left, style alternate mission rows

IanMayo:
This is Ian's attempt at making long form based messages more legible.

It's very dodgy.


IanMayo:
Closing.  Let's look forward to a better implementation.

PR #119, opened 2020-01-03
Feature/118 mouse performance

### LINKED ISSUE #118: Optimise routing hexes
To solve a current issue, we have some logging installed in the mouse listener.  It outputs the next hex cell each time any mouse movement is registered.

We only actually need to recalculate the route if the cell under the mouse is different to the one the previous time the  event triggered.

IanMayo:
## 🧰 Ticket
Closes #118 

## 🚀 Overview: 
Reduce number of times we produce list of cells between two locations, and also reduce the number of Leaflet redraws.

## 🔗 Link to preview

## 🤔 Reason: 
Console logging showed that code to calculate possible routes was being called too often.

## 🔨Work carried out:
Introduce two sets of `if` constructs to test if we should bother recalculating route.

- [x] Tests pass

[Optimise routing hexes](https://app.gitkraken.com/glo/board/XZIuhoko5QAPaF0f/card/Xg9ahiNdWgAQmf_I)

### DIFF
--- a/packages/client/src/Helpers/MovementListener.js
+++ b/packages/client/src/Helpers/MovementListener.js
@@ -114,54 +114,57 @@ export default class MovementListener {
           this.planningLine.setLatLngs([this.startHex.centrePos, cursorLoc])
         }
 
-        // are we in a safe cell
+        // find cell under cursor
         const curCell = this.grid.cellFor(cursorLoc)
 
-        // is this an achievable cell?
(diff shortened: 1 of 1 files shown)


IanMayo:
@foxleigh81 - while we still had the console logging line present, it became obvious that we were re-calculating the route too often.  This PR removes the unnecessary recalculations and map updates.

IanMayo:
Close - implementations have moved on a long way since this point.

PR #149, opened 2020-01-14
Feature/142 Show leg markers

### LINKED ISSUE #142: 3.0 Resolve Turn - show leg markers
As a < Umpire >, I want < see attractive turn markers > so that < I can interpret planned routes >.

# Acceptance criteria

- [ ] Turn markers for planned (and historic) routes shown on relevant lines
- [ ] Turn marker is perpendicular to the line at that point.  The rotation of the line can be calculated as the angle of the lines before and after.  The angle of those lines can be calculated by asking Leaflet for the direction between the start/end points of the line.

It a turn marker is at the start (or end) of a line, it is just made perpendicular to the existing line.

# Step to reproduce
- [ ] Branch off this branch: 91-see-my-forces-state
- [ ] Start the game `MWARC Initialised`, login as `White Game Control`
- [ ] Hover over a green track at the top of the map, the planned route will appear

# UX
Turn marker is the diagonal line in this screenshot:
image

It is ok for the turn markers to be sized in screen coordinates, rather than map coordinates. This means the marker stays the same size as the map zooms in/out - like the platform markers do.

# Notes
The turn markers in this example are created in `MapAdjustingUmpireListener`, here:
https://github.com/serge-web/serge/blob/feature/91-see-my-forces-state/packages/client/src/Helpers/MapAdjudicatingUmpireListener.js#L201

IanMayo:
## 🧰 Ticket
Fixes #142 

## 🚀 Overview: 

_possible_ logic for solution:
1. The `for` loop at line 130 loops through all of the planned routes
2. The `forEach` function at line 136 loops through all of the planned steps in this route - handling the steps within the function titled `extendLine`
3. As we loop through, we could persist (save) values that we need later on.
4. So, we'd need: previous point along route, end of route marker, first point along new route
5. Once we have all of these, we've got enough data to determine the turn, and we can store the turn angle in the `payload` element that we put into the `turnMarkers` array.  When we come to build the actual markers, we'll have the turn angle available, and can select the relevant one.


IanMayo:
@lm93547  - good progress, the output is looking great.  

The `turn-end` marker is actually `turn-south`.

For the first marker, on Tanker-02, there aren't any positions at all, it's probably the same for when the fishing vessels are static for a few turns. I think you should introduce a 'no-turn' circle marker for this instance.    So, the start of Tanker-02 will just be a circle marker.

We also can't have the hard-coded lat/long in the `getTurnMarkerOrientationFor` function.  But, adopting the `no-turn` icon should avoid the need for it.

image

IanMayo:
Close. Now being managed in #206

PR #152, opened 2020-01-14
Use callback function to trigger form popup

IanMayo:
Work in progress for triggering React forms from LeafletJS


foxleigh81 (approved):
Looks good to me. I'll merge it in now so I can see it in action and see what I can do with it.

PR #153, opened 2020-01-14
138. Map autoupdate based on forces

### LINKED ISSUE #138: 2.0 Force Laydown - update wargame on message
As a < Game Control >, I want < to see the new Red force positions > so that < I can verify they are acceptable >.

# Acceptance criteria

- [ ] `MWARC Initialised` wargame loaded, playing as `Red CO`
- [ ] The game should be at `Turn 0 - Adjudication Phase` (I can provide updated wargame if you don't have it at this state)
- [ ] Before the game starts, Red is able to change the position of any red vessels/assets. This is called "Force Laydown:
- [ ] After Lilit's fixes, the `Force Laydown` messages are being sent via Node
- [ ] When received, our reducer needs to modify the `Forces` part of the wargame to with the correct location of the named asset.
- [ ] Also, when we have a new wargame state, the mapping control should be re-rendered

@lilitkarapetyan  - I've taken some early steps towards this in this branch:
* https://github.com/serge-web/serge/tree/feature/91_see_my_forces_laydown_callback

Covered in this PR:

* https://github.com/serge-web/serge/pull/139

lilitkarapetyan:
## 🧰 Ticket
Fixes #138 

## 🚀 Overview: 
This PR proves the message flow for updating the map on receipt of particular types of message.

It also provides a model for defining executable blocks of code that update on React changes.

## 🔗 Link to preview
https://serge-review-feature-91-efwsfv.herokuapp.com/

## Steps to reproduce:
* open two tabs, one as White, one as Red
* Both login to `MWARC Initialised`
* Here in Turn Zero, Adjudication Phase, Red is able to choose where his assets are (Force Laydown)
* When Red positions his assets, the asset will also move on the White player's map

### DIFF
--- a/packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.js
+++ b/packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.js
@@ -95,34 +95,49 @@ export const playerUiReducer = (state = initialState, action) => {
     return { isParticipant, allRolesIncluded, observing, templates }
   }
 
-  const modifyPlatformTypesBasedOnMessages = (platformTypes, newState) => {
-    // TODO: modyfy platform types based on laydown or other mapControll type messages
-    console.log(platformTypes)
-    return platformTypes
+  const modifyForcesBasedOnMessages = ({ allPlatformTypes, channels, allForces }) => {
+    let res = allForces
+    const mapChannel = Object.values(channels).find(({ name }) => (name === 'Mapping'))
+    if (mapChannel && mapChannel.messages && mapChannel.messages.length) {
+      for (const message of mapChannel.messages) {
+        res = modifyForcesBasedOnMessage(allForces, message)
+      }
+    }
+    return res
   }
 
-  const handleForceLaydown = (/* object */ payload, /* object */ newState) => {
+  const modifyForcesBasedOnMessage = (allForces, message) => {
+    let res = allForces
+    if (message.details && message.details.forceDelta) {
+      res = handleForceDelta(message, allForces)
+    }
+    return res
+  }
+
+  const handleForceLaydown = (/* object */ message, /* object */ allForces) => {
     // find the force
-    const force = newState.allForces.find(item => item.name === payload.force)
-    const asset = force.assets.find(item => item.name === payload.name)
+    const forceIndex = allForces.findIndex(item => item.name === message.force)
+    if (forceIndex === -1) return allForces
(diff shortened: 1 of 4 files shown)


IanMayo:
@lilitkarapetyan - I'm getting 'declared but not used' on one of your new methods:
image

Yes, I've just deleted it, and things still ran ok. Did you think it was necessary?

packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.js
```diff
@@ -95,34 +95,49 @@ export const playerUiReducer = (state = initialState, action) => {
...
+    let res = allForces
+    if (message.details && message.details.forceDelta) {
+      res = handleForceDelta(message, allForces)
+    }
+    return res
+  }
+
+  const handleForceLaydown = (/* object */ message, /* object */ allForces) => {
     // find the force
-    const force = newState.allForces.find(item => item.name === payload.force)
```
IanMayo, line 106: @lilitkarapetyan - you switched to using `findIndex` rather than `find`. Was this to solve a particular problem? Or is it just more robust programming method?

packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.js
```diff
@@ -143,8 +158,7 @@ export const playerUiReducer = (state = initialState, action) => {
       newState.gameDescription = action.payload.data.overview.gameDescription
       newState.allChannels = action.payload.data.channels.channels
       newState.allForces = action.payload.data.forces.forces
-      newState.allPlatformTypes = modifyPlatformTypesBasedOnMessages(action.payload.data.platform_types.platformTypes, newState)
```
IanMayo, line 146: Aah, @lilitkarapetyan - this is where the `modifyPlatformTypesBasedOnMessages` was previously called. I guess you can delete it now, right?

IanMayo:
@lilitkarapetyan  - I have one other question about the flow of information here:

* Red moves a counter
* A message gets sent to Node
* White is listening for `_changes` api, so gets new message
* White updates its map

Why doesn't Red also get the message from changes?

I know that when we send a `Chat` message, the sending browser doesn't need it.  But, I _think_ it could be better if Red does handle these state change messages. White uses the `handleForceLaydown()` method to change the `allForce` structure.  But, this structure won't get updated for `Red`.

So, _maybe_ Red needs to listen out for its own messages.  Or, when we do `sendMapMessage` we just drop the message into the reducer for the current browser?

What do you think?

foxleigh81 (approved):

PR #161, opened 2020-01-16
updated wargame.json to use arrays instead of nested objects

foxleigh81:
## 🚀 Overview: 
Updated wargame.json to use an array of objects for `plannedTurns`

## 🤔 Reason: 

Arrays are better to work with in this instance.

## 🔨Work carried out:

- [x] Updated wargame.json

## 📝 Developer Notes:

Some refactoring is likely to be needed to accommodate the change of data shape.

### DIFF
--- a/packages/client/public/wargame.json
+++ b/packages/client/public/wargame.json
@@ -1,439 +1,430 @@
 {
 	"channels": {
-		"channels": [
-			{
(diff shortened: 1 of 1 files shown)


IanMayo (dismissed):
Great, thanks.

IanMayo (approved):

PR #162, opened 2020-01-16
Feature/3 1 umpire change asset vis in playing mode

### LINKED ISSUE #159: 3.1 Change Entity Visibility (in Planning phase)
As a < Game Control >, I want < to control asset visibility in planning phase > so that < I can correct an omission during adjudication phase >.

It's possible that an Umpire may miss out on updating some asset visibility during the adjudication phase, and only realise while looking at a player's map during the playing phase.

It would be useful for a Umpire to rush back to their screen, click on the non-visible asset, and indicate that it is/isn't visible to that player force.

# Acceptance criteria

- [ ] Two browsers open, one as Umpire, one as Red
- [ ] Game in planning phase
- [ ] Form shown when umpire clicks on an asset
- [ ] Click on the form for an asset that isn't visible to Red
- [ ] Click on the control/checkbox to mark it as visible to the red force
- [ ] See the asset appear on the red browser map

# UX

# Implementation logic

On `listenTo()`:
1. Add popup handler to marker
2. When popup clicked, display the visibility (yes) no for this platform for each force (other than it's own force)
3. User can change visibility.  When user clicks on `Save`, message is sent, and the visibility of that asset changes for that force.

On `clearListeners()`: 
1. unbind popups for all markers

IanMayo:
## 🧰 Ticket
Fixes #159 

## 🚀 Overview: 
Allowed admin role to select an asset, then to change its visibility

At this stage, we don't have an actual form to use. So, we're mocking that interaction via buttons.

Note: these interactions are only provided while the game is in `Planning` phase.

## 🔗 Link to preview


packages/client/src/ActionsAndReducers/playerUi/helpers/handleVisibilityChanges.js
```diff
@@ -0,0 +1,30 @@
...
+/** create a marker for the supplied set of details */
+export default (/* object */ message, /* object */ allForces) => {
+  const payload = message.payload
+  payload.forEach(visChange => {
+    const asset = findAsset(allForces, visChange.assetId)
+    if (!asset) {
+      console.warn('failed to find asset while handling vis id', visChange.assetId)
+    } else {
+      if (visChange.newVis) {
+        // ok, we're making something visible
```
foxleigh81, line 12: This sort of comment would be more useful if written with context and as an instruction, that way it is more obvious as to what it does as a glance.

Instead of writing `ok we're making something visible` write it as an imperative, e.g.: `Make [name of the thing rather than 'something'] visible to [who it is visible to]`

If [name of the thing] and [who it is visible to] then indicate that in the comment. e.g.

`Makes $asset visible to $force`

It's not massively important that you do this now but it's something I plan to enforce post-rebuild so we might as well start as we mean to go on.

packages/client/src/Components/Mapping/helpers/MapPlanningPlayerListener.js
```diff
@@ -58,30 +43,105 @@ export default class MapPlanningPlayerListener {
...
+    this.routeLine.addTo(map)
+    this.plannedLine.addTo(map)
+
+    // use layer groups to store data - so we can confidently remove them
+    this.waypointMarkers = L.layerGroup().addTo(map)
+
+    // TODO: drop these fake state change triggers
     const context = this
-    this.createEasyButtonFor(map, 'fa-dice-one', 'immobile state selected', function (btn, map) {
+    this.btn1aImmobile = createButton(false, '1a. immobile state', function (btn, map) {
```
foxleigh81: It would be preferable to do the refactor replace `context` with `this`. 

I have refactored this in another PR though so it can be ignored for now.

packages/client/src/Components/Mapping/helpers/MapPlanningPlayerListener.js
```diff
@@ -58,30 +43,105 @@ export default class MapPlanningPlayerListener {
...
       context.platformStateAssigned(context.currentMarker, {
         mobile: false
       })
-    })
-    this.createEasyButtonFor(map, 'fa-dice-two', 'mobile state & speed selected', function (btn, map) {
+    }).addTo(map)
+    this.btn1bMobile10 = createButton(false, '1b. mobile 10kts', function (btn, map) {
```
foxleigh81: Pretty much all of these reference `btn` and `map` but none of them use either variable, is this a copy/paste thing? If so then they should be removed.

I have refactored this in another PR though so it can be ignored for now.

packages/client/src/Components/Mapping/helpers/MapPlanningUmpireListener.js
```diff
@@ -0,0 +1,139 @@
+import createButton from '../../../Helpers/createDebugButton'
+
+export default class MapPlanningPlayerListener {
```
foxleigh81: Should this not be `mapPlanningUmpireListener`?

packages/client/src/Components/Mapping/helpers/MapPlanningUmpireListener.js
```diff
@@ -0,0 +1,139 @@
...
+    this.btnChangeRedOn.remove()
+    this.btnChangeRedOff.remove()
+    this.btnChangeBlueOn.remove()
+    this.btnChangeBlueOff.remove()
+    this.btnSendChanges.remove()
+  }
+
+  changeVisPopupFor (/* object */asset, /* object */ perceptions) {
+    var res = ''
+    for (var key in perceptions) {
```
foxleigh81, line 74: `var` should be `let`

packages/client/src/Components/Mapping/helpers/assetsVisibleToMe.js
```diff
@@ -0,0 +1,32 @@
+import { UMPIRE_FORCE } from '../../../consts'
+
+/** provide a list of the assets that are currently visible to me
```
foxleigh81, line 3: Who is 'me' in this context? Rather than using words like `I`, `you`, `me` etc... it's more useful to provide real context, e.g if the `me` in the context is an umpire then the references to `me` should be replaced with `umpire`, if it's the current user then it should be replaced with `currentUser` etc...

packages/client/src/Components/Mapping/helpers/forceFor.js
```diff
@@ -0,0 +1,10 @@
+export default function forceFor (/* object */ allForces, /* String */ assetName) {
```
foxleigh81, line 1: There should be comments explaining what this does.

packages/client/src/Components/Mapping/index.jsx
```diff
@@ -130,32 +128,74 @@ const Mapping = ({ currentTurn, role, currentWargame, selectedForce, allForces,
...
+  const createThisMarker = (asset, grid, force) => {
+    // set the asset location
+    asset.loc = grid.hexNamed(asset.position).centrePos
+
+    const inForceLaydown = hasPendingForces(forcesRef.current, myForceRef.current)
+
+    // set the asset force
+    asset.force = force.name
+
+    var assetIsDraggable
```
foxleigh81, line 164: `var` should be `let`

packages/client/src/Components/Mapping/index.jsx
```diff
@@ -130,32 +128,74 @@ const Mapping = ({ currentTurn, role, currentWargame, selectedForce, allForces,
...
+        assetIsDraggable = myForceRef.current !== 'umpire' && asset.force === myForceRef.current
+        break
+      default:
+        console.log('Error - unexpected game phase encountered in Mapping component')
+    }
+
+    const userIsUmpire = myForceRef.current === 'umpire'
+    const marker = markerFor(asset, force.name, myForceRef.current, platformTypesRef.current, assetIsDraggable, userIsUmpire)
+
+    // did we create one?
```
foxleigh81, line 179: This is another example where an imperative comment with context would be more useful. Who is `we` and what did we possible create `one` of?

packages/client/src/Tests/ActionsAndReducers/playerUI/handleVisibilityReducer.test.js
```diff
@@ -0,0 +1,98 @@
...
+    by: 'Blue',
+    newVis: false
+  }, {
+    assetId: 'C05',
+    by: 'Red',
+    newVis: true
+  }
+]
+
+const allForces = [
```
foxleigh81, line 25: This JSON data shouldn't be in here, it should be in a separate file and referenced in any files that use it. Even if you only use data like this in one file, it's still good practice to separate it out, that way it becomes reusable code.

packages/client/src/Tests/mappingHelpers/assetsVisibleToMe.test.js
```diff
@@ -0,0 +1,84 @@
+/* global it expect */
+
+import assetsVisibleToMe from '../../Components/Mapping/helpers/assetsVisibleToMe'
+import { UMPIRE_FORCE } from '../../consts'
+
+/* sample dataset that is used across all tests in this file
+*/
+const allForces = [
```
foxleigh81, line 8: This looks like it's more or less the same as the `allForces` data from the previous test, so this is a good example of how useful making code reusable can be. :)

The differences in code are good as well as it shows that your functions are filtering code as expected.

packages/client/src/Tests/mappingHelpers/assetsVisibleToMe.test.js
```diff
@@ -0,0 +1,84 @@
...
+        }
+      },
+      {
+        name: 'foxtrot'
+      }
+    ]
+  }
+]
+
+it('I am umpire force', () => {
```
foxleigh81, line 59: This is personal preference but in my head, user stories can start with `I ...` (although I'd still argue for more context there too) but unit tests should again be imperative and also follow an 'it' statement in a way that is lexically correct, e.g.:

`it('correctly returns an umpire force')`

Although that may not be a perfect example is that's not really what it is doing. Perhaps something like:

`it('Returns 6 visible assets to the Umpire force and 'Alpha' should be amongst them')` (or something like that).

packages/client/src/Tests/mappingHelpers/findAsset.test.js
```diff
@@ -0,0 +1,49 @@
+/* global it expect */
+
+import findAsset from '../../Components/Mapping/helpers/findAsset'
+
+const allForces = [
```
foxleigh81, line 5: So this `allForces` looks like it might be significantly different, so it might be worth making an alternate `allForces` file for this, or if the differences in data don't HAVE to be there, then just alter your test to test for the generic `allForces` code from above.

yarn.lock
```diff
@@ -8956,11 +8956,6 @@ jest-message-util@^23.4.0:
     slash "^1.0.0"
     stack-utils "^1.0.1"
 
-jest-mock-console@^1.0.0:
```
foxleigh81, line 8959: Has this been removed? Was it not needed?

foxleigh81 (dismissed):
I've approved the PR even though I've made a lot of comments, that's because the comments are not deal-breakers and as the new ways of working stuff haven't even been put in a formal document yet, I'm not going to fail a PR for it. It's just all good stuff to bear in mind for the future.

yarn.lock
IanMayo: Hmm @foxleigh81,  I didn't consciously remove it.  Hmm, I can't remember if we're meant to commit `yarn.lock` files or not. I think I read conflicting opinions.

packages/client/src/Tests/mappingHelpers/assetsVisibleToMe.test.js
IanMayo: Aah, I see. Yes, I'll adopt it. I was struggling to recognise the value of the `it` bit.  Thanks.

packages/client/src/Tests/ActionsAndReducers/playerUI/handleVisibilityReducer.test.js
IanMayo: Aah @foxleigh81  , I was worried about a new test requiring a JSON change that would break an existing test. But, since all the tests get run, the dev will find out quickly enough if this happens. Great.

packages/client/src/Components/Mapping/helpers/MapPlanningUmpireListener.js
IanMayo: Wow, oh yes!

yarn.lock
foxleigh81: I suspect that you may have got an outdated yarn.lock file then, perhaps you accidentally did `npm something` rather than `yarn something` at some point? Or for some reason when you update your dependencies, the `yarn.lock` just didn't update. This might potentially be causing `yarn test` to fail if the CI tool doesn't have `jest-mock-console` installed. I'd expect this to still install it though provided it's in package.json so hopefully, it's not an issue.

PR #166, opened 2020-01-17
Feature/158 provide orders route management

### LINKED ISSUE #158: 5.0 Provide orders - route management
This issue covers non-visible functionality that currently forms part of the MapPlanningLayerListener.js module.

The module currently contains 3 pure virtual functions:
````
  resetCurrentLeg (/* dictionary */ plannedLegs, /* string */ turnName, /* object */ planningMarker)
  submitClearLastLeg (/* dictionary */ plannedLegs, /* string */ turnName, /* object */ planningMarker)
  submitClearWholeRoute (/* dictionary */ plannedLegs, /* object */ planningMarker)
````

The functions are to be extracted as helpers, implemented, and covered with unit tests.

Sample planned turns can be found in `wargame.json`

I suggest the testing covers the functions deleting from the start, the end, and mid-way. Also routes that just contain a single route element.

### PR DESCRIPTION
**Pull request by foxleigh81**
## 🧰 Ticket
Closes #158 

## 🚀 Overview: 
Adds TDD based functions for route management

## 🔨Work carried out:

- [x] Wrote tests for `resetCurrentLeg`
- [x] Wrote `resetCurrentLeg`
- [x] Wrote tests for `submitClearLastLeg`
- [x] Wrote `submitClearLastLeg`
- [x] Wrote tests for `submitClearWholeRoute`
- [x] Wrote `submitClearWholeRoute`
- [x] Moved unique helper functions into `Mapping` component
- [x] Hooked up new functions to `mapPlanningPlayerListener`

## 📝 Developer Notes

The hooked up functions do not currently update the markers or legs.

### CODE REVIEW

File: packages/client/src/Components/Mapping/helpers/submitClearLastLeg.js [hunk 1]
```diff
@@ -1,10 +1,8 @@
 /** the user has clicked on the end of a route, but he just wants to
    * drop the last leg
    */
-const submitClearLastLeg = (/* dictionary */ plannedLegs, /* string */ turnName, /* object */ planningMarker) => {
-  // pop the last leg in the dictionary
-
-  // update the planning marker
+const submitClearLastLeg = (/* dictionary */ plannedLegs) => {
+  return plannedLegs.reduce((d, i, idx, l) => idx < l.length - 1 ? [...d, i] : d, [])
```
IanMayo: oh dear, it will be a while before my JS skills are up to this standard.

Is there an 'old fashioned' version that's more accessible to this luddite?

packages/client/src/Components/Mapping/helpers/submitClearWholeRoute.js
```diff
@@ -1,10 +1,6 @@
 /**  a route has been planned, but the user wishes to start again. clear it */
-const submitClearWholeRoute = (/* dictionary */ plannedLegs, /* object */ planningMarker) => {
-  // erase all planned legs
-
-  // update the planning counter
-
-  // reset the initial available locations
+const submitClearWholeRoute = () => {
```
IanMayo, line 2: dropping the params, that's clever :-)

IanMayo (changes requested):
Could we just have the one change, to replace the reduce function?

If that method fails, I wouldn't understand the algorithm enough to check if it's working or not.

The function won't be called very often - so the performance isn't critical.

packages/client/src/Components/Mapping/helpers/submitClearLastLeg.js
foxleigh81: Done. It now looks like this:

```
const submitClearLastLeg = (/* dictionary */ plannedLegs) => {
  const newPlannedLegs = plannedLegs
  newPlannedLegs.pop()
  return newPlannedLegs
}
```

IanMayo (approved):
Great, thanks.

PR #177, opened 2020-01-20
[serge-168] Add dynamic serverPath based on base URL

### LINKED ISSUE #168: Infrastructure - robustly handle server not found
The front-end currently falls over if we are unable to connect to the server.  Aah, the backend will, too.

When this happens, we should fail gracefully, and show a page explaining that Serge cannot see the server.

Aah, actually, we can handle this more gracefully - we can also add a URL parameter for `host` - this would let a dev connect to their local server without doing the `.env` file modification, followed by a rebuild.  Or, for someone to connect to the Heroku hosted instance.

zukilover:
## 🧰 Ticket
Closes #168 

## 🚀 Overview: 
Refactor `host` with base URL instead of origin

Example of URL:
image

## 🤔 Reason: 
The front-end currently falls over if we are unable to connect to the server. Aah, the backend will, too.
When this happens, we should fail gracefully, and show a page explaining that Serge cannot see the server.

## 🔨Work carried out:

- [x] Refactor serverPath with base url value
- [x] Tests pass

## 📝 Developer Notes:
Front end should now read from either its base URL or from `?host` URL parameter

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,14 +21,22 @@ In your command line, navigate to the project and then run the following command
 
 This will start the full application in 'demo' mode, you can see the application in action by visiting: [http://localhost:8080](http://localhost:8080).
 
-For front-end developers relying on live rebuilds, they should first follow the above steps, and note the IP address & Port provided by the server script.  Then the `packages\client\.env` file should be mofidied, by using the server address for `REACT_APP_SERVER_PATH`.  The `https` should probably also be modified to `http`, too.
+For front-end developers relying on live rebuilds, they should first follow the above steps, and note the IP address & Port provided by the server script. Note: leave that Serge process running.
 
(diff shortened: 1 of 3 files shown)


IanMayo:
@zukilover - it looks like the parameter is called `host` rather than `serverPath`:
image

IanMayo:
Thanks for this @zukilover  - it's a great solution.  But, while experimenting with the URL, I was still getting a dumb crash.

I've done a _fairly bad_ job of showing a message when it wasn't able to connect to the backend.

But, could you modify it please so that it shows the full URL that the client was _trying_ to connect to?

Thanks :-D

IanMayo (approved):
Great solution - wish I'd thought of it :-)

PR #178, opened 2020-01-20
Development branch for Turn Markers

IanMayo:
Development branch, for Lucas to work on:
* [ ] update turn markers to new _fixed_ rotation angle
* [ ] introduce vertical turn marker
* [ ] debug display of turn markers

The turn markers are only shown for the route of the currently selected ship.  Before we can plan the route for a ship, we have to tell the ship what to do, since in some states the ship is effectively stationary.

This _will_ be supported via a popup, but for now we're using toolbar buttons.  So, start by clicking on a ship.  Then, click on the state `Transiting` to produce a route.  Then the drag cursor will be shown - move this around to "use up" the allowance for that 30 minute turn.

How to trigger turn markers (also see screencast below):
* open this URL: https://serge-review-feature-5--dsgibe.herokuapp.com/?wargame=wargame-k5kw38gf&access=p5543
* scroll down the map
* click on a Blue ship
* select `Transiting` from the toolbar buttons at the top left
* select the speed to travel at (higher speed means we travel further in each step)
* drag the drag cursor around to plan the route
* Debug lines will appear in the console, indicating which turn marker is being determined, and 

Screen Recording


IanMayo:
I've seen one turn type that is still wrong, T06 here:
Screen Shot 2020-01-20 at 08 58 40

IanMayo:
Integrate @lm93547 zoom icon changes

IanMayo:
Closing. Now managed in #206

PR #198, opened 2020-01-22
Feature/3 0 early adjudication

### LINKED ISSUE #191: 3.0 Adjudicate Wargame - Phase 1
As a < Umpire >, I want < to perform adjudication of planned routes > so that < I can produce a new state of the world >.

# Acceptance criteria

- [x] In wargame, in adjudication phase, as White, with planned turns for all/most forces
- [x] Have historic and planned tracks displayed
- [x] Planned for immediate next turn shown in bold, beyond that in feint
- [x] For selected asset, planned states shown as labels
- [x] For selected track, I have menu with 'Accept planned turn'.  Clicking this will show 'future state' marker for that asset, at the end of that planned route
- [x] A button is available for 'submit all states'  It shows the count of states produced.  Clicking on it sends new 'State of the world' message.

IanMayo:
## 🧰 Ticket
Fixes #191

## 🚀 Overview: 
* [x] UI interactions to allow individual plans to be accepted
* [x] UI interactions to allow all plans to be accepted
* [x] Algorithm to convert planned routes to actual routes
* [x] UI interactions submit new state of world
* [x] UI interactions to allow detailed planned routes to be displayed


packages/client/src/ActionsAndReducers/playerUi/helpers/handleForceDelta.js
```diff
@@ -0,0 +1,39 @@
...
+  PERCEPTION_OF_CONTACT,
+  SUBMIT_PLANS,
+  STATE_OF_WORLD
+} from '../../../consts'
+
+/** create a marker for the supplied set of details */
+export default (/* object */message, /* object */ allForces) => {
+  console.log('message', message)
+  const msgType = message.details.messageType
+  if (!msgType) {
```
foxleigh81, line 18: Is it possible we can move away from console.logging errors and actually use `throw`?

packages/client/src/ActionsAndReducers/playerUi/helpers/handleStateOfWorldChanges.js
```diff
@@ -0,0 +1,17 @@
+import findAsset from '../../../Components/Mapping/helpers/findAsset'
+/** create a marker for the supplied set of details */
+export default (/* object */ payload, /* object */ allForces) => {
+  const detail = payload.detail
+  detail.data.forEach(force => {
+    force.assets.forEach(entry => {
+      // ok, get the asset
```
foxleigh81, line 7: Similar to a comment I made on a previous PR, it would be better if you could make the comments a little less informal and a bit more imperative. e.g.

// Loop through the assets object and return a new object with additional items.

packages/client/src/Components/Mapping/helpers/MapPlanningPlayerListener.js
```diff
@@ -78,13 +79,10 @@ export default class MapPlanningPlayerListener {
     // command to submit whole planned route
     const context = this
```
foxleigh81: Is `context` needed here or can you refactor it out?

packages/client/src/Components/Mapping/helpers/clearButtons.js
```diff
@@ -0,0 +1,4 @@
+export default function clearButtons (/* array */ buttons) {
+  buttons.forEach(button => button.remove())
+  return []
```
foxleigh81, line 3: If you are returning an empty array anyway, then line 2 and the parameter are not needed. In fact, I was thinking about this the other day with my `getClearedRoute` function. It's such a basic function that just calling it 'clearArray' or something like that would make more sense as that way it's also reusable.

packages/client/src/Components/Mapping/helpers/routeLinesFor.js
```diff
@@ -0,0 +1,302 @@
...
+  return brng
+};
+
+/** we plot a line marker perpendicular to the direction
+ * of travel at route waypoints.  Determine the marker to use for this bearing
+ */
+function bearingMarkerFor (/* number */ angle) {
+  let icon
+
+  if (angle === 180 || angle === 360 || angle === 0) {
```
foxleigh81, line 114: Is there a reason this isn't a switch?

packages/client/src/Components/Mapping/index.jsx
```diff
@@ -161,8 +159,10 @@ const Mapping = ({ currentTurn, role, currentWargame, selectedForce, allForces,
   const clearControlledAssets = () => {
   }
 
-  const formRequestCallback = (form, payload) => {
-    console.log('Popup form requested for:', form, payload)
+  const newStateOfWorldCallback = (payload) => {
```
foxleigh81, line 161: ` const newStateOfWorldCallback = payload => {`

packages/client/src/Tests/mappingHelpers/newStateFromPlannedTurns.test.js
```diff
@@ -0,0 +1,80 @@
+/* global expect it */
+
+import newStateFromPlannedTurns from '../../Components/Mapping/helpers/newStateFromPlannedTurns'
+
+const legs1 = [
```
foxleigh81, line 5: Please move the data into their own files in the `Mocks` folder.

packages/client/src/Tests/mappingHelpers/newStateFromPlannedTurns.test.js
```diff
@@ -0,0 +1,80 @@
...
+  {
+    turn: 2,
+    state: 'Transiting'
+  },
+  {
+    turn: 3,
+    state: 'Fishing'
+  }]
+
+const noLegs = []
```
foxleigh81, line 47: You can leave this one here :P

packages/client/src/consts.js
```diff
@@ -44,6 +44,13 @@ export const FORCE_LAYDOWN = 'ForceLaydown'
...
 export const SUBMIT_PLANS = 'SubmitPlans'
+export const STATE_OF_WORLD = 'StateOfWorld'
+
+// time period to wait if server returns an error. One frequent cause of error
+// during development is that the server is stopped.  We're introducing a
+// throttle value to prevent the browser going into a race condition
+// as it sends 1000s of requests to the server
+export const ERROR_THROTTLE = 3000
 
 // Nov 2019. Ian modified the server path to use the
```
foxleigh81, line 55: I just noticed this, I know it's not from this commit, is this a note to highlight a hack? If so then it would be good to use a comment flag. The ones we should be using are as follows:

`//TODO:` Indicates an item which is still to do in the file
`//REFACTOR:` Indicates an area of code which could be written better, this is generally when you yourself have written something quickly and know it can be more efficiently written rather than to comment on someone else's code.
`//FIXME:` Indicates a bug or other problem which needs to be fixed but can't be done now
`//HACK:` Sometimes we just need to get stuff working and we have to do it in a way we're really not happy with. We can highlight them with the HACK flag.
`//NOTE:` This one is used for when you want to explain a decision or a reason for some code but not actually document the code itself.

These will be added to my ways of working guidelines when I start on the rebuild so it's up to you if you want to add that now or not.

foxleigh81 (dismissed):

packages/client/src/Components/Mapping/helpers/clearButtons.js
IanMayo: We need to call `button.remove()` to make it detach itself from the map UI.

packages/client/src/ActionsAndReducers/playerUi/helpers/handleForceDelta.js
IanMayo: Oh dear - I should have removed it.  Googling `throw`

IanMayo: Ah, I wasn't aware of `throw`. I see we'll need another level of infrastructure, to contain the `catch` elements.

packages/client/src/ActionsAndReducers/playerUi/helpers/handleStateOfWorldChanges.js
IanMayo: _will try harder_

packages/client/src/Components/Mapping/helpers/MapPlanningPlayerListener.js
IanMayo: Just tried - yes, I can lose it.  I believe there are still occasions when it's required - typically in Leaflet click callbacks, I think.

packages/client/src/Components/Mapping/helpers/routeLinesFor.js
IanMayo: I thought we could only do equality tests in a switch. We need to do some < tests:
image

packages/client/src/Components/Mapping/index.jsx
IanMayo: Ok.  So, if there's just one parameter, we don't need parentheses. Right?

packages/client/src/consts.js
IanMayo: Sure, added //NOTE:

packages/client/src/Components/Mapping/helpers/clearButtons.js
foxleigh81: ah, makes sense now :)

packages/client/src/Components/Mapping/helpers/routeLinesFor.js
foxleigh81: To the best of my knowledge, it's just an expression evaluator, the same as `if`

packages/client/src/ActionsAndReducers/playerUi/helpers/handleForceDelta.js
foxleigh81: You don't HAVE to use a `catch` element to use `throw`, it's just that without it the program will terminate. Sometimes that is the desired result, sometimes it's not.

packages/client/src/Components/Mapping/helpers/routeLinesFor.js
IanMayo:
```
switch(expression) {
  case x:
    // code block
    break;
```

To represent the above angle test in a switch, we could only do:
```
  switch(angle > 0 && angle < 90) {
    case: true:
       break
    case: false
       break
}
```
That wouldn't help @foxleigh81 😆

PR #199, opened 2020-01-22
Feature/platform types update

foxleigh81:
## 🚀 Overview: 
Changed the `name` field to use a friendly name rather than a kebab-cased name.

## 🤔 Reason: 
Names are easy to 'slugify' but less so to convert back to friendly.


IanMayo (approved):
Approved - we need the `friendly-name` for the perceived type dialogs.

PR #203, opened 2020-01-22
Fix for plotting complex routes

### LINKED ISSUE #202: Bug in storing planned route
**Describe the bug**
A player can plan a turn using a number of discrete steps.  But, when we store that route, we're just storing the direct route from the start cell to the end cell.

**To Reproduce**
1. Enter game as Player, instruct mobile platform to travel at more than 10 kts (maybe need to be Blue Frigate to do that)

**Expected behavior**
The route should he represented as the series of discrete steps

**Screenshots**
Animation of selecting route

IanMayo:
## 🧰 Ticket
Fixes #202

## 🚀 Overview: 
Fixed logic associated with planning complex routes (storing hex along whole turn, not just for this leg)

screencast


IanMayo:
@foxleigh81 - I confess, I've confused fixing a bug with refactoring the class.  It would have been better practice to do them separately.

Will try to do better next time...

PR #208, opened 2020-01-23
reinstate Cypress tests

### LINKED ISSUE #205: Reinstate Cypress Build
We now have an Open Source licence for Cypress.   We should reinstate it in the build, via the `travis.yml` script.

IanMayo:
Fixes #205 

Now that we have Open Source license for Cypress we can reinstate the tests.  They're passing ok locally - let's check they still pass under Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,8 @@ script:
 - cd ../e2e
 # manually get yarn to install
 - yarn install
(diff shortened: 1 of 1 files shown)


IanMayo:
Close the PR - the updated builds aren't being processed by Travis.

PR #216, opened 2020-01-23
Feature/164 record perception

### LINKED ISSUE #164: 5.3 Record player perception of other force
As a < Player >, I want < to record/correct my perception of an asset from another force > so that < I can plan tactical strategies >.

A significant part of this wargame is that Red can hide its assets (typically fishing vessels) as a Green asset - and use stealth to get near to Blue assets.

Red can also do this picture management to try to track which are the Blue assets.

We include UI to support this.

Blue/Red can do this "picture management" in playing or planning phase.  But, since the players don't have much to do during the adjudication phase, we can start off by giving them this capability during that phase.

This functionality can completely replace the functionality in `MapAdjudicatingPlayerListener`

The list of available platform types come from the `wargame.platformTypes` element in state.  The platform-types block does contain the SVG filename of the relevant icon.  _if_ the icon is shown in the platform types drop-down, then _maybe_ we should use that icon path instead of the pre-defined CSS path we're currently using (this will mean new platform-types can be added added from the user-configured wargame document, rather than via code change.

The transmission/receipt of these messages should be similar to how we current handle these events:
* change platform visibility
* change vessel position (force laydown)

# Acceptance criteria

- [x] Logged into `MWARC Initialised` as Red force, any turn, game in adjudicating phase (switch to White force if you need to advance the game).  We can click on any non-red asset, and assign its Force/Perceived-type
- [ ] When we change the what our perception of that asset is and save, the screen should update to show the new force/icon type.  The icon will also update for another player logged into our force in a different browser.

# UX
https://www.figma.com/file/2qiEitO08s8jXamHWI3VpP/SERGE-mapping-tool?node-id=0%3A1

screenshot

_Ideally_ we would show the icons in the platform-type dropdown, and shaded color blocks in the force drop-down.  But, we should start off with plain text for both, in order to test functionality/usability.

# Notes

IanMayo:
## 🧰 Ticket
Closes #164 

## 🚀 Overview: 
Creates a new component suite that can update the perception of a marker on click.

## 🔗 Link to preview
https://serge-review-feature-16-aatez3.herokuapp.com/

## 🤔 Reason: 
As a < Player >, I want < to record/correct my perception of an asset from another force > so that < I can plan tactical strategies >.

## 🔨Work carried out:

- [x] Imported the `mapPopupHelper` class from @lilitkarapetyan 
- [x] Moved the `MappingForm` component to be a child of `Mapping`
- [x] Created the `FormChildPerception` component
- [x] Added an `index.js` file to the `helpers' folder and import/exported all the helper functions
- [x] streamlined the helper imports in the `mapping` component
- [x] Connected `MappingForm` component to store
- [x] Connected `FormChildPerception` to store
- [x] Used `mapPopupHelper` to connect child store to parent store

## 📝 Developer Notes:
- The `index.js` file in `helpers` should be used from now on, however, importing files individually will also work so this shouldn't be a breaking change. It's worth keeping an eye on your codebase though, just in case.


IanMayo:
@foxleigh81 - I don't know what's a achievable, but are you able to make the popup look more material-like?
image

packages/client/src/Components/Mapping/components/FormChildPerception/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+  // A copy of the store to capture the updates
+  const newStore = store
+
+  const handleSubmit = (e) => {
+    e.preventDefault()
+    callbackFunction(newStore)
+  }
+
+  const handleRadioChange = ({ target }) => {
+    setForcePerception(target.value)
```
IanMayo, line 21: I'm probably mistaken, but aren't we storing the data twice. Do we need to `setForcePerception` _and_ `onSToreUpdate`?

packages/client/src/Components/Mapping/components/FormChildPerception/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+    newStore.perception.force = target.value
+    // save data in helper class to not lose it after popup recreate
+    onStoreUpdate(newStore)
+  }
+
+  const handleSelectChange = ({ target }) => {
```
IanMayo: Wouldn't `handlePlatformTypeChange` be closer to expressing business logic in the variable name?

IanMayo:
@foxleigh81 - the marker arrow is in the wrong place:
image

I guess the options are:
* position the dialog so it's centered over the asste
* move the arrow to the left-hand corner
* remove the arrow

Aah, there's also a UX request relating to it.  Would it be difficult to make the dialog draggable?  Some dialogs may obscure another platform that's related to the job in hand.  Making it semi-transparent would be a fall-back solution.

packages/client/src/Components/Mapping/components/FormChildPerception/index.jsx
foxleigh81: State and store are two different things, both are needed :)

packages/client/src/Components/Mapping/components/FormChildPerception/index.jsx
foxleigh81: completely agree. Done

foxleigh81:
I removed the arrow and made the popup look like the figma version.

IanMayo:
@foxleigh81 - it would be great if you could reduce the screen usage some more, please:
image

packages/client/src/Components/Mapping/components/FormChildPerception/index.jsx
IanMayo: I've had a play.  

1. Without `setForcePerception`, the UI doesn't update
2. Without `newStore.perception.force = target.value`, the counter doesn't update
3. But, we can remove `onStoreUpdate` without any impact.

foxleigh81:
> @foxleigh81 - it would be great if you could reduce the screen usage some more, please:

How's this? I've taken the padding down significantly to reduce the size and also added a small shadow so there is a bit more contrast when it overlaps the white part.

Screenshot 2020-01-24 09 25 59

IanMayo:
Are you able to reduce the whitespace to the right of the control labels?  _Ideally_, the column width would match their required width.  As a fallback we can set a width in pixels, and check by eye that no form labels go beyond that.

IanMayo:
I approve the work in this PR - I can't formally approve it, since I created it (as a way of getting a trial deployment)

PR #217, opened 2020-01-23
Correctly manage port number

IanMayo:
## 🧰 Ticket
The code to configure the port number was breaking the Cypress e2e tests, since Cypress builds need a port number.

However, the fixed script broke the comms on a MS Windows installation.

This PR fixes the runtime comms on MS Windows, and is available for test on other environments, including Heroku.


IanMayo:
This version works on local, Travis (Cypress) and Heroku. Fixed!

PR #222, opened 2020-01-24
[SERGE-170] Attach bypass URL to the role info

### LINKED ISSUE #170: Infrastructure - provide URL to come back into this wargame as this player
We can now login to a wargame via a URL.

It would be useful to be able to copy this URL from the UI.  So, we go through the login process as normal, but once we're in, we can get a copy of the URL params necessary to come back into that role.

Hmm, and the UI?

How about if the `Role` label near the bottom-right ('CO', 'Game Control', etc) was clickable - taking us to the URL.  Or, we can right-click on it, and 'Copy URL' using normal browser functionality:
image

Note: if the `host` parameter has been provided in the URL for the current page, this newly constructed URL should still include that `host` parameter.

zukilover:
## 🧰 Ticket
Closes #170 

## 🚀 Overview: 
Add bypass URL populated from the context state to the role info link

## 🤔 Reason: 
It would be useful to be able to copy this URL from the UI. So, we go through the login process as normal, but once we're in, we can get a copy of the URL params necessary to come back into that role.

## 🔨Work carried out:
- [x] Add bypass URL to role info at `AdminAndInsightsTabsContainer.jsx`
- [x] Update corresponding element's CSS
- [x] Tests pass

### DIFF
--- a/packages/client/src/Views/AdminAndInsightsTabsContainer.jsx
+++ b/packages/client/src/Views/AdminAndInsightsTabsContainer.jsx
@@ -34,11 +34,31 @@ class AdminAndInsightsTabsContainer extends Component {
       insights: 'Insights',
       model: FlexLayout.Model.fromJson(json),
       channelNames: [],
+      byPassUrl: null
     };
+  }
+
+  getByPassUrl() {
+    const [ state ] = this.context
+    const { currentWargame, allForces, selectedForce, selectedRole } = state
+    const currentUrl = new URL(document.location.href)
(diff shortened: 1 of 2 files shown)


IanMayo:
@foxleigh81 - I'd appreciate your review of this PR, please.

foxleigh81:
As this has a UI aspect to it, it would be useful to see a screenshot of the outcome. The code looks fine though so I'll approve it.

foxleigh81 (approved):

PR #247, opened 2020-01-27
Hotfix - formatting on leaflet marker pins

IanMayo:
We have styling for counters, but it was mistakenly being applied to all leaflet icons.

We 'caught' the shadow/border formatting, but not the 1px standard border.

So, put the formatting on the planning marker, not all leaflet icons.

Before:
image

After:
image


IanMayo:
Merging, I need the fix for a downstream issue

Issue #240, opened 2020-01-26
3.2 Entity Condition


IanMayo:
As a < umpire >, I want < to specify platform condition > so that < some player assets become unavailable >.

# Acceptance criteria

- [ ] Logged into wargame as White and as Red/Blue forces
- [ ] In adjudication phase, Game Control is able to click on a platform. A popup form (or control buttons) allow me to change the condition of a platform (according to the conditions provided in the platform-type for that asset)
- [ ] This new condition forms part of the new state of the world message, and is resolved accordingly
- [ ] The UI _should_ then update to reflect the new condition

# UX
Control form:
![image](https://user-images.githubusercontent.com/1108513/73134771-5b264a80-4032-11ea-8704-9263489aaad1.png)

Displaying different conditions:
![image](https://user-images.githubusercontent.com/1108513/73175620-84161080-4102-11ea-881c-b8be6e9e7e88.png)

## Preview
The tanker at the top-left is in disabled condition in this wargame:
https://serge-review-feature-3--zpqliz.herokuapp.com/?wargame=wargame-k5pap52f&access=p2311

Experiment with clicking on the platform to the right of it to see the `Selected` marker status

IanMayo:
Closing, part of #249

PR #248, opened 2020-01-27
Cosmetically indicate if platform in non-standard condition

IanMayo:
## 🧰 Ticket
Supports #240 

## 🚀 Overview: 
We're saying an asset is in a non-standard condition if its condition is something other than the first one listed.
When this applies, we'll cosmetically modify the marker - to encourage the user to hover over it.

This is currently done by applying a grey border to the marker. But, this border is lost when the marker switches to its 'selected' light-grey border.

## 🖥️ Screenshot
image

But, really @clenoble  would like the counter to take a grey shade all over:
image

@foxleigh81 - do you have any CSS tricks that will give the grey "wash" over the image?

(I'm reluctant to create another image _over_ this one, since we'll be introducing more entities on the map, and it may introduce click-handler troubles.

### DIFF
--- a/packages/client/src/Components/Mapping/helpers/markerFor.js
+++ b/packages/client/src/Components/Mapping/helpers/markerFor.js
@@ -21,14 +21,24 @@ export default (asset, grid, force, myForce, platformTypes, userIsUmpire, /* str
         icon: divIcon
       }
     )
-    res.bindTooltip(asset.name)
-    res.name = asset.name
-    res.force = asset.force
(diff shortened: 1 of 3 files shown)


IanMayo:
Aah, @clenoble - it appears we _were_ using hard-coded values, so I've been able to assign hard-coded values for disabled.  Here is green:
image

IanMayo:
Updated. Don't make it darker, make it fainter:
image

IanMayo:
Merge, before diverge too far from develop.

Issue #165, opened 2020-01-17
5.0 Plan Orders - Specify State


IanMayo:
As a < Player >, I want < to tell a vessel what to do > so that < I can control my forces >.

In the Planning phase, the players tell their ships what to do.  Asset ship/asset has a `platformType` value. We can look up this value in the `platformTypes` object in the wargame state.  

In that platform-state are two important elements.  `states` give a list of the possible things that platform type can do.  Each one includes a flag for whether it's a mobile state. For mobile states, the player chooses the speed to travel at.

So, when a player is in planning mode (`MapPlanningPlayerListener.js`) , and he clicks on one of his platforms, the dialog box shown below should open.  

The speed buttons should only be enabled if the player has selected a 'mobile' state in the top-section.

Ian is happy to implement collating the data for the popup, and to handle the returned data.  The _point_ of this issue is to prove Lilit's new way of doing dialogs.

# Acceptance criteria

- [ ] Logged into `MWARC Initialised` as `Red`
- [ ] Game in `Planning` phase (re-log in as White to move it forward, if necessary)
- [ ] Clicking on a Red asset opens the popup
- [ ] Switching on a 'mobile' state enables the speed radio buttons
- [ ] Clicking on 'plan route' triggers a callback that contains a state element like:
```
{ state:'transitting', speedKts:'20'}
```

# UX
![image](https://user-images.githubusercontent.com/1108513/72604640-384db500-3913-11ea-90c6-4de36963e98a.png)

# Notes

IanMayo:
Ok, couple of issues:

1. The test for whether to show speeds isn't whether the state is 'moored' or not. The test is whether the currently selected state (according to platform-type) has `isMobile` true or not (see below)
2. If the user has selected a state that isn't mobile, the callback should not contain the speed.

![image](https://user-images.githubusercontent.com/1108513/73281814-6b802600-41e8-11ea-8d61-49c2ba4315e8.png)

foxleigh81:
@IanMayo no worries, I'll take a look at that now.

IanMayo:
Done.

PR #252, opened 2020-01-28
the check to display speed (kts) or 'for x turns' is now controlled b…

foxleigh81:
## 🧰 Ticket
Fixes issue raised in #165 

## 🚀 Overview: 
The determining factor for showing or hiding the speed(kts)/for x turns elements was set to use the `mooring` state, this was incorrect and has now been changed to use the `mobile` state.


IanMayo:
@foxleigh81 - do we have the logic for:
 `don't return a speed if the state isn't a mobile one`

IanMayo:
Minor other request please. In `index.jsx` , could the default value for `turnsAtThisState` be 1?

They wouldn't select the state if they didn't want the asset to be in that state for at least one turn.

foxleigh81:
> @foxleigh81 - do we have the logic for:
> `don't return a speed if the state isn't a mobile one`

If the state isn't mobile then the option to return a speed (`handleSpeedChange`) is never called, so the state would never be updated.

IanMayo (approved):

Issue #213, opened 2020-01-23
UI - don't show channel title if only one channel


IanMayo:
For the current wargame, we will only have one channel - the mapping channel.

In wargames where there is only one channel, we do not need to display the channel title (`MAPPING` in the following screenshot), which will "gain" us some screen space.

![Screen Shot 2020-01-23 at 12 51 57](https://user-images.githubusercontent.com/1108513/72986019-2a9aa280-3ddf-11ea-98d8-501acaaddc95.png)

An _advanced_ version of this would be to only show the channel title if we're in more than one channel. So, the game may contain multiple channels, but we're only a member of one.

IanMayo:
Done.

PR #256, opened 2020-01-29
hide tab bar when one channel showing

nousernames2:
## 🧰 Ticket
#213 

## 🚀 Overview: 
Check channels length and if map is only channel then display accordingly

## 🔗 Link to preview

## 🤔 Reason: 
In wargames where there is only one channel, we do not need to display the channel title (MAPPING in the following screenshot), which will "gain" us some screen space.

## 🔨Work carried out:

```
    const channelsArray = Object.values(state.channels);
    const isOnlyMap = channelsArray.filter(channel => channel.name.toLowerCase() === "mapping").length === 1;

    if (channelsArray.length === 1 && isOnlyMap) {
      return <Mapping currentTurn={state.currentTurn} role={state.selectedRole} currentWargame={state.currentWargame} selectedForce={state.selectedForce} allForces={state.allForces} allPlatforms={state.allPlatformTypes} phase={state.phase} channelID={"map"} imageTop={imageTop} imageBottom={imageBottom} imageLeft={imageLeft} imageRight={imageRight}></Mapping>
    }

    if (channelsArray.length === 1 && !isOnlyMap) {
      return <Channel channel={channelsArray[0]} />;
    }
```

- [ ] Tests pass

## 🖥️ Screenshot
[If the work is UI related then paste a screenshot of the update here.]
image


packages/client/src/Views/ChannelTabsContainer.jsx
```diff
@@ -169,6 +172,17 @@ class ChannelTabsContainer extends Component {
     const [ state ] = this.context;
     let force = state.allForces.find((force) => force.uniqid === state.selectedForce);
 
+    const channelsArray = Object.values(state.channels);
+    const isOnlyMap = channelsArray.filter(channel => channel.name.toLowerCase() === "mapping").length === 1;
+
+    if (channelsArray.length === 1 && isOnlyMap) {
+      return <Mapping currentTurn={state.currentTurn} role={state.selectedRole} currentWargame={state.currentWargame} selectedForce={state.selectedForce} allForces={state.allForces} allPlatforms={state.allPlatformTypes} phase={state.phase} channelID={"map"} imageTop={imageTop} imageBottom={imageBottom} imageLeft={imageLeft} imageRight={imageRight}></Mapping>
+    }
```
IanMayo: @jcdiprose - I'm pretty sure there's a nested if here:
```
if(channel.len == 1) {
   if(isOnlyMap) {
     return <Mapping>
   } else {
     return <Channel>
  }
}
```

nousernames2: Either or, I prefer the seperate if statements, would you prefer it nested?

PR #261, opened 2020-01-30
Feature/249 adjudication

### LINKED ISSUE #249: 3.0 Adjudication
As a < umpire >, I want < to adjudicate the current turn > so that < I can move the game forward >.

This step covers the creation of the form  that is used by the Game Control to perform adjudication. That is, to accept the planned state (maybe including a route) for an asset, or provide an alternative outcome

Beyond providing the new state for the platform, the Game Control is also able to change who can see the asset, and what its condition is.

# Acceptance criteria

- [ ] Logged in as White, red and blue have provided planned routes, in Adjudication stage
- [ ] Click on an asset. The popup form appears.  Accept the planned state, and close the form.  The UI updates to show the asset at its new state/location, with the previous location shown faint.
- [ ] Click on another asset. The popup form appears.  Reject the state, select a new non-mobile state, and close the form.  The "new" marker will be shown over the previous markers (since it's a non-mobile state)
- [ ] Click on another asset. The popup form appears.  Reject the state, select a new mobile state, and close the form. The map will be in route-planning mode, with the umpire able to specify the route for a single turn. Once the route is planned, the umpire can click on the asset again, the form will popup, and he can "Accept" the planned route. The UI updates to show the asset at its new state/location, with the previous location shown faint.

Note: the above functionality is working in the game, but it is controlled by buttons rather than the form.

The popup form should be generated from within the `showAdjudicationAssetMenu`.  The UI popup form will replace the command buttons in the lower half of the method.

- [ ] Click on an asset, the form will popup.  If the route has been accepted, the top section is disabled (but with a 'revert' button).  But, the lower sections still allow the visibility and/or condition of the asset to be changed.  On changing these, the UI does not change, but the underlying datastore is modified to reflect the new selections.

Note: the visiblity/condition functionality is not working. I'm happy to provide support in creating the 'original' and 'edited' visibility and condition items.

# UX
https://www.figma.com/file/2qiEitO08s8jXamHWI3VpP/SERGE-mapping-tool?node-id=89%3A0

I suggest that @foxleigh81  has a Skype with @clenoble  to understand the flow of logic through this set of panels:
image

# Notes
A wargame has been created that "arrives" in the correct state for this.

This link will bring you in as Umpire, with planned routes present for green, blue and red:
http://localhost:3000/?host=http://localhost:8080&wargame=wargame-k5xxsr4b&access=p2311

foxleigh81:
## 🧰 Ticket
Closes #249 

## 🚀 Overview: 
Adds an adjudication form to the app as well as a refactor of how the mapping forms work

## 🤔 Reason: 

As a < umpire >, I want < to adjudicate the current turn > so that < I can move the game forward >.

## 🔨Work carried out:

- [x] Refactored `Mapping` and `FormContainer` components to allow for calling the right child component straight from `Mapping`
- [x] Modified `mapPopupHelper` to assist with the above
- [x] Built Form UI
- [x] Connected form ui to data
- [x] Connected component to wider store

## 📝 Developer Notes:
- @clenoble I've added you as a reviewer as I've slightly tweaked your design. I want to make sure you're happy with the changes I've made (they are visible in the screenshot above)
- The logic to choose which form to display is now in the `Mapping` componet and looks like this:

```js
  if (myForceRef.current !== 'umpire') {
        (myForceRef.current !== marker.asset.force)
          ? popup.useComponent(MappingForm, 'perception')
          : popup.useComponent(MappingForm, 'planned-status')
      } else {
        popup.useComponent(MappingForm, 'adjudication')
      }
```

The string to use as the second parameter can be found in the `componentRegistry` object in `FormContainer`


clenoble:
@foxleigh81 Do you have a URL where I can test this? The gif goes very quickly and there's no way to pause it to see all the changes for each interaction. Having the "Accepted" and "Rejected" statuses is a good idea, but it's redundant with Planned and Actual. What will be the status once the planned route is rejected and a new one is planned? Will it still be "Rejected"? 
For the save button at the bottom, I thought we agreed yesterday it was not needed in adjudication phase, only during planning, but maybe I missed something in the discussion.

IanMayo:
@clenoble  - playable version here: https://serge-review-feature-24-enrr0a.herokuapp.com/

We have to do some _housekeeping_ when the popup closes, to store the data.  So we need "some" kind of button at the bottom.  "Save" seems a better title than "Submit" - since "Submit" only happens at the end of the adjudication phase, for all assets.

clenoble:
Thanks @IanMayo for the link. 

When the user in white cell first opens the pop up, he can only accept or reject the planned route. He should not be able to interact with State or Speed before clicking on Reject. Once he has clicked on Accept, he should not be able to interact with State or Speed either unless he clicks on Revert. 

When the platform is immobile or destroyed, the top part of the form (State and speed) should be disabled. 

It's great to have the speed selection appearing only when a mobile state is selected. If you could add a short animation (sliding down from transiting for example), it would make it obvious why this is appearing and attract the user's attention.

There is no equivalent to the green checkmark I had on top. It does not correspond to the status (Accepted or Rejected). It corresponds to whether the user has checked this route (either accepted or rejected and plan a new one) or has not done anything yet. If we use "Rejected", it will not indicate whether a new route was planned, and it will add confusion. 

Also the same green checkmark is visible in the State of the World panel. If we replace it with Accepted or Rejected there, we will miss the major point: the important distinction here is whether each platform has been reviewed, not whether or not it has been accepted or rejected. We could keep the design and use a single "Reviewed" status.

foxleigh81:
> Thanks @IanMayo for the link.
> 
> When the user in white cell first opens the pop up, he can only accept or reject the planned route. He should not be able to interact with State or Speed before clicking on Reject. Once he has clicked on Accept, he should not be able to interact with State or Speed either unless he clicks on Revert.
> 
> When the platform is immobile or destroyed, the top part of the form (State and speed) should be disabled.
> 
> It's great to have the speed selection appearing only when a mobile state is selected. If you could add a short animation (sliding down from transiting for example), it would make it obvious why this is appearing and attract the user's attention.
> 
> There is no equivalent to the green checkmark I had on top. It does not correspond to the status (Accepted or Rejected). It corresponds to whether the user has checked this route (either accepted or rejected and plan a new one) or has not done anything yet. If we use "Rejected", it will not indicate whether a new route was planned, and it will add confusion.
> 
> Also the same green checkmark is visible in the State of the World panel. If we replace it with Accepted or Rejected there, we will miss the major point: the important distinction here is whether each platform has been reviewed, not whether or not it has been accepted or rejected. We could keep the design and use a single "Reviewed" status.

Thanks @clenoble I'll make those changes this morning. The only one I don't think I'll be able to do is the animation as (I believe) that will require a fair bit of code rewriting to make that achievable. However, I'll quickly look into it, just in case I'm wrong :)

IanMayo:
Thanks for that @clenoble 

I _don't think_ we can disable the state controls if the platform is immobile or destroyed, since the game currently treats the condition as just text.  It doesn't understand that some may mean the platform can't move.  Yes - it could in the future...

foxleigh81:
> I _don't think_ we can disable the state controls if the platform is immobile or destroyed, since the game currently treats the condition as just text. It doesn't understand that some may mean the platform can't move. Yes - it could in the future...

That shouldn't be a problem. I already have it set to disable speed if the platform is set to anything other than 'working'. I'm not 100% keen on using text-based assessments and I'd rather the conditions data be more like:

```json
[
    {
        "name": "working",
        "operational": true
    }
]
```

but for this wargame, I suspect that leaving it as `if condition !== 'working'` will be fine for now. Correct me if I'm wrong.

IanMayo:
You're completely right that we should embed the logic in the wargame design document.  This is the list of states the clients want:
image

So, 'Full capability' is the term you need to match.

Sadly, both of the first two terms allow for movement.  

It wouldn't be _all that_ challenging to let the condition come with a constraint.  We _may_ have chance to do it.  I suggest you put some code into the form _just in case_ the condition switches to an object that has a "mobile" boolean flag.   If it's in there, I may manage to make the modification myself.

foxleigh81:
>  I suggest you put some code into the form _just in case_ the condition switches to an object that has a "mobile" boolean flag. If it's in there, I may manage to make the modification myself.

It already checks for 'isMobile' so maybe you can use that, however it gets that flag from the transition status, not the condition.

IanMayo:
> It already checks for 'isMobile' so maybe you can use that, however it gets that flag from the transition status, not the condition.

Yes, the `mobile` condition flag is different. Actually, it should be `isActive`.  If `isActive` is true, then States can be selected.  If it's false then States cannot be selected.  This sits a level above `state.isMobile` - we're not deciding on if speed can be assigned, but if a state can be assigned.

foxleigh81:
> > It already checks for 'isMobile' so maybe you can use that, however it gets that flag from the transition status, not the condition.
> 
> Yes, the `mobile` condition flag is different. Actually, it should be `isActive`. If `isActive` is true, then States can be selected. If it's false then States cannot be selected. This sits a level above `state.isMobile` - we're not deciding on if speed can be assigned, but if a state can be assigned.

ok I've added the following to the `Adjudication` component:

```js
const checkIfActive = () => {
    const activeArray = ['working', 'full-capability']
    return activeArray.includes(_.kebabCase(markerCondition))
  }

  useEffect(() => {
    setIsActive(checkIfActive())
  }, [markerCondition])
```

We can add anything we want to that `activeArray` and if one of those values is selected it will hide everything in the top fieldset apart from the buttons. Obviously as soon as we can we should alter the data so we can replace all of that with:

```js
useEffect(() => {
    setIsActive(markerCondition.isActive)
  }, [markerCondition])
```

foxleigh81:
@clenoble @IanMayo I think I've covered everything now. This should be ready to review again. I even managed to work out how to make the animations work :)

IanMayo:
@foxleigh81 - WRT the `isActive` flag, this morning MWC came up with a new way of interpreting conditions. The first one is interpreted as "normal", the last one is interpreted as "destroyed", and the ones in the middle have varying degrees of limitation.

So, if the condition is any but the last one, the player (and umpire) can create new planned/actual states for it.  The umpire/observer will verify that the players are trying to do reasonable things within a particular condition.

foxleigh81:
> So, if the condition is any but the last one, the player (and umpire) can create new planned/actual states for it. The umpire/observer will verify that the players are trying to do reasonable things within a particular condition.

So should I rewrite it as:

```js
 useEffect(() => {
    setIsActive(markerCondition !== _.toLower('destroyed'))
  }, [markerCondition])
```

IanMayo:
> so should I rewrite as ...

If only :-) . We can't rely on the text being 'destroyed'.  We can only rely on it matching the last of the 'condition' entries for this platform type.

Note: I see the popup does have access to the `allPlatforms` structure.   Actually, there's also the sneaky `asset.platformTypeDetail` that we insert in the `markerFor` processing.

foxleigh81:
> > so should I rewrite as ...
> 
> If only :-) . We can't rely on the text being 'destroyed'. We can only rely on it matching the last of the 'condition' entries for this platform type.

ooh, they want me to check it's position in the array? I can do that but then we need to make sure we're very careful when we make updates to that array.

IanMayo:
> we need to make sure we're very careful when we make updates to that array.

That array is currently edited in vs-code.  But, in the future it will be edited by the backend server pages. In there we'll have full control over how the data gets represented.

_Maybe_ we should drop this "hidden logic" (last entry means disabled), and make the conditions into composite structures that include an "full/partial/zero" `activity` attribute.

clenoble:
@foxleigh81 For the status on top, when the user rejects the planned route, the status should be pending (or show nothing) until the user has planned a new route. Otherwise, they might reject and forget to plan a new route. It would still be shown as reviewed and they will have no way of spotting what they forgot.

Once a user has rejected the planned route, there should be 2 buttons on top "Plan route" and "Revert". "Plan route" allows the user to close the popup, clear (user-)planned route and plan a new route. 
Screenshot_2020-01-31 Figma(1)

For the header of the popup, I will need to re-think what's really needed. For the player, I don't think we need the color and type, but it might be more important for white (also because the popup for white is larger and therefore the marker might not be directly in the field of vision).

IanMayo:
@clenoble - yes, umpire needs a "plan route" button.  This is after he has rejected the planned route, selects a mobile state, and needs to drag the planning marker.

If the state isn't mobile, we don't need the "Plan route" button.

foxleigh81:
> @clenoble - yes, umpire needs a "plan route" button. This is after he has rejected the planned route, selects a mobile state, and needs to drag the planning marker.
> 
> If the state isn't mobile, we don't need the "Plan route" button.

@IanMayo Just so I'm clear. What is the expected functionality of the 'plan route' button? Should it clear out any existing data or should it just close the popup?

IanMayo:
> What is the expected functionality of the 'plan route' button? Should it clear out any existing data or should it just close the popup?

first it will empty any planned route, it's ok for the form to receive a callback to do that, if you like). Then it should close the form.  Once the player has dragged the route, the popup form will reappear. It will still be showing the (for example) `Transiting - 10 kts` planned state, and the umpire can accept this.

foxleigh81:
@IanMayo @clenoble the 'Plan route' button now exists, it just uses the regular `handleSubmit` function but passes through the `planStatus` state to the `Mapping` component

PR #262, opened 2020-01-30
Bugfix/leaflet css and planning markers

IanMayo:
Investigate issue where planning markers aren't shown for local version of leaflet.css, but are for remote copy.

With the red line, we see planning markers.  With the green line we don't:
image


IanMayo:
Closing. Resolved a missing image dependency that was available on remote hosting, but not locally.

PR #276, opened 2020-02-01
Visualise planning

### LINKED ISSUE #124: 2.1.2 - Visualise planning process for current force
As a < Player >, I want < see a summary of my planning process > so that < track planning >.

# Acceptance criteria
- [ ] An `ORDERS` panel to the left-hand side of the map is available to display planning progress. It does not to be semi-transparent, and it can fill the full vertical height
- [ ] On game load, a list of the assets (platforms) for the current force (or all assets for White) i is displayed n the panel (note: the `MapPlanningPlayerListener` class will populate the contents of the `allRoutes` object - which should trigger the UI update)
- [ ] As the wargame object is updated from LeafletJS (map listener), the summary information in the ORDERS panel should be updated.
- [ ] The `Send` button in the UI will trigger the `routeCompleteCallback`.  Note: Ian can refactor the logic to transform the `allRoutes` into the `payload` expected in this function
- [ ] A green tickmarker will be displayed after each platform name, if route.current.length > 0
- [ ] Initially we don't need the icon markers. Just the asset name, and (optional) tick marker

Notes:
- [ ] `develop` branch now contains sources that put an `allRoutes` object into `index.jsx`
- [ ] In `index.jsx`, the data-object containing the planned routes is called `allRoutes`.
- [ ] There is an entry in `allRoutes` for each platform that is being planned for.
- [ ] The name of each platform can be found in `route.asset.name`

Suggestions:
- [ ] in `mapping/index.jsx` we switch from a single `<div>` to two divs, one for the orders panel and the other for the map
- [ ] the `orders` panel re-renders on deep changes to `allRoutes`.  Or, we pass a callback to the `MapPlanningPlayerListener` - which fires the callback when substantial changes have been made to the `allRoutes` object.

I guess here is where we could define an `Orders` element that listens for changed in the `allRoutes` object.
image

# UX

Here is a version of the panel, though we will accept a simpler version:
image

In this phase, we will limit the panel to a vertical slice at the left-hand side of the map. It won't be semi-transparent.

# Notes: 
You can get directly in as Blue, at the Planning Stage here:
http://localhost:3000/?host=http://localhost:8080&wargame=wargame-k5xyxas9&access=p5543

(once the database change has been pulled through to `develop`)

I've posted animations into slack for how to create a new route for a ship, or to clear the route for a ship.  Here are copies:
adding new route

clearing planned route

lilitkarapetyan:
## 🧰 Ticket
closes #124

image


IanMayo:
Thanks @lilitkarapetyan - this looks a really good piece of work :-D

IanMayo:
Hello Lilit, there's just one different feature required, please.

For this panel to help the player track their process, it needs to update when they plan new legs.  So, if you're playing as Blue, and it's in Planning phase - when you plan a new turn for a ship, the count of planned turns for that ship should update.

When I investigated, it was clear that we couldn't just listen to `allRoutes` - since React only listens to the top level of the object.   If it helps, I could fire a callback each time the player adds/removes a planned turn.  Then I guess this callback could trigger your panel re-rendering.

Is that achievable?

PR #289, opened 2020-02-03
Feature/226 asset rename

### LINKED ISSUE #226: 5.3 - Perception of other force (asset name)
As a < Player >, I want < to record the name of a contact > so that < I can refer to it by name >.

A user pointed out that it was wrong for players to see the name for *Grey* asset.  But, as the game progresses, they may learn the name of assets, either through simulated intelligence briefings, them getting close enough to ID it, or other.

So, in the wargame definition, each asset should have a  unique `contact number`, something like `T032`.  The number won't give away which force the asset belongs to.   Players of other forces will only see the contact number.  But, in the **Perception** dialog, they will be able to enter a name.  So, this isn't the actual name - it's just their perception.  Once a player has a perceived name for an asset from another force, we display that name, not the contact number.

A player can clear the name in the popup, which will make Serge switch back to displaying the contact number.

# Acceptance criteria

- [ ] Game running, logged in as player
- [ ] Hover over an asset from another force, see it's contact number
- [ ] Click on asset, popup dialog shows contact number, but has text field for `Name`
- [ ] Player enters name, and presses `Save`
- [ ] The new name is now displayed when hovering over that counter

# UX
Tiny bit of ux, please @clenoble .

foxleigh81:
## 🧰 Ticket
Closes #226 

## 🚀 Overview: 
Refactoring of child forms and the addition of a new MarkerName form which allows users to give a marker a temporary name.

## 🤔 Reason: 
As a < Player >, I want < to record the name of a contact > so that < I can refer to it by name >.

## 🔨Work carried out:

- [x] Split the header out of each child form component into a singular `MarkerName` component
- [x] Connected `MarkerName` to the `mapPopupHelper` store
- [x] Renamed the components folder names to match the component names (best practice)
- [x] Marker name is replaced with a form on click
- [x] Option to submit a new name or revert to old name exists
- [x] Form updates state
- [x] Form updates store
- [x] Ability to rename asset restricted to player only

## 📝 Developer Notes:
- This PR contains a few refactors to the structure of the `Mapping` components children, please be aware of these changes when merging.
- As you can see by the screenshot, the marker itself does not update without a refresh, this is probably because some more integration needs adding between the two stores.


IanMayo (approved):
Nicely done.

PR #295, opened 2020-02-04
popup  draggable

### LINKED ISSUE #290: Make popups draggable
Sometimes a popup menu will obscure (hide) important information behind it.

It would be easier if the player could drag a popup form around.

The solution is to instruct Leaflet to make the popup draggable.

A demonstration of doing this is here:
https://plnkr.co/edit/S1GPRm6sNwGDkD6oSCHs?p=preview

Given the technical challenges we've faced in supporting Leaflet popups, it's probably best for @lilitkarapetyan to investigate this.

lilitkarapetyan:
## 🧰 Ticket
closes #290


IanMayo (approved):
Yes, it works great - once Ian has refreshed his browser :-(

PR #296, opened 2020-02-04
Responsive popup draggable

IanMayo:
Change to make popups appear in a more sensible location.


IanMayo:
Drop this, we're going to pursue a strategy where the popup is always in the same place.

PR #299, opened 2020-02-04
fixing issues referenced in #297

### LINKED ISSUE #297: Adjudication Form logic
We have a form for the White Force to perform adjudication:
image

It requires these changes:
1. On "Accept", "Plan Route" should not be enabled/revealed
2. On "Reject", "Plan Route" should only be enabled for states that are:
    * mobile
    * deploying (see example in PlanningForm)
3. If In "Deploying" state, the button should say "Deploy Asset"
4. On "Reject", it should not show "Reviewed". 
5. In the "Condition" drop-down, only the last condition should prevent "Status" from being managed.
6.  If the "Condition" is the last entry, then "Accept/Reject" are not required.
7.  Once a route is planned, and the form pops up again, It should appear as "Reviewed" - Accept/Reject are not required.  **Ian can handle this in MPPL.

foxleigh81:
## 🧰 Ticket
Closes #297 

## 🚀 Overview: 
Updated the logic on the Adjudication form component

## 🔨Work carried out:

- [x] On "Accept", "Plan Route" is not be enabled/revealed
- [x] On "Reject", "Plan Route" is only be enabled for states that are:
- [x] In "Deploying" state, the button "Plan Route" button says "Deploy Asset"
- [x] On "Reject", it does not show "Reviewed".
- [x] In the "Condition" drop-down, only the last condition prevents "Status" from being managed.
- [x] If the "Condition" is the last entry, then "Accept/Reject" are not shown.
- [x] Moved `OrdersPanel` into the mapping components folder
- [x] Fixed the `+/-` logic on the toggle for `OrdersPanel`


packages/client/src/Components/Mapping/components/Adjudication/index.jsx
```diff
@@ -29,17 +30,25 @@ const Adjudication = ({ store, onStoreUpdate, callbackFunction }) => {
...
-  const checkIfActive = () => {
-    const activeArray = ['working', 'full-capability']
-    return activeArray.includes(_.kebabCase(markerCondition))
+  const checkIfActive = (val) => {
+    return val !== getDisabledCondition
+  }
+
+  const checkIfDeploying = () => {
+    return markerStatus === 'Deploy'
   }
```
IanMayo, line 40: One subtle change here. You were close, good guess, but I'd like us to avoid the dependency on the string.

For `platformTypes` some states include a `deploying` flag:
image

We should test against that flag, please.

foxleigh81: Done.

packages/client/src/Components/Mapping/components/Adjudication/index.jsx
```diff
@@ -29,17 +30,25 @@ const Adjudication = ({ store, onStoreUpdate, callbackFunction }) => {
 
   newStore.formType = 'adjudication'
 
-  // HACK: This is just for the current wargame, this will need to be replaced with a non text-based comparison
+  const conditions = allPlatforms[0].conditions
```
IanMayo: This is a mistake - it's using the conditions available for the first platform listed.  We have to use the conditions for the `platformType` of this asset.

foxleigh81: On it.

foxleigh81: Done. Sorry about that, that's the peril of writing debug code and not flagging it!

IanMayo (approved):
Approved.

PR #313, opened 2020-02-06
Hotfix/state other than full can't open dialog

IanMayo:
Fix for issue where in adjudication mode, an asset sometimes can't be selected after it has been accepted/rejected.

The issue was that the planning marker was being placed over the current marker, obscuring it.

@foxleigh81 - could you test it please?


foxleigh81:
Great, works well.

foxleigh81 (approved):

PR #314, opened 2020-02-06
added workaround for the name change bug

### LINKED ISSUE #303: Allow user to delete Perceived Name, before writing new one.
A user was having trouble entering a perceived name into the text box this afternoon.

She wanted to delete the existing text (C023), and then type in the new one.

But, each time she pressed delete, the C023 to re-inserted.

I guess the current workflow is to prevent null/empty strings. But, I know (well, think) the game robustly handles a missing name - it's the normal situation. When it's an empty String the game just displays the contact number.

I was able to train that user, but I won't be able to on Friday. So, could we modify the text box so that the user can delete the name?

recording

foxleigh81:
## 🧰 Ticket

Closes #303 

## 🚀 Overview: 
Added a workaround for the bug listed in the ticket

## 🔨Work carried out:

- [x] Replaced 'value' with 'placeholder' so the user can see the previous value but doesn't need to delete it
- [x] Styled the placeholder so that it is a faded grey.

### DIFF
--- a/packages/client/src/Components/Mapping/components/MappingForm/styles.scss
+++ b/packages/client/src/Components/Mapping/components/MappingForm/styles.scss
@@ -49,6 +49,13 @@
        position: relative;
        padding-left: 40%;
        margin-bottom: 1rem;
+
+input {
+    &::placeholder {
+        color: rgba(#828282, .4)
+    }
+}
+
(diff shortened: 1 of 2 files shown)


IanMayo (approved):
Works great.

PR #316, opened 2020-02-14
fixing issue causing map to fall over if 'perception' isn't defined

IanMayo:
Pending fix from previous development cycle.

Creating PR to help us decide what to do with it.


IanMayo:
@foxleigh81 - I'd appreciate your guidance on how to handle the changes in this branch.

(no rush)

IanMayo:
I think this was an issue that came out during the game, but didn't need a fix at the time.

It relates to the legacy mapping, but I still think it's worth handling, so the code that we "pull over" to the new framework is in the best condition.

I recall very little of the change, but it does appear I was asking for your support in how to resolve the conflict. 

Hopefully you're able to advise on what we need to keep in `index.jsx`

IanMayo:
Closing. It appears the changes have already been applied in `develop`.

PR #322, opened 2020-02-28
Feature/serge refactor component port

foxleigh81:
## 🚀 Overview: 
Imports new `ProgressIndicator` component into Storybook and also the serge client package.

## 🤔 Reason: 
I needed a proof of concept to show that this will work properly and give myself and others a pattern to work with for future components.

## 🔨Work carried out:

- [x] Updated various config files as and when the need arose
- [x] Added new `ProgressIndicator` component (replaces old `ProgressBar` component from the client package)
- [x] Set new component up in Storybook
- [x] Created component register in the components package to allow for easy import (see 'Developer notes' for more info)
- [x] Created various build scripts to make life easier for devs
- [x] Imported new component into the client package to ensure it works as expected.

## 📝 Developer Notes:
- It is vitally important that this is not prematurely merged into `develop` at this stage, it WILL cause broken builds as the new build scripts have not been integrated into the CI platform.
- The component register now allows for easy import of components into the client package, instead of having to type:

```js
import ComponentName from '@serge/components/build/local/component-name'
```

You can instead type:

```js
import { ComponentName } from '@serge/components'
```

The downside is that any new component must be added to the registry manually to make it work (the register is the `index.ts` file in `src/`. I am working on creating an automatic register but I've not managed to get it working yet.


packages/client/src/Views/GameSetup.jsx
```diff
@@ -1,146 +1,134 @@
...
+  }
 
-    if (!this.checkWargameNameSaveable()) return;
+  const currentTabSetter = tab => dispatch(setCurrentTab(tab))
 
-    if (typeof this.state.newWargameName === 'string' && this.state.newWargameName.length > 0) {
-      this.props.dispatch(saveWargameTitle(this.props.wargame.currentWargame, this.state.newWargameName));
+  // Loop through wargame data and build up an array of tab objects
+  const allTabs = () => Object.entries(wargame.data).map(tab => {
+    console.log('running')
```
IanMayo: @foxleigh81  - can the above debug line be removed?

packages/components/src/local/progress-indicator/README.md
```diff
@@ -0,0 +1,28 @@
+# Progress Indicator
+
```
IanMayo, line 2: I guess you've got some more content to go in here. I guess it's having a least one, so we're familiar with your intentions for language/tone/level of detail.

foxleigh81: Actually no, not really. The parts inside  are auto-populated from inside storybook. I might add some extra comments to make that clearer though.

I'd say that this part:

`The Progress Indicator component is used to indicate the users progress through a form.`

Could be more fleshed out I guess but I wasn't sure what else to put.

packages/client/src/Views/GameSetup.jsx
foxleigh81: Whoops. I'll fix that in a sec.

IanMayo (approved):

Issue #374, opened 2020-04-17
Tie correct Dialogs to Assets


IanMayo:
## 💬 Brief description of feature

We have to display the correct dialog to Asset click handlers, depending upon the phase of the game, and the force the player is logged in as.

## 📖 User Story

* As a Player I wish to provide planning orders for my platforms
* As a player I wish to record my perception of an asset from another force
* As a umpire I wish to perform adjudication on planned states for assets

Clicking on an asset should give these outcomes:
* if the game is in a planning phase, and I click on an asset of my force (or one I have delegated control over), the planning dialog should open
* if the game is in a planning phase, and I click on an asset of another force (or one I don't delegated control over), the perceptions dialog should open
* if the game is in a planning phase, and I am umpire, and I click on an asset I control, the planning form should open.
* if the game is in a planning phase, and I am umpire, and I click on an asset I don't control, no form should open
* if the game is in adjudication phase, and I'm an umpire, the adjudication dialog should open
* if the game is in adjudication phase, and I'm not an umpire, and I click on an asset of another force, the perceptions dialog should open

## ☑️ Acceptance criteria

- [ ] Above requirements met
- [ ] Tests pass

It should be possible to assess success by having knobs that allow selection of:
* player force (white/blue/red)
* game phase (planning/adjudication)

Actually, maybe the bulk of the logic should be in a single helper function with this API:
```
function relevantDialog(playerForce: string, 
    assetForce: string, 
    gamePhase: string): enum<dialogToOpen>
```

IanMayo:
Closed in #379

PR #389, opened 2020-04-23
Feature/374 tie up logic

foxleigh81:
## 🧰 Issue
#374 

## 🚀 Overview: 
Clicking on a marker now displays the correct form string in the map bar.

## 🔗 Link to preview
https://www.reddit.com/r/memes/comments/g6ofcq/i_aint_clucking_around/

## 🤔 Reason: 
This sets up our ability to populate the map bar with the correct components.

## 🔨Work carried out:
- [x] Tied up the logic
- [x] Updates the state
- [x] Output example string in storybook
- [x] Tests pass

## 🖥️ Screenshot

Screenshot 2020-04-23 at 15 31 39

## Confirmations

## 📝 Developer Notes:
- This is foundational work for later work.


IanMayo:
I _think_ there's an outstanding change.

If we do a click that results in no form being available (such as White clicking on Blue in Planning phase),  I think we should clear the form area - rather than leave the previous form there.

foxleigh81:
> I _think_ there's an outstanding change.
> 
> If we do a click that results in no form being available (such as White clicking on Blue in Planning phase), I think we should clear the form area - rather than leave the previous form there.

What should go in its place? Looking at Celines designs, the map bar part is actually transparent so we could just say that there is no form there and the map is visible underneath?

IanMayo:
Yes, that's fine.
In this part-way version, just the original empty panel would seem ok.

foxleigh81:
> Yes, that's fine.
> In this part-way version, just the original empty panel would seem ok.

On it.

IanMayo (approved):

PR #393, opened 2020-04-24
add map ref for childrens and hide/show grid coords on zoomend

### LINKED ISSUE #377: Hide grid labels on zoom above 11
The labels for the hex cells are only shown at zoom levels 11 or below.

So, as we zoom in, the labels appear, then they disappear as we zoom out:
zoom labels

The 'legacy' way of doing this is here:
https://github.com/serge-web/serge/blob/develop/packages/client/src/Components/Mapping/index.jsx#L106

But, this won't work in our new declarative way of doing it.

I'm not really sure how we can introduce this behaviour.

_Maybe_ we need to create a fresh Leaflet Layer that has min/max zoom levels.

If that's the case, then this project may give us some tips:
https://github.com/auto-mat/leaflet-zoom-show-hide

**Note:** this feature should branch from `feature/369_assets_layer` if it's still alive. Otherwise, it should branch from `feature/328-mapping-refactor`.

lilitkarapetyan:
## 🧰 Issue
closes #377

## 🔗 Link to preview

https://deploy-preview-393--serge-storybook.netlify.app/

## 🤔 Reason: 
I dont like the beta package https://github.com/auto-mat/leaflet-zoom-show-hide
So I think the "The 'legacy' way of doing this" is more suitable.
I had done the function with one can forward some props to child component, and for Mapping component childrens I had added mapRef variable (P.S.: I am not able to get it works with standar react ref types and had used `any` type) 

## 🖥️ Screenshot
I had found some issues connected with coords 
P.S this screenshot was done before the issue was fixed.

### DIFF
--- a/packages/components/src/local/hex-grid/index.tsx
+++ b/packages/components/src/local/hex-grid/index.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react'
+import React, { ReactNode, useState } from 'react'
 import L from 'leaflet'
 import { PointLike } from 'honeycomb-grid'
 import { Polygon, Marker, LayerGroup } from 'react-leaflet'
@@ -12,8 +12,18 @@ import { MapContext } from '../mapping'
 import SergeHex from '../mapping/types/serge-hex'
 
 /* Render component */
-export const HexGrid: React.FC<PropTypes> = ({ gridCells }: PropTypes) =>
-  <MapContext.Consumer>
+export const HexGrid: React.FC<PropTypes> = ({ gridCells, mapRef }: PropTypes) => {
+
+  const [showCoords, setShowCoords] = useState(false)
+  // only show the markers when zoomed in
+  if (mapRef && mapRef.current.leafletElement) {
+    mapRef.current.leafletElement.on('zoomend', () => {
+      setShowCoords(mapRef.current.leafletElement.getZoom() >= 11)
+    })
+  }
+
+  return (
+    <MapContext.Consumer>
     { (context): ReactNode => {
       // collate list of named polygons
       const polygons: { [id: string]: L.LatLng[] } = {}
@@ -58,21 +68,26 @@ export const HexGrid: React.FC<PropTypes> = ({ gridCells }: PropTypes) =>
           />
         ))}
         </LayerGroup>
-        <LayerGroup key={'hex_labels'} >{Object.keys(centres).map(k => (
-          <Marker
-            key = {'hex_label_' + k}
-            position={centres[k]}
-            width="120"
-            icon={L.divIcon({
-              html: k,
-              className: styles['default-coords'],
-              iconSize: [30, 20]
-            })}
-          />
-        ))}
-        </LayerGroup>
+        {showCoords &&
+          <LayerGroup key={'hex_labels'} >{Object.keys(centres).map(k => (
+            <Marker
+              key = {'hex_label_' + k}
+              position={centres[k]}
+              width="120"
+              icon={L.divIcon({
+                html: k,
+                className: styles['default-coords'],
+                iconSize: [30, 20]
+              })}
+            />
+          ))}
+          </LayerGroup>
+        }
       </>
     }}
   </MapContext.Consumer>
+)
+}
+
 
 export default HexGrid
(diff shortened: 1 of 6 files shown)


IanMayo:
@lilitkarapetyan  - the code fails when I change the grid-cell size:
https://deploy-preview-393--serge-storybook.netlify.app/?path=/story/local-mapping--with-grid

recording

IanMayo:
Thanks for spotting the label issue @lilitkarapetyan . Reported here: https://github.com/serge-web/serge/issues/394

IanMayo:
@foxleigh81 - I can't really tell if @lilitkarapetyan  is using our _standard_ way of passing data.  Would you mind having a look?  Part of my things that `mapRef` should be in the context.

packages/components/src/local/mapping/index.tsx
```diff
@@ -94,6 +95,10 @@ export const Mapping: React.FC<PropTypes> = ({
     setCurrentForm
   }
 
+  const childrenWithProps = Children.map(children, child =>
```
foxleigh81: Context should remove any need to use this. Please refactor this to pass `mapRef` down via context

packages/components/src/local/hex-grid/types/props.d.ts
```diff
@@ -6,4 +6,5 @@ export default interface PropTypes {
    * the grid of hex cells
    */
   gridCells?: SergeGrid<SergeHex<{}>>
+  mapRef?: any
```
foxleigh81: Please stick to the format of using Types, do not add a type without describing what it is.

packages/components/src/local/hex-grid/index.tsx
```diff
@@ -12,8 +12,18 @@ import { MapContext } from '../mapping'
 import SergeHex from '../mapping/types/serge-hex'
 
 /* Render component */
-export const HexGrid: React.FC<PropTypes> = ({ gridCells }: PropTypes) =>
-  <MapContext.Consumer>
+export const HexGrid: React.FC<PropTypes> = ({ gridCells, mapRef }: PropTypes) => {
+
+  const [showCoords, setShowCoords] = useState(false)
+  // only show the markers when zoomed in
+  if (mapRef && mapRef.current.leafletElement) {
```
foxleigh81: I believe typescript now allows you to do this: 

`mapRef?.current.leafletElement`

packages/components/src/local/mapping/types/props.d.ts
```diff
@@ -89,5 +89,5 @@ export default interface PropTypes {
    * @default false
    */
   zoomAnimation?: boolean
-
+  children?: any
```
foxleigh81: Please add comments.

foxleigh81 (changes requested):
I too get the `Cannot read property 'leafletElement' of null` error when I try to resize. 

I'm fine with us using mapRef but I'd like to see it passed down with context rather than cloning the element. 

Also, I'd like to know the reasons for not using `leaflet-zoom-show-hide`, I'm afraid "I don't like it' isn't a good enough reason. What problems does it cause and what problems would using it solve?

Made a few other small comments too. :)

packages/components/src/local/mapping/index.tsx
lilitkarapetyan: Changed to context using

packages/components/src/local/hex-grid/types/props.d.ts
lilitkarapetyan: removed

packages/components/src/local/hex-grid/index.tsx
lilitkarapetyan: thanks, got it

lilitkarapetyan:
leaflet-zoom-show-hide
2 year no updates 
beta version
isn't popular one
I dont think we need it just for one (or few) zoom checkings

IanMayo (approved):
Nice work @lilitkarapetyan

packages/components/src/local/mapping/index.tsx
```diff
@@ -73,6 +73,7 @@ export const Mapping: React.FC<PropTypes> = ({
   /* Initialise states */
   const [showMapBar, setShowMapBar] = useState(false)
   const [currentForm, setCurrentForm] = useState('')
+  const [zoomLevel, setZoomLevel] = useState(zoom || 0)
```
IanMayo, line 76: This is good, a zoom-level hook could be useful in other features being able to decide whether to show themselves or not.

foxleigh81 (approved):
Nicely done :)

foxleigh81:
> leaflet-zoom-show-hide
> 2 year no updates
> beta version
> isn't popular one
> I dont think we need it just for one (or few) zoom checkings

Yeah, good enough reasons for me :) Thank you. It's good to capture thoughts like this on PR's in case we need to reference them in the future :)

PR #404, opened 2020-04-28
403 - Introduce planned route rendering

### LINKED ISSUE #402: Element to display route being planned
## 💬 Brief description of feature

As the user drags the **planning marker**, we plot a line indicating the extent of that route.

This is represented in legacy Serge mapping as both a shaded set of rectangles and a distinct line.

For an asset with a movement range that is restricted by speed, the extend of the planned route is limited to the available cells.

## 📖 User Story

A planning line is displayed between two named cells.  

If the destination is limited by a set of available cells, the line should be limited to that extent.

It is expect that this feature can be verified through adding a `destination` cell to the existing `available cells` story.

## ☑️ Acceptance criteria
* [ ] with an empty destination cell, no planning line is displayed
* [ ] with a destination set, a planning line is plotted from the origin to the destination
* [ ] if a set of available cells are provided, the planning line is restricted to their extend

IanMayo:
## 🧰 Issue

Fixes #402 

## 🚀 Overview: 

Displays a line (& filled hexes) to show the path of the route that's currently being planned.

## 🔗 Link to preview

https://deploy-preview-404--serge-storybook.netlify.app/?path=/story/local-mapping--with-allowable-range

## 🔨Work carried out:

* [x] create new structure, to store cells that represent new route
* [x] change hex styling test, to verify if cells are part of planned route, before checking if they're part of available cells
* [x] collate polygon centres, to be plotted as line
- [x] Tests pass

## 🖥️ Screenshot

PlannedAnimation

## Confirmations

## 📝 Developer Notes:

There may be some performance inefficiencies in this PR, related to switching between representing hexes as honeycomb-hexes, and as plain strings.  That will be investigated in another PR.

### DIFF
--- a/packages/components/src/local/hex-grid/index.tsx
+++ b/packages/components/src/local/hex-grid/index.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useState, useContext} from 'react'
 import L from 'leaflet'
 import { PointLike } from 'honeycomb-grid'
-import { Polygon, Marker, LayerGroup } from 'react-leaflet'
+import { Polygon, Marker, LayerGroup, Polyline } from 'react-leaflet'
 /* Import Stylesheet */
 import styles from './styles.module.scss'
 
@@ -14,7 +14,7 @@ import SergeHex from '../mapping/types/serge-hex'
 /* Render component */
 export const HexGrid: React.FC<PropTypes> = ({ gridCells }: PropTypes) => {
       
-      const { gridCells: gcProp, allowableCellList, zoomLevel  } = useContext(MapContext).props
+      const { gridCells: gcProp, allowableCellList, zoomLevel, plannedRouteList  } = useContext(MapContext).props
 
       // collate list of named polygons
       const polygons: { [id: string]: L.LatLng[] } = {}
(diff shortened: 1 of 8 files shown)


packages/components/src/local/hex-grid/index.tsx
```diff
@@ -59,6 +64,17 @@ export const HexGrid: React.FC<PropTypes> = ({ gridCells }: PropTypes) => {
         centres[hex.name] = centreWorld
       })
 
+      // create a polygon for each hex, add it to the parent
+      const plannedRoutePoly: L.LatLng[] = []
```
foxleigh81, line 68: This looks like it could be a map and filter.

foxleigh81 (approved):
One refinement that could be made but otherwise, it's all good. Happy to approve this.

packages/components/src/local/hex-grid/index.tsx
IanMayo: I know it's a performance hog.  

Once this PR is out of the way, I think I need to do an implementation change.  We're passing these lists of hexes round as string cell-names.

That means in too many places we have to do: `gc.find(cell => cell.name == listCellName)`

I think that instead, i should be passing around the actual lists of Cells, not cell names.

PR #409, opened 2020-04-30
[DO NOT MERGE] experiment with creating cells within useEffect block

IanMayo:
Experimental branch, to investigate how we can reduce the amount of rendering

## Issue
The whole map is getting re-rendered (with all its children) when any state changed.  I'm going to work through the code, and verify we're getting the minimum necessary leaflet redraws


IanMayo:
Closing. The UseEffect practice has been formally adopted in another PR. Experiment complete.

PR #410, opened 2020-05-01
fix polygon render on classname change

### LINKED ISSUE #403: Updating leaflet elements
As we implement our functionality in StoryBook components, we're increasing our levels of Leaflet integration.

But, I've come across an issue. If we change the size of a polygon, it gets re-rendered in Leaflet.  But, if we only change the css-class of the element, then it does not get re-rendered.

I found that I could "force" the element to re-render by changing the `key`.

But, this feels like it is fighting leaflet, and seems unhealthy.

It would be good if we could remove the need to mangle the key value.

Here's an example of having to append a uniq id (random number in this case) to the key:
image

lilitkarapetyan:
## 🧰 Issue
closes #403 

## 🚀 Overview: 
Leaflet elements not works as react component. As we know the leaflet can take only static html as item. So those elements updating with functions. By default updating function updates only on positions change.

## 🤔 Reason: 
I had created custom react-leaflet Polygon component witch one will extends "Path" class from "react-leaflet"  and added the equal code as in "react-leaflet" Polygon class and add the changes the update function to detect classname changes and update those. 

## 📝 Developer Notes:
I am not able to extend Polygon class from "react-leaflet" as it exported:
`export default withLeaflet<Props, Polygon>(Polygon)`
So I created the the custom component

### DIFF
--- a/packages/components/src/local/hex-grid/helpers/polygon.ts
+++ b/packages/components/src/local/hex-grid/helpers/polygon.ts
@@ -1,0 +1,46 @@
+import { Path, withLeaflet } from 'react-leaflet'
+import type { LatLng, PathProps, PolylineProps } from 'react-leaflet'
+import { Polygon as LeafletPolygon } from 'leaflet'
+import { difference } from 'lodash'
+
+// uodate LeafletPolygon class to be able to covert it to LeafletElementtype
+class LeafletPolygonWithPositions extends LeafletPolygon {
+  positions = []
+}
+
+// define props
+type Props = {
+  positions: LatLng[] | LatLng[][] | LatLng[][][],
+} & PathProps & any
+
+// Leaflet element PathProps & PolylineProps required props for Path class
+type LeafletElement = PathProps & PolylineProps
+
+// create custom polygon class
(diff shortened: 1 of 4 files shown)


packages/components/src/local/hex-grid/helpers/polygon.ts
```diff
@@ -0,0 +1,46 @@
...
+    // update the component on positions change (copied from standard leaflet polygon)
+    if (toProps.positions !== fromProps.positions) {
+      this.leafletElement.setLatLngs(toProps.positions)
+    }
+    this.setStyleIfChanged(fromProps, toProps)
+
+    // logick wich one will compare classnames and modify the element class attribute instead of re render it
+    // this will alow to not touch elements witch ones have no changes in classname
+    if (toProps.className !== fromProps.className) {
+      const fromClasses: string[] = fromProps.className.split(' ')
```
IanMayo, line 37: hey, this is a fancy piece of lodash logic :-D

IanMayo (commented):
Great piece of work @lilitkarapetyan

PR #411, opened 2020-05-01
Tidy mapping state & introduce planning marker

### LINKED ISSUE #402: Element to display route being planned

IanMayo:
## 🧰 Issue

Fixes #402

## 🚀 Overview: 

Lots of `mapping` component was refactored to reduce updates.  This was done by putting as much processing as possible into `useEffect` blocks that only got triggered when related data changed.

## 🔗 Link to preview

https://serge-review-hotfix-tid-vwfshm.herokuapp.com/storybook/?path=/story/local-mapping--with-allowable-range

## 🔨Work carried out:

* switch to useEffect in `mapping` and `hex-grid`
* add draggable marker, with handlers for drag and drop
* propagate end of drag back up to mapping

- [x] Tests pass

## 🖥️ Screenshot

image

## Confirmations


packages/components/src/local/hex-grid/helpers/to-screen.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/** convert this point in world coordinates to cell
+ * @param {any} point location in world coordinate system
+ * @param {L.latLng} origin location of hex origin (0,0) in world degree system
+ * @param {number} tileSize diamater of hex tiles in Degrees (used as
+ * scaling factor)
+ * @returns {L.LatLng} coordinate in 2d Hex degrees
+ */
+const toScreen = (point: L.LatLng, origin: L.LatLng, tileSize: number): PointLike => {
+  var latVal = (origin.lat - point.lat) / tileSize
```
foxleigh81: Don't use `var` use `let` or `const`

packages/components/src/local/hex-grid/index.tsx
```diff
@@ -6,97 +6,253 @@ import { Marker, LayerGroup, Polyline } from 'react-leaflet'
...
 
-      // Use direct property if available, otherwise, use context prop.
-      const gc = gridCells || gcProp
-      const setCellStyle = (cell: SergeHex<{}>, pc:Array<SergeHex<{}>>, ac: Array<SergeHex<{}>>): string => 
-      `${pc && pc.includes(cell) ? 'planned' : ac && ac.includes(cell) ? 'allowable' : 'default'}-hex`
+      //  allow the achievable range to be changed
+      const [planningRange, setPlanningRange] = useState<number | undefined> (planningRangeProps)
 
-      // Watch the 'allowableCellList' property for changes and update the state accordingly
+      /** uility function, to style a cell according to if it's in
```
foxleigh81: This should be in `./helpers`

packages/components/src/local/mapping/helpers/createGrid.ts
```diff
@@ -50,17 +51,55 @@ const createGrid = (bounds: L.LatLngBounds, tileDiameterMins: number): SergeGrid
...
+  sergeGrid.toWorld = (point: PointLike): L.LatLng => {
+    return toWorld(point, sergeGrid.origin, sergeGrid.tileDiameterDegs)
+  }
+  sergeGrid.toWorld2 = (point: PointLike, origin: L.LatLng, delta: number): L.LatLng => {
+    return toWorld(point, origin, delta)
+  }
+  sergeGrid.cellFor =  (latLng: L.LatLng): SergeHex<{}> | undefined => {
+
+    // convert to hex coordinates
+    var hexCoords: PointLike = sergeGrid.toScreen(latLng)
```
foxleigh81: No `var`

packages/components/src/local/mapping/helpers/createGrid.ts
```diff
@@ -50,17 +51,55 @@ const createGrid = (bounds: L.LatLngBounds, tileDiameterMins: number): SergeGrid
...
   sergeGrid.origin = correctedOrigin
   sergeGrid.tileDiameterDegs = tileSizeDegs
+  sergeGrid.centerOffset = centreOffset
+  sergeGrid.toScreen = (point: L.LatLng): PointLike => {
+    return toScreen(point, sergeGrid.origin, sergeGrid.tileDiameterDegs / 2)
+  }
+  sergeGrid.toWorld = (point: PointLike): L.LatLng => {
+    return toWorld(point, sergeGrid.origin, sergeGrid.tileDiameterDegs)
+  }
+  sergeGrid.toWorld2 = (point: PointLike, origin: L.LatLng, delta: number): L.LatLng => {
```
foxleigh81: Can you find a better name for this? What is the difference between `toWorld` and `toWorld2`?

packages/components/src/local/mapping/helpers/createGrid.ts
```diff
@@ -50,17 +51,55 @@ const createGrid = (bounds: L.LatLngBounds, tileDiameterMins: number): SergeGrid
...
+  const cellOrigin = hexOne.coordinates()
+
+  // capture the offset between a cell centre, and the cell origin
+  const centreOffset = L.point(centreH).subtract(L.point(cellOrigin))
+  
   const sergeGrid: SergeGrid<SergeHex<{}>> = asSerge as SergeGrid<SergeHex<{}>>
   sergeGrid.origin = correctedOrigin
   sergeGrid.tileDiameterDegs = tileSizeDegs
+  sergeGrid.centerOffset = centreOffset
+  sergeGrid.toScreen = (point: L.LatLng): PointLike => {
```
foxleigh81, line 83: I don't really understand what is happening here. Are you assigning a function to a variable in `sergeGrid`? If so, why? You already have a helper called `toScreen` so what is going on here, can't you just pass data to that? Can you add comments so it's clear?

The same goes for the other instances of `sergeGrid.thing` below.

foxleigh81 (changes requested):

packages/components/src/local/mapping/helpers/createGrid.ts
IanMayo: The function comment is in the class definition.

I've renamed it to `toWorldLocal`

packages/components/src/local/mapping/helpers/createGrid.ts
IanMayo: Sure, making them class members, they can refer to other class data (particularly the origin & grid size).  That means that in the code, when we have a `gridCells` object, we can just call `gridCells.toWorld('H02')` - and it will us the origin/tilesize that's already stored in the grid.

packages/components/src/local/mapping/helpers/createGrid.ts
IanMayo: Correction, I've now deleted it. It wasn't used.

packages/components/src/local/mapping/helpers/createGrid.ts
IanMayo: It's a way of hiding the value inside the method.  The `toWorld` method does not need the cell origin or grid size passed to it, since it can access it from the parent object.  It also reduces the need for other code to know about how to do coordinate transforms.

foxleigh81: Fair enough, can you add some comments to the code to make it clear for the future.

foxleigh81 (approved):

PR #412, opened 2020-05-07
Feature/352 dialogue forms postback

foxleigh81:
## 🧰 Issue
#352 

## 🚀 Overview: 
Adds the last elements to the dialogue forms data-handling, provides ability to display selected values and also post data back to state.

## 🔗 Link to preview

https://serge-review-feature-35-l5v4kj.herokuapp.com/storybook/?path=/story/local-adjudicateturnform--default

## 🤔 Reason: 
Forms are useless if they don't display accurate data and can post back.

## 🔨Work carried out:

- [x] Migrates form states from `MapBar` to `Mapping` components
- [x] Adds `postBack` value to allow for updating state from forms
- [x] Updates `RCB` component to accept a value and uses it to determine the selected/checked data
- [x] Updates `Turns` component to accept a value
- [x] Splits form state into `populate` and `values`
- [x] Updates all forms to use the new form state
- [x] Implements correct state values on forms within `Mapping`
- [x] All forms now post back to state
- [x] Stories updated 
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
- Storybook appearing on a random port each time was bugging me so I have fixed the storybook port to `6611`
- `SetXFormData` values now have `populate` and `values` objects in them. The former tells the forms what options to include and the latter tells the form what the initial values are. The `values` section is also what is updated when the form is submitted.
- The styling, transitions and other visual niceties will be added in a future 'theming' PR; for now, I'm only concerned about the logic.


IanMayo:
In `adjudicate form`, the `visible to` needs to be checkboxes, not radio buttons - since the asset can be visible to multiple forces.

IanMayo:
In `plan turn form` when `transiting` is selected, player should be able to select speed.

packages/components/src/local/adjudicate-turn-form/index.tsx
```diff
@@ -15,23 +11,71 @@ import ForcePicker from '../form-elements/force-picker'
...
-        <PlannedRoute />
-        <RCB type="radio" label="Status" options={status} />
-        <RCB type="radio" label="Speed (kts)" options={speed} />
+      { plannedRouteStatusVal === 'accepted' && <span> Reviewed </span>}
+      { conditionVal.toLowerCase() !== 'destroyed' && <fieldset>
+        <PlannedRoute name="plannedRouteStatus" status={plannedRouteStatusVal} updateState={clickHandler} />
+        {
+          plannedRouteStatusVal === 'rejected' && <div>
+            <RCB type="radio" label="Status" options={status} value={statusVal} updateState={changeHandler}/>
+            { statusVal.toLowerCase() !== 'moored' && <RCB type="radio" label="Speed (kts)" name="speed" options={speed} value={speedVal} updateState={changeHandler}/> }
```
IanMayo: should the above line be split across multiple lines?

IanMayo:
Aah, sometimes ForcePicker is checkbox (perceived as), other times it's radio-buttons (visible to).

IanMayo (changes requested):

foxleigh81:
> Aah, sometimes ForcePicker is checkbox (perceived as), other times it's radio-buttons (visible to).

Interesting. Ok I can do that.

packages/components/src/local/adjudicate-turn-form/index.tsx
foxleigh81: Yes, it should :)  Well spotted.

packages/components/src/local/adjudicate-turn-form/types/props.d.ts
```diff
@@ -1,19 +1,33 @@
...
+   */
+  formHeader: string
+  /**
+   * All types in this definition are options for a form input
+   */
+  formData: FormData
+ /**
+   * The method for posting data back to state
+   */
+  postBack?: any
```
IanMayo: I think it's worth capturing the structure of the _data nugget_ that gets posted back, since by definition it's a data structure that's shared between multiple components.

_Maybe_ the TS type will be something like the types we encountered in MappingContext:
```
  setSelectedAsset: React.Dispatch<React.SetStateAction<SelectedAsset>>
```

Aah, I think we're nearly there.  If the `values` element of `FormData` is declared as an independent type,  say `FormValues`, then that element becomes an element of `FormData`, and the type that gets passed in the postback.

foxleigh81: Basically, any time you see `any` here it's because I was trying to focus on getting the functionality working rather than getting bogged down in typing. However it's a fair cop, I shouldn't really be getting into the habit of raising PR's without going through the `any` entries. I'll make sure this and others are as resolved to types as possible before I mark this as resolved.

IanMayo:
Sorry I didn't spot this earlier. In the `Plan route` form, the `steps` field is only visible for non-mobile states.

So, if it's 'moored", the player may indicate that they will stay moored for 5 turns.  But, if it's 'transiting' the player drags the marker to indicate where the platform goes in each turn.

foxleigh81:
> Sorry I didn't spot this earlier. In the `Plan route` form, the `steps` field is only visible for non-mobile states.
> 
> So, if it's 'moored", the player may indicate that they will stay moored for 5 turns. But, if it's 'transiting' the player drags the marker to indicate where the platform goes in each turn.

Fixed :)

IanMayo (approved):

Issue #401, opened 2020-04-28
Element to display available cells - Phase 2


IanMayo:
## 💬 Brief description of feature

Create react element to show available cells for a platform, initially titled `AvailableCells`

## 📖 User Story

As a player, I wish to see the cells that a platform could travel to in the next step

For this phase, we will be implementing the terrain-type sensitivity. That was deferred from issue #397 

The repo contains a file of data indicating which cells are land, sea, other, called `cell-types.js`. The _legacy_ code to read in the array, and process it, is in `GridImplementation.js`.

## ☑️ Acceptance criteria

* hex cells are shaded according to whether the platform can travel there in the next step, using these elements:
    * origin cell (hex coords)
    * travel restrictions for asset (sea/air/land)
    * number of cells (range) for that platform in the next turn

- [ ] Tests pass

## 📐 Designs / Wireframes

![image](https://user-images.githubusercontent.com/1108513/80368673-fe326e80-8884-11ea-994b-6ae5e32a5151.png)

IanMayo:
Closed in #413

PR #413, opened 2020-05-14
401 - adds a function to filter the allowable cells by cell-types

AxelleDRouge:
## 🧰 Issue
The available cells need to have travel restrictions (sea/land/air) [#401 ](https://github.com/serge-web/serge/issues/401)

## 🚀 Overview: 
I added a function to define the cell type, then use it to filter the list of cells allowable

## 🔗 Link to preview
[https://serge-review-feature-40-8kib4x.herokuapp.com/storybook/?path=/story/local-mapping--with-allowable-range](https://serge-review-feature-40-8kib4x.herokuapp.com/storybook/?path=/story/local-mapping--with-allowable-range)

## 🔨Work carried out:
- [x] Added a function to add the cell type to each cells in the grid
- [x] Added a variable to extract the cells from the available cells by type

- [x] Tests pass

## Confirmations


foxleigh81:
@AxelleDRouge Hi Axelle, is this set to [WIP] as it's not ready to merge yet? If so then can you convert it into a draft PR instead?

Impressed with what I've seen so far :)

Edit: Not sure if you converted it or if it was already a draft and my coffee deprive brain just didn't see it. Either way, thanks :)

AxelleDRouge:
@foxleigh81  It was already a draft PR

foxleigh81:
> @foxleigh81 It was already a draft PR

Yup, definitely need more coffee then.

packages/components/src/local/mapping/index.stories.tsx
```diff
@@ -53,25 +53,25 @@ const OSMTileLayer = {
  * DEFAULT VIEW
  */
 export const Default: React.FC = () => <Mapping
-  tileDiameterMins = {5}
```
IanMayo, line 56: @foxleigh81  - does @AxelleDRouge need a VS-Code settings file?  Or, should these formatting changes have been caught in the pre-push hook?

foxleigh81: Yeah, husky should have picked those up before it was pushed. @AxelleDRouge did you run `yarn install` in the repo root? If you only ran it inside the `components` package, you won't have everything you need.

IanMayo:
On first impressions it looks like the wrong cells are being allowed/not-allowed.  This is because the cell-types data-file was produced for a grid of a different size.  

This video shows the cells _are_ being blocked, but they appear to be in the wrong place over the map:
screencast

But, the video shows a shortcoming: the user should not be able to draw the line/route onto the non-allowed cells. I expect this is due to an instance where we're checking against `allowableCells`, rather than `filteredAllowableCell`.

packages/components/src/local/mapping/index.stories.tsx
AxelleDRouge: I think I did yes.
I'll try again to be sure

AxelleDRouge: It worked except for the export node command, but replaced by set for windows it works

AxelleDRouge:
Indeed, I correct this problem

packages/components/src/local/mapping/helpers/cell-types.ts
```diff
@@ -0,0 +1,17 @@
+import { SergeHex } from '@serge/custom-types'
```
IanMayo: This isn't a technical shortcoming, but having two files titled `cell-types.ts` feels scary.

Could we rename the helper function (& filename) to `cellTypeFor`?

Aah, I think I've also spotted a potential performance improvement.  I may be wrong, but it looks like we read/parse `cellTypesList` lots and lots of times.  But, it doesn't change.

Is it an option to load the data object once when `create-grid` runs, then pass the structure in to this function as a parameter?

AxelleDRouge: Yes inded those could be improved, I was trying to be consistent with the cellName function but it does create that shortcoming

AxelleDRouge: I changed the name and extracted the list in create-grid

IanMayo (approved):
Really impressive first piece of work, well done @AxelleDRouge

Issue #414, opened 2020-05-14
Data amends


foxleigh81:
## 💬 Brief description of feature

Parts of the data model passed via mapping needs to be updated to work with the new dialogue forms.

## ☑️ Acceptance criteria

- [ ] Forms inside `mapping` display correct data
- [ ] Stories updated
- [ ] Tests pass

IanMayo:
Done.

PR #416, opened 2020-05-14
Assets now display perceived data including name

foxleigh81:
## 🧰 Issue
#414 

## 🚀 Overview: 

The data passed to the forms in the `Mapping` component doesn't work correctly. This PR corrects both the sanitisation of data passed to the forms as well as updating the data model itself.

## 🤔 Reason: 
The forms work in their own stories but not within the `Mapping` component.

## 🔨Work carried out:

- [x] Updated data passing in `Mapping`
- [x] Updated the data model
- [x] Updated the stories.
- [x] Tests pass

## Confirmations


IanMayo:
I've found an issue. I think it's that there isn't a list of speeds for the RCB control.

screencast

foxleigh81:
> I've found an issue. I think it's that there isn't a list of speeds for the RCB control.
> 
> screencast

I have fixed this but in doing so I came across a larger issue.

Because the forms use their own internal state and are not yet connected to a wider state in `mapping` the forms cannot retain their updated value - always defaulting back to their initial state from mock data. At first, I wasn't too concerned about this as it was something that would fix itself when we start capturing state in mapping later on but I'm now concerned that this behaviour is hiding bugs, so this PR is going to be extended to include that mapping state work.

IanMayo:
One naming issue.  We have, for example, two objects called `AdjudicateTurnForm` - one is the react element, the other is the data passed to the element.

Could we rename the data interface type to `AdjudicateTurnFormData`?

foxleigh81:
> One naming issue. We have, for example, two objects called `AdjudicateTurnForm` - one is the react element, the other is the data passed to the element.
> 
> Could we rename the data interface type to `AdjudicateTurnFormData`?

Fixed, will be available on the next push.

IanMayo (approved):

PR #417, opened 2020-05-16
Feature/414 data amends move form logic

IanMayo:
Progresses #414
## 🧰 Issue

This PR is a trial in moving the logic for collating form data to the point immediately before the form is created.

In this way, we only create the data for the relevant form

## 🤔 Reason: 

While trying to debug an issue where the `perception` form wasn't getting the expected data, I tried to rationalise the code, as an experiment.

It seems successful in terms of simplifying the work that the `mapping` component has to do, and in reducing data in the `context` object.

## 🔨Work carried out:

* [x] Move collate form data code from mapping to map-bar
* [x] Refactor code out to helper functions

## Confirmations


IanMayo:
Aah, @foxleigh81 - this is ready for review.  

I moved a bunch of functionality out of mapping & mapping-context, & now just collate the form data immediately before the form opens.

packages/components/src/local/map-bar/helpers/collate-adjudication-form-data.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+import collateAdjudicationFormData from './collate-adjudication-form-data'
+
+import platformTypes from '@serge/mocks/platform-types.mock'
+import selectedAsset from '@serge/mocks/selected-asset.mock'
+import forces from '@serge/mocks/forces.mock'
+import { AdjudicateTurnFormPopulate, AdjudicateTurnFormValues } from '@serge/custom-types'
+
+it('contains relevant population results', () => {
+  const res: AdjudicateTurnFormPopulate = collateAdjudicationFormData(platformTypes, selectedAsset, forces).populate
+  expect(res.condition.length).toEqual(5)
```
foxleigh81: `const { condition, speed, status, visibleTo } = res`

would be preferred here and again in the value section.

packages/components/src/local/map-bar/index.tsx
```diff
@@ -3,6 +3,10 @@ import cx from 'classnames'
 import { ArrowRight } from '@material-ui/icons'
 import assetDialogFor from './helpers/asset-dialog-for'
 
+import collatePlanFormData from './helpers/collate-plan-form-data'
+import collateAdjudicationFormData from './helpers/collate-adjudication-form-data'
+import collatePerceptionFormData from './helpers/collate-perception-form-data'
+
```
foxleigh81, line 9: Now we are pulling them in this way, it strikes me that those functions are very similar and could probably be refactored into a single function.

foxleigh81 (approved):

packages/components/src/local/map-bar/helpers/collate-adjudication-form-data.spec.ts
foxleigh81: There are a few other examples of this in other files too.

IanMayo: There's a _cost_ of adding an extra line of code.  But, can you explain the benefit?

(I recognise the above would reduce lines if 4 consts were being declared, but, it's not reducing lines in this instance)

packages/components/src/local/map-bar/index.tsx
IanMayo: Hmm, they share the `populate` and `values` general data structure, which I like. But the contents of those structures are quite different to each other.

If we change the structure of, say, the perception form, then we would expect `collatePerceptionFormData()` to be modified, but it would be more challenging to check for -ve consequences if we were modifying a single super-function, IMHO.

packages/components/src/local/map-bar/helpers/collate-adjudication-form-data.spec.ts
foxleigh81: It's a code-style thing. Our code style is that we use destructuring, it makes code easier to read, it reduces the possibility of errors (typos for example) and in some instances (although not this one) it caches values and therefore reduces calls to external data.

Looking at that again, you could actually not bother with `res` at all and you could destructure from the `collageAdjudicationFormData` directly.

I approved the PR as it's not something you HAVE to do but if you want to stick to a consistent coding-style, then it's something you should do.

packages/components/src/local/map-bar/index.tsx
foxleigh81: Fair argument, I'm happy to leave it.

packages/components/src/local/map-bar/helpers/collate-adjudication-form-data.spec.ts
IanMayo: I have dereferenced the results from the method. The TS typing bits are a bit cumbersome: https://github.com/serge-web/serge/pull/417/commits/f1dbb9938d78a22eabb5893742df591f80feb7cf#diff-dcb8674751fb00c2ffec04cd29fae785R11

PR #421, opened 2020-05-19
Feature/mapping test

foxleigh81:
## 🚀 Overview: 
Puts the `Mapping` component from the `components` package into the `client` pacakge.

## 🤔 Reason: 
To make sure everything works as expected.

## Link to preview:
https://serge-review-feature-ma-t1rwsv.herokuapp.com/?wargame=wargame-kaf9gvb3&access=p5543

## 🔨Work carried out:

- [x] Updates the `config` and `helpers` packages to build to js so they can be used as part of the wider `serge` package ecosystem without relying on TS.
- [x] Updates styling on the `mapping` component to be less storybook specific.
- [x] Adds a decorator to the `mapping` component stories to restore the required styling for storybook
- [x] Imports the `mapping` component into the `client` package and test ran it.
- [ ] Changing the perceived force of an asset is reflected on the map 
- [x] Tests pass

## 🖥️ Screenshot
Screenshot 2020-05-19 at 14 01 54

## Confirmations

## 📝 Developer Notes:
- This just imports the blank map, however it works as expected with the existing data. Importing the child components will probably be where the data mismatch becomes an issue.


IanMayo:
@foxleigh81 - I've added a test to the description. It uses `change perception` to verify the server round-trip.   We can send this message without all the _plan route_ functionality that I still have to implement.

We can trigger the "round-trip" to send the perceived update to the backend by reproducing the `perceivedStateCallback` method in `client\components\mapping\index.jsx`.  The perceived form just needs to re-package the `PerceivedDataValues` into the structure used in `perceivedStateCallback`, and fire it into a `sendMessage()` method.

Aah, there's a point. The `mapping` component could do with the `sendMessage` functionality that's in the legacy mapping component. It wraps the message in a uniform container, then sends it to the backend processing. 

So, I think the new mapping component should no longer pass the `postback` directly to the context.  It should wrap that function in the above `sendMessage` function, and pass the `sendMessage` to the mapping context.

foxleigh81:
> @foxleigh81 - I've added a test to the description. It uses `change perception` to verify the server round-trip. We can send this message without all the _plan route_ functionality that I still have to implement.
> 
> We can trigger the "round-trip" to send the perceived update to the backend by reproducing the `perceivedStateCallback` method in `client\components\mapping\index.jsx`. The perceived form just needs to re-package the `PerceivedDataValues` into the structure used in `perceivedStateCallback`, and fire it into a `sendMessage()` method.
> 
> Aah, there's a point. The `mapping` component could do with the `sendMessage` functionality that's in the legacy mapping component. It wraps the message in a uniform container, then sends it to the backend processing.
> 
> So, I think the new mapping component should no longer pass the `postback` directly to the context. It should wrap that function in the above `sendMessage` function, and pass the `sendMessage` to the mapping context.

Cool. You put it in the wrong place, but I've moved it :)

IanMayo:
Great, this is really exciting.

The data in the perception form isn't editable.  I've added a 'preview' link to the issue description that goes straight in as a Blue player - so it's quick & easy to get to a point where perception can be tested.

Issue #333, opened 2020-03-24
Refactor game admin channel component into component library


foxleigh81:
* [ ] Pull Serge GameAdmin styles into StoryBook component
* [ ] User StoryBook GameAdmin component in main Serge app

IanMayo:
@foxleigh81 provided a first version of this control in PR #429 

But, it lacks Serge styling.  See here: https://serge-review.herokuapp.com/storybook/?path=/story/local-gameadmin--default

@Lilit - are you able to think of a way to pull the Serge styles into StoryBook?  

There are a couple of instances where the storybook styling is different to Serge, which makes things more difficult to review.

But, if this isn't easy - just say, and we can work around it.

IanMayo:
Aah, @lilitkarapetyan - I've found another remaining part of this issue.

The Serge client is still using the legacy (original) Game Admin control, not the new component.

Could you switch the Serge client to use the new Component please?

IanMayo:
Handled separately.

PR #429, opened 2020-05-27
Feature/game admin

foxleigh81:
## 🧰 Issue

Supports #333

## 🚀 Overview: 
Converts the `GameAdmin` component to the new format and includes a Storybook story

## 🔗 Link to preview

https://serge-review-feature-ga-1lpfdo.herokuapp.com/storybook/?path=/story/local-gameadmin--default

## 🤔 Reason: 
An example of how a component which isn't as complex as `mapping` is needed as a template for future conversions.

## 🔨Work carried out:

- [x] created new `store` package
- [x] moved 'api' to store
- [x] moved 'schemas' to store
- [x] moved 'actionsAndReducers' to store
- [x] Split 'actionsAndReducers' into 'actions' and 'reducers'
- [x] Moved helpers into package root'
- [x] Moved 'actionConstants' into package root and renamed to 'constants'
- [x] Installed TS
- [x] created new src folder and moved all relevant files into there
- [x] refactored exist 'store' codebase to work with new structure
converted player-ui and schemas to TypeScript
- [x] Refactored player-ui to work as a context provider to the components package
- [x] Moved helpers from the `clients` package into the `helpers` package
- [x] Converted imported helpers into TS
- [x] Made an update to the develop guidelines to include rules for source code ordering
- [x] Built `MessagesListRenderProp` component
- [x] Built `MessagesListChatChannel` component
- [x] Built `MessagesCreatorChatChannel` component
- [x] Built `GameAdmin` component
- [x] Built all required stories
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:

- As you can see, a lot of ancilliary work was needed to get the components to work well in the new format. This has resulted in the creation of the new `store` package which can replace the functionality of the existing `api`, `ActionsAndReducers`, `Schemas` and `Store` folders in the `Client`package. This isn't a drop-in replacement however and a refactoring of the client package to use these components will be needed however this should be a straightforward task. Do not delete the folders in the clients package until this refactoring has been done.
- As my time is limited, I only made TS conversions where I had to. I have also just used 'any' if the type definition wasn't immediately
apparent. This will need to be fixed in a future task.
- Integration with the `client` package is still needed. The new `store` package should be in a usable state already and it would be preferable if any views that get refactored to work with the new components are refactored to use the `store` package instead of the stuff in the `client` package. Also Pure function rules must apply. ONLY the views can access the `store` all components must receive their dynamic data via props.


IanMayo:
After a message is sent, the text box should be cleared.

IanMayo:
Is is possible to pull in Serge styling?  Or isn't that the idea in StoryBook.  Aah, maybe we style the component here in StoryBook, then it goes into Serge with that same styling.

foxleigh81:
> Is is possible to pull in Serge styling? Or isn't that the idea in StoryBook. Aah, maybe we style the component here in StoryBook, then it goes into Serge with that same styling.

Indeed, this is the idea. You will need to hire a SCSS person so that they can work on turning the `themes` package into a global styling solution. Once this is done, you'll also want to pull those global styles into Storybook so that you can be sure that the styling in Storybook and in the client package is identical.

foxleigh81:
> After a message is sent, the text box should be cleared.

Done.

packages/components/src/local/game-admin/styles.module.scss
```diff
@@ -0,0 +1,17 @@
+// $red: red;
```
IanMayo, line 1: @foxleigh81  - did you intend to leave these commented out?

IanMayo (approved):

PR #430, opened 2020-05-29
Hotfix/investigate cypress fail

IanMayo:
Test branch to see where Cypress is breaking


IanMayo:
Drop this. Cypress appears to be running inconsistently.

PR #434, opened 2020-06-01
Hotfix/investigate cypress fail 3

IanMayo:
Third attempt to investigate Cypress issue


IanMayo:
Drop this. Cypress appears to be behaving inconsistently

PR #435, opened 2020-06-05
Hotfix/routes and allowable areas

IanMayo:
Hotfix to resolve missing route markers & allowable routes.


packages/components/src/local/asset-icon/index.tsx
```diff
@@ -36,17 +36,24 @@ export const AssetIcon: React.FC<PropTypes> = ({
...
-      type,
-      force,
-      visibleTo,
-      controlledBy,
-      condition,
-      status
-    })
-    setShowMapBar(true)
+    if (selectedAsset && selectedAsset.uniqid === uniqid) {
+      // ok, it's the current asset. clear it
```
foxleigh81: Remember that I said comments shouldn't be conversational. This should be.

// Clear the current asset.

packages/components/src/local/mapping/index.tsx
```diff
@@ -210,6 +190,7 @@ export const Mapping: React.FC<PropTypes> = ({
           origin: lastCell.name,
           travelMode: planningConstraints.travelMode
         }
+        console.log('new constraints', newP)
```
foxleigh81: Did you mean to leave this in?

packages/components/src/local/mapping/index.tsx
```diff
@@ -252,7 +233,9 @@ export const Mapping: React.FC<PropTypes> = ({
         for (let ctr = 0; ctr < plannedTurn.turnsVal; ctr++) {
           const step: RouteStep = { turn: ++turnStart, status: { state: status.name } }
           // store this step
-          store = routeAddStep(store, selectedAsset.uniqid, step)
+          if (selectedAsset) {
+            store = routeAddStep(store, selectedAsset.uniqid, step)
```
foxleigh81: Be careful with circular references like this.  You might be safer doing `updatedStore = routeAddStep(etc...)` obviously don't forget to declare `updateStore` before you reference it.

foxleigh81:
```suggestion
           // before conditional 
           let updatedStore
           
           // in conditional
            updatedStore = routeAddStep(store, selectedAsset.uniqid, step)
            
            // afterConditional
            setRouteStore(updatedStore)
```

IanMayo: In the loop, we need to add the new step to the modified store:

 updatedStore = routeAddStep(***updatedStore***, selectedAsset.uniqid, step)

foxleigh81: Leave it as it is then if that's the way it has to be. Maybe rename the variable though as there is a constant with the same name earlier in the file and it was momentarily confusing.

PR #497, opened 2020-09-12
Amend/495 refactor message component

### LINKED ISSUE #495: Get theme/parent-styling working for StoryBook elements
We're creating new elements as standalone components that can be demonstrated/tested within StoryBook.

But, components don't get the parent styling.

Here is the "Game Admin" chat-channel:
image

But this is how it appears in StoryBook:
image

It is possible to direct StoryBook to include parent styling, as demonstrated in line 77 here:
https://github.com/foxleigh81/alex-foxleigh-portfolio/blob/2020/packages/components/.storybook/main.js

That bit of code adds the variables as globally accessible variables in the storybook project, which means any file that is referenced in there also has access to those variables. It wouldn't be visible to the .stories.tsx file but it would be visible to the scss file.

An example of apply this approach to themes is given in the `2020` branch of that repo, here:
https://github.com/foxleigh81/alex-foxleigh-portfolio/tree/2020

Successful completion of this issue would be for the Storybook element for `Game-Admin` to look like the one in the actual game.

zukilover:
## 🧰 Issue
Closes https://github.com/serge-web/serge/issues/495

## 🚀 Overview: 
This fixes the message styling in the Game Admin storybook. 

## 🔗 Link to preview

App: https://serge-review-amend-495--orf2s0.herokuapp.com/

Storybook: https://serge-review-amend-495--orf2s0.herokuapp.com/storybook/?path=/story/local-pages-gameadmin--default

## 🤔 Reason: 
We're creating new elements as standalone components that can be demonstrated/tested within StoryBook.
But, components don't get the parent styling.

## 🔨Work carried out:
- [x] Create AdminMessagesList component
- [x] Create AdminMessage component
- [x] Update GameAdmin story
- [x] Tests pass

## 🖥️ Screenshot
screenshot-localhost_6611-2020 09 12-17_57_30

## Confirmations

## 📝 Developer Notes:
In the old app structure, messages list and message component are shared between the Channel and Game Admin sections.
Upon review, I've found that even though they have common class names, they should be separated into two different molecules components. This is because, those within out-of-game (game admin) panel, most of the shared classes are actually overridden. 
I have developed separate component sets, admin-messages-list and admin-message for the game admin panel, and leave out the messages-list and message to be used by the channel (in-game) panel.


IanMayo:
away from computer right now, but that all sounds great.

packages/components/src/local/admin-message/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import PropTypes from './types/props'
+
+/* Import Stylesheet */
+import styles from './styles.module.scss'
+
+/* Render component */
+export const AdminMessage: React.FC<PropTypes> = ({ message }: PropTypes) => {
+  const { details } = message
+  const { force, role, name } = details.from
+  const isOwnMessage = name === force
```
IanMayo: @zukilover - I _think_ the above logic may be wrong.  I _think_ we need to compare the force that sent it to the force of the logged in player.   This could be passed into the component in props, since parent components will know it.  

So, I guess you could pass "Blue" into the Story definition.

IanMayo:
@zukilover - I really like what you've done here. The stories are both good, simple tests.

Hmm, should `Mark all as read` work? Or is it dependent upon a server-side call?

zukilover:
> Hmm, should `Mark all as read` work? Or is it dependent upon a server-side call?

It has not been refactored yet. I'm doing it now. In the storybook, I will add a 'page' level story to show how it works in the real case.

packages/components/src/local/molecules/admin-message/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import PropTypes from './types/props'
+
+/* Import Stylesheet */
+import styles from './styles.module.scss'
+
+/* Render component */
+export const AdminMessage: React.FC<PropTypes> = ({ message }: PropTypes) => {
+  const { details } = message
+  const { force, role, name } = details.from
+  const isOwnMessage = name === force
```
IanMayo: @zukilover - I may be mistaken, but I think we should be matching the force of the logged in player.  I guess we can pass that in as props.  For the Story, I guess we could have checkbox knobs to allow switching between `White` and `Blue` force, which will switch which messages are _own_.

zukilover: You're correct. I must've missed on that own message checking logic. I'll update soon.

IanMayo: After almost a year away from the project, you are completely forgiven 🤣

zukilover: Updated now :D

packages/components/src/local/molecules/admin-message/types/props.d.ts
```diff
@@ -0,0 +1,16 @@
+import { Message } from '@serge/custom-types'
+
+export default interface PropTypes {
+  /**
+   * The message content
+   */
+  message: Message
+  /**
+   * Current force
+   */
```
IanMayo, line 10: @zukilover  - I may be mistaken, but it looks like only `force.name` is used from the `force` component.  I suspect we could just use: `force: string`, and use the string directly in the admin-message element.

IanMayo: I've just fixed this here: https://github.com/serge-web/serge/pull/497/commits/76dd3e1e3e60cbd783de74e83e2a55f79badec78

packages/components/src/local/molecules/admin-message-creator/index.tsx
```diff
@@ -43,7 +43,7 @@ export const AdminMessageCreator: React.FC<Props> = ({ from, channel, role, post
   }
 
   return (
-    <Box className={styles['message-creator__wrap']}>
+    <Box className={styles['message-creator-wrap']}>
```
IanMayo, line 46: @zukilover - the linter rejected the use of `_` in the class name.  It all seems to work ok, but are you happy with the edits in this commit?

zukilover: it works for me. thanks.

packages/components/src/local/molecules/admin-message/types/props.d.ts
zukilover: looks good to me, you're a hero!

IanMayo (approved):
Good work, thanks.

PR #511, opened 2020-09-28
494 - only allow Assets to be combined in certain situations

### LINKED ISSUE #494: Task Groups - only allow Assets to be combined in certain situations
Assets can only be combined into composite units under certain circumstances (such as being in the same cell).

So - we need to check if two assets can be combined.  I can produce the logic for this, but I need some help in calling that logic.

We don't know this before drag starts, since it depends on what the asset is being dragged over.

So, we should call a function to check if we need to create the `dropZone` element(s).

I can't find where to put this logic, though it may be in `groups\index.tsx`

But it can be solved with a call to a function with this API:
```
   function canCombineWith(forces: any, draggingId: string, underId: string) : boolean
```
The initial _fake_ implementation would be:
```
{
  // fake implementation. Only allow dropping if its onto a name that's earlier in the alphabet
  dragging = findAsset(forces, draggingId)
  under = findAsset(forces, draggingId)
  return dragging.name > under.name
}
```

lilitkarapetyan:
## 🧰 Issue

closes #494

## 🚀 Overview: 
new function `canCombineWith` for `Groups` Component
new property `disable` for `Dropzone` Component

## 🤔 Reason: 

`canCombineWith?: (draggingItem: GroupItem, item: GroupItem, parents: Array<GroupItem>, type: NodeType) => boolean`

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

изображение
изображение
изображение

## Confirmations


IanMayo:
Thanks @lilitkarapetyan .

It stops the `MPA` being dropped onto the `Frigate`.  But, the space still appears underneath `Frigate` and if I drop `MPA` onto it, `MPA` becomes a new group with `Merlin`

screencast

IanMayo:
This may be related, I don't know.

As we drag `MPA` over `MERLIN`, `MERLIN` gets renamed to `MPA`.

screencast

IanMayo (approved):
Thanks to @lilitkarapetyan  for doing the skeleton work.

PR #521, opened 2020-10-08
Feature/508 submit plans

### LINKED ISSUE #504: Refactor admin pages to components
While the `Platform Types` page is being written as components, the other legacy pages aren't.

They should be refactored into components, testable using StoryBook

### LINKED ISSUE #508: Handle transmission of planned routes
Once a player has finished planning for their force they press "Send".

This operation needs to:
* collate data into suitable format for sending
* pull new copy of wargame
* update wargame with player's planned routes
* transmit wargame

Once the planned routes has been sent, the player should not be allowed to perform any more planning.

The last step will result in a new wargame being received by the mapping control, which _may_ trigger a map update.

IanMayo:
## 🧰 Issue

Fixes #508

## 🚀 Overview: 

Implemented full cycle for submitting plans

## 🔗 Link to preview

https://serge-review-feature-50-wuubvr.herokuapp.com/storybook/?path=/story/local-mapping--with-assets

## 🔨Work carried out:
- [x] Centrally track if a set of plans have been submitted
- [x] Disable planning forms once plans submitted
- [x] Re-allow planning once new phase
- [x] Include message in planning form to explain why disabled
- [x] Include message in adjudication form to explain why disabled
- [x] Test within game 

- [x] Tests pass

## Confirmations


packages/components/src/local/adjudicate-turn-form/index.tsx
```diff
@@ -95,6 +95,10 @@ export const AdjudicateTurnForm: React.FC<PropTypes> = ({ formHeader, formData,
         platformType={icon.platformType}
       >
         {formHeader}
+        { plansSubmitted &&
+         <h5 className='sub-title'>(Form disabled, plans submitted)</h5>
```
foxleigh81, line 99: Why an `h5` here just out of curiosity?

packages/components/src/local/plan-turn-form/index.stories.tsx
```diff
@@ -58,6 +58,35 @@ export const Default: React.FC = () => <PlanTurnForm
     }
   }}
   icon={iconData}
+  plansSubmitted={false}
+/>
+
+export const TurnsPlanned: React.FC = () => <PlanTurnForm
+  turnPlanned={postback}
+  formHeader="Planning header"
+  formData={{
```
foxleigh81, line 67: The spec file uses a `formData` mock but this does not? Is the content different in the story? If so, why?

foxleigh81 (commented):
I've added a few comments. There may be valid reasons but I'd want to know those reasons before approving.

packages/components/src/local/adjudicate-turn-form/index.tsx
IanMayo: It was to make the text smaller.

packages/components/src/local/plan-turn-form/index.stories.tsx
IanMayo: Aah, it's so we can use `knobs` to vary the speed.

packages/components/src/local/adjudicate-turn-form/index.tsx
foxleigh81: So `h1`, `h2`, `h3` etc.... have nothing to do with font size. That's a common misconception. They are there to represent the hierarchy. So an `h1` for example is the most important part of that section/page and a `h5` is not very important at all.  See this article for more information: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements

I'd recommend choosing a more appropriate header level and then using CSS to change the font-size.

packages/components/src/local/plan-turn-form/index.stories.tsx
foxleigh81: That's fair enough. I think you'd still be able to do that without repeating code but without looking into it, I'm not 100% sure about that.

packages/components/src/local/adjudicate-turn-form/index.tsx
IanMayo: Ok, here's my thinking.  I don't know why, but the `formHeader` is an `h2`.  I don't mind actual size, but I wanted it less dominant than `h2`  - so `h5` worked. Here's what `inspect` shows:
image

foxleigh81: In that case, a `h3` would be appropriate :) In fact, as that lives inside a `h2` tag, you don't need it to be a header at all. You can just make it a span and use a class to set the size.

PR #528, opened 2020-10-12
Feature/504 refactor admin pages to components

### LINKED ISSUE #504: Refactor admin pages to components

zukilover:
## 🧰 Issue

Closes #504 

## 🚀 Overview: 

Refactor admin pages into components

## 🤔 Reason: 

While the Platform Types page is being written as components, the other legacy pages aren't.
They should be refactored into components, testable using StoryBook

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/components/src/local/molecules/option-menu/index.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const Default: React.FC = () => {
+  const menus = [
+    {
+      onClick: (): void => window.alert('Click on menu #1'),
+      icon: faBath,
+      title: 'Save clean copy of wargame (no messages)',
+      label: 'Clean copy'
+    },
+    {
+      onClick: (): void => window.alert('Click on menu #2'),
```
foxleigh81: The `menus` constant is duplicated in the spec file. It would be better to move this to an external file and reference it in both. That way we are not repeating code.

foxleigh81 (dismissed):
One small change and then am happy for this to be approved.

packages/components/src/local/atoms/game-title/index.stories.tsx
```diff
@@ -1,13 +1,26 @@
 import React from 'react'
+import { styled } from '@material-ui/core/styles'
+import Box from '@material-ui/core/Box'
 
 // Import component files
 import GameTitle from './index'
 import docs from './README.md'
```
IanMayo, line 6: @zukilover - I _think_ we use `BlueContainer` in a number of stories.  Should we define a single instance somewhere centrally?   That would make it easier to update if the higher level styling changes.

IanMayo:
Thanks for the new work @zukilover 

I've pushed a change that makes the WargameList scroll-bar have `auto` rather than permanently on.

I've also pushed a story that includes more wargames, so the scroll bar gets used.

But, the colors on the scroll bar don't look great.  I'm not sure what the fix is, but could you have a go at it?

Screen Shot 2020-10-14 at 11 49 46

IanMayo:
Oh, and @zukilover - the new `DeepScan` checking seems to have highlighted two potential (but unlikely) issues.

Could you include them in your fixes please?

IanMayo:
One (very minor) observation, and a learning opportunity for me. Is `<SearchList>` really an organism?  It seems simpler, like a molecule.  I'm keen to learn :-)

zukilover:
> One (very minor) observation, and a learning opportunity for me. Is `<SearchList>` really an organism? It seems simpler, like a molecule. I'm keen to learn :-)

You're 100% correct. As a rule of thumb: if it breaks down into atomic components, it is a molecule. 
I missed this one, because I have been moving the `SearchListRow` component back from molecules to atoms. The `SearchListRow` contains these duplicate and delete action buttons that made me think it should go to molecules. Since it is mostly composed on basic HTML, I moved it back to atoms.

I'll move the `SearchList` component into molecules.

IanMayo:
@zukilover - now that we have the AdminLogin as a component, could you modify it so that the login happens when pressing the `Enter` key while the focus is in the text field?

IanMayo:
@zukilover  - in "Add role" it looks like you're getting very close to @lilitkarapetyan 's work in this PR, have a look at the screenshots:
https://github.com/serge-web/serge/pull/546

I'll try to get her PR merged this morning.

packages/components/src/local/molecules/option-menu/index.stories.tsx
zukilover: Added mocks for menus

packages/components/src/local/atoms/game-title/index.stories.tsx
zukilover: Blue container moved to global preview configuration

zukilover:
> @zukilover - now that we have the AdminLogin as a component, could you modify it so that the login happens when pressing the `Enter` key while the focus is in the text field?

Submit on enter added.

zukilover:
> Oh, and @zukilover - the new `DeepScan` checking seems to have highlighted two potential (but unlikely) issues.
> 
> Could you include them in your fixes please?

Deepscan issues fixed.

zukilover:
> Thanks for the new work @zukilover
> 
> I've pushed a change that makes the WargameList scroll-bar have `auto` rather than permanently on.
> 
> I've also pushed a story that includes more wargames, so the scroll bar gets used.
> 
> But, the colors on the scroll bar don't look great. I'm not sure what the fix is, but could you have a go at it?

I have added custom scrollbar via `react-scrollbar`

IanMayo:
@zukilover  - I'm sure this is just cosmetic, and probably only happens in StoryBook, but the TextInput text isn't visible:
image

zukilover:
> @zukilover - I'm sure this is just cosmetic, and probably only happens in StoryBook, but the TextInput text isn't visible:

I have added customColor property for that. Pls check.

IanMayo (approved):
Huge piece of work, thanks @zukilover

PR #536, opened 2020-10-16
Channel message bubble

### LINKED ISSUE #502: Simple UI for plain chat message
Some channels will have message editors supporting complex data objects.

But, for some channels, they will just work like a simple chat channel. In these instances the original design was for the chat writing text element to be permanently present, with a `Send` button.

image

The message-type dropdown & collapsible writing area would only be present if the channel supported any non-chat message types for the current force/player.

Mpre detail on UI here:
https://www.figma.com/file/2qiEitO08s8jXamHWI3VpP/SERGE-mapping-tool?node-id=964%3A44

I guess this is the sequence of tasks in this issue:
* [x] Storybook component for default part of message
* [x] Storybook component for private part of message
* [x] Storybook component for composite message (both of the above).  Test page should include knobs that let reviewer test perspective from White or Blue force.  Messages aligned to right for "own" messages, and to left for "others".  Plus, private message only displayed for white force.
* [x] Storybook component for default text-entry fields
* [x] Storybook component for private text-entry fields
* [x] Storybook component for  composite write-message panel (containing both of above)
* [x] Storybook component for whole channel (few messages, plus text-entry fields). This should have Storybook `knobs` that let tester review layout for both White-force player and Blue-force player (since private field only visible to white force)
* [x] Introduce logic to determine whether to use `chat channel` or traditional (existing) complex channel
* [x] Introduce new component into Game UI. 
* [ ] Test

Some of the minor elements required for these components will already be present in Storybook.  Please re-use them when applicable.

Note: when creating storybook components, please follow the Atomic Design Principles in organising them: https://bradfrost.com/blog/post/atomic-web-design/

IanMayo:
## 🧰 Issue

Fixes #502

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Feedback: the two binary wargame files shouldn't have been committed to the branch. Please delete them.

IanMayo:
This all looks fine so far.

IanMayo:
Aah, one change, please.  There may be some confusion over the `channel` components:
image

Could you rename the component to `ChatMessageBubble` please?  (don't worry about renaming the branch though).

IanMayo:
Thanks. I think there may be better logic for which side the message appears on.

The messages on the left have a vertical colored bar with the force for who sent it.  

So, if the control received a string containing the force-id of the current player, the force-id of who sent it, and an array of force-colors (force-id and CSS color components), then the control could first compare current player force to who sent it (to decide which side to put it on), and then use the force-colors array to retrieve the color to use for the highlight bar.

Note: in `@serge/mocks` there's a `forces.mock.ts` and a `game-messages.mock.ts`.

You can access this mock data with code like this: `import { forces, platformTypes } from '@serge/mocks'`

IanMayo:
Note: have refactored method to produce the array of forces & colors.  It's available in `develop`.  You can see how its used here: https://github.com/serge-web/serge/pull/541

IanMayo:
Hmm, we already have a `text-input` component:
https://serge-review-channel-me-gx8bun.herokuapp.com/storybook/?path=/story/local-atoms-textinput--filled

But, I do think there is a reason for the new component.  The Chat UI has light coloured boxes:
image

So, you could create a `chat-text-input` component that takes the new styling.  Note: the existing `Chat Message Bubble` should also be updated to the new styling.

loveheenavasu:
> chat-text-input

okay I'll make changes accordingly

IanMayo:
Thanks for the update.  Sorry, but I forgot to tell you about a UX update.

The UX now has an "attachment" icon.  Clicking on this icon expands/collapses the "Private" message block.   It's only a small change.

We made the change since it having the "Send" next to the main text box didn't really help the understanding that the Private message would get sent, too.  Calling it a private attachment makes it more obvious.

Here's the updated UI: https://www.figma.com/file/2qiEitO08s8jXamHWI3VpP/SERGE-mapping-tool?node-id=961%3A8

loveheenavasu:
> Thanks for the update. Sorry, but I forgot to tell you about a UX update.
> 
> The UX now has an "attachment" icon. Clicking on this icon expands/collapses the "Private" message block. It's only a small change.
> 
> We made the change since it having the "Send" next to the main text box didn't really help the understanding that the Private message would get sent, too. Calling it a private attachment makes it more obvious.
> 
> Here's the updated UI: https://www.figma.com/file/2qiEitO08s8jXamHWI3VpP/SERGE-mapping-tool?node-id=961%3A8

okay I'll make chages accordingly

packages/components/src/local/molecules/chat-channel-message/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  Red: 'Red'
+}
+
+/* Render component */
+export const ChatChannelMessage: React.FC<Props> = () => {
+  return (
+    <div
+      className={`${styles['chat-channel-message']}`}
+    >
+      <ChatMessageBubble
```
IanMayo: data is being hard-coded into this component. It has to come in through props.

Again, Storybook `knobs` are being used to control state. We can't have `knobs` in a production component.

The props should be configured like in `index.stories.tsx` in `Channel Message Detail`.

loveheenavasu: okay I'll make changes accordingly

packages/client/src/Components/Channel.jsx
```diff
@@ -103,6 +103,8 @@ class Channel extends Component {
             role={state.selectedRole}
             timestamp="2020-09-18T05:41:17.349Z"
             title="lorem ipsum do lor sit amet lorem ipsum do lor sit amet lorem ipsum do lor sit amet lorem ipsum do lor sit amet lorem ipsum do lor sit amet lorem ipsum do lor sit amet"
+            privateMessage="This is an example of a private message."
```
IanMayo: you should not be hard-coding data into the actual component.  Data objects like this should be either provided in tests, or taken from the database.

Note that the existing code (from line 108 onwards) uses a `ChannelMessagesList` element to render the series of `messages` entries.

You should probably either create a `ChatMessagesList` element, or just map the `messages` items to our `ChatChannelMessage`.

IanMayo:
Note: I'm updated the second (lower) wargame so that Blue is now a member.

1.  Messages sent from another force should be shown on the left, and messages from own force on the right.
2. When logged in as White/Game Control, when I type into the `Private` box, the box loses focus after each key press, and I have to click back in the box before I can type the next letter.  Video below.

screencast

IanMayo:
Note: I've had to merge my branch into yours - in order to fix an emergent bug in a library we're using.

IanMayo:
One other thing.  All the messages are being shown as from "White".  Some are actually from "Blue":
image

packages/components/src/local/molecules/chat-message-bubble/types/props.d.ts
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Private message
+   */
+  privateMessage?: string
+  /**
+   * Show Private message
+   */
+  isUmpire?: boolean
+  /**
+   * Chat Message Position
```
IanMayo, line 39: messagePosition needs a better comment (and possibly name) - to explain what it's for

packages/components/src/local/molecules/chat-message-bubble/types/props.d.ts
```diff
@@ -0,0 +1,47 @@
...
+  timestamp?: string
+  /**
+   * Message sender role
+   */
+  role?: string
+  /**
+   * Message read state
+   */
+  hasBeenRead?: boolean
+  /**
```
IanMayo, line 18: if we're not using authorForceID and playerForceId properties they should be deleted.

IanMayo:
Drop. Development happened in other branch.

PR #553, opened 2020-10-28
Feature/78 Refactor template references

### LINKED ISSUE #78: Don't store templates in channels
We currently store the form templates in the channel descriptions.  

But, changes to the template aren't reflected in the game.

We should store a URL (or ID) for the template in the channel definition, which would allow for the template to be updated during the game design process.

[this change introduces risk - and should be deferred until at least late Dec 2019]

Note: there is value in the existing wargames.  _ideally_ a new implementation would allow for the legacy wargames to contain the templates, but newly created wargames should contain template IDs.  Ooh, how about: making a change to a channel replaces the template with the ID.

zukilover:
## 🧰 Issue
Closes #78 

## 🚀 Overview: 
The whole template object is currently stored in the database when the user creates/updates a channel.
The new implementation, stores only the template `_id` so that the player interface will be able to refer to the correct template with its latest structures.
This is using transitional approach where it can still read a channel that still uses a template object value on its structure.

## 🤔 Reason: 
We currently store the form templates in the channel descriptions.
But, changes to the template aren't reflected in the game.

## 🔨Work carried out:

- [x] Refactor channel table
- [x] Refactor edit subscription row
- [x] Refactor player ui reducer
- [x] Refactor new message component
- [x] Tests pass

## Confirmations

### DIFF
--- a/packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.js
+++ b/packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.js
@@ -85,7 +85,9 @@ export const playerUiReducer = (state = initialState, action) => {
       if (chosenTemplates.length === 0) {
         templates = newState.allTemplates.filter((template) => template.title === 'Chat')
       } else {
-        templates = chosenTemplates.map((template) => template.value)
+        templates = chosenTemplates.map(({ value }) => {
+          return typeof value === 'string' ? newState.allTemplates.find(item => item._id === value) : value
+        })
       }
     }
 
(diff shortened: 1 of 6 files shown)


IanMayo:
@zukilover - I've been trying to play with this.

I tried editing a message template, but I think that's currently broken :-(

Anyway, I tried to make a change to a template, then saved it. The admin page then threw this error:
image

Maybe I forced the app into an inconsistent state, but it would neat to handle the error.

This fix seemed to work:
```
      const matchedName = allMessages.find((messageType) => messageType.title && schema.title && messageType.title.toLowerCase() === schema.title.toLowerCase())
```

IanMayo:
Tested template reference.  I changed the templates in a channel (`Blue HQ`), and inspection of the database shows that template references are being stored for that channel, not the template itself:
image

IanMayo (approved):

PR #588, opened 2021-01-23
[SERGE-586] Upgrade mapping stories

### LINKED ISSUE #586: Move Mapping stories to StoryBook 6
My next task is to implement allowing players to move the start location of some of their assets during Turn-Zero.

It would be useful to do this via the `Mapping - with Assets` Story.  But, it's not currently working:
image

This issue is to update the `Mapping \ index.stories` file to StoryBook 6.

zukilover:
## 🧰 Issue

Closes #586 

## 🚀 Overview: 

Upgrade mapping stories to comply with Storybook 6 standards

## 🤔 Reason: 

Current story implementation is outdated and doesn't follow Storybook 6 CSF

## 🔨Work carried out:

- [x] Tests pass
- [x] Update mapping stories
- [x] Update mapping related component specs (assets, hex-grid)
- [x] Update world state stories template (using function clone: `Template.bind({})`)
- [x] Fix player login component (the text input was missing `updateState` prop)

## Confirmations

## 📝 Developer Notes:
Since storybook `args` usually needs `Template.bind({})`, it will break deepscan `USELESS_ARROW_FUNC_BIND` rule. I recommend disabling such rule to pass the pipeline.


IanMayo (changes requested):
Great work, thanks.

DeepScan has complained about something.  The construct occurs 10 times, but it's really just a single issue:
image

Can we fix it, or do we need to suppress the warning?

IanMayo:
Have posted StackOverflow question to try to resolve above issue - I couldn't sort it:
https://stackoverflow.com/questions/65883518/how-to-avoid-js-useless-arrow-func-bind-error-in-storybook-template

IanMayo (approved):

PR #602, opened 2021-02-11
Feature/307 formatting for destroyed state

### LINKED ISSUE #307: Adjudication Panel - destroyed state
In the adjudication panel, if the umpire has selected `Destroyed` then there isn't anything to accept/reject.  So, the Accept/Reject buttons should be disabled (or hidden).  Only `Save` will be provided.

Note: in my recent PRs I've made good progress on this, but the logic still isn't quite right. We lose "Save" button if Asset is set to Destroyed. But, if that was done by mistake then it can't be changed - since there isn't a "SAve" button.

Maybe we need to test using the value in the `condition` combo, rather than the `condition` value in the asset.

IanMayo:
Fixes #307

### DIFF
--- a/packages/components/src/local/adjudicate-turn-form/helpers/adjudication-manager.ts
+++ b/packages/components/src/local/adjudicate-turn-form/helpers/adjudication-manager.ts
@@ -266,6 +266,10 @@ class AdjudicationManager {
     }
   }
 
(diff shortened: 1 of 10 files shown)


IanMayo:
Not quite complete, but merging so we have minimum branches open for Lilit's monster "delete unused code" PR.

Issue #604, opened 2021-02-16
RFI Status Board


IanMayo:
![image](https://user-images.githubusercontent.com/1108513/108077500-7bbdaa00-7064-11eb-8eb2-bb615610b47f.png)

## 💬 Brief description of feature

Players of the Umpire force faced a challenge in managing the number of Requests for Information (RFIs) that were sent by the Red/Blue force players.

They introduced a table on a whiteboard to manage/track the RFIs, but would like an electronic version instead.

This task covers the provision of support for managing and tracking RFIs.  It includes the provision of a dedicated UI panel for RFIs, only visible to members of Umpire force, plus the introduction of a role that is responsible for assigning RFIs.

We will implement the role of RFI Manager by adding a new toggle in the Role editor of Game Admin.  I suggest we use the `book-reader` font-awesome icon for this role, and the tooltip would be `Can approve RFI responses`.  Like some of the other role toggles, this will only be available for players of Umpire force.

Note: RFI Manager role added in this PR: https://github.com/serge-web/serge/pull/618

This toggle will be in `role.d.ts` as `isRFIManager`.

## 📖 User Story

As an umpire I wish to manage and track RFIs in order to avoid failing to reply to a player question

## <a name="suggested-steps" />Suggested steps
1. Add flag in admin for Umpire Roles to have RFI Manager flag.  DONE
2. Create player interface.  If force != umpire and only template available is RFI, show RFI submit form
3. Allow form to be sent
4. Create RFI status board.  Special channel that shows RFIs across whole application.  Allows RFI Manager to assign issue to anyone from Umpire force
5. Allow assigned user to edit response from within status board
6. Allow assigned user to submit answer for review, from within status board
7. Allow RFI Manager to approve answer
8. Let response be displayed in Player interface

## ☑️ Acceptance criteria

1. Player able to submit RFI question.   Question displayed in channel for all channel members.  Umpire sees RFI as `Pending`.
2.  A channel titled "RFI Status" will be added.  This has roles assigned to it.  But, it doesn't display the contents of a particular channel.  It displays RFI messages sent in any channel.
3. RFI Status Channel has dedicated UI for tracking/managing RFIs
4. RFI Maintainer role(s) are able to assign RFIs to members of Umpire force.  When RFI is assigned, it gets re-submitted to the database with new status (but with original ID).  Also gets submitted if it moves to another player, or if the answer gets edited.
5. RFI updates a visible to Umpire Force, but not Player force.
[...]

PR #623, opened 2021-02-26
Additional data types to support RFIs

IanMayo:
Introduce new message data elements necessary for RFI Status Board #604


IanMayo:
@zukilover  - I've added you as reviewer in here, to check the new data elements match what you're expecting.

PR #634, opened 2021-03-24
Feature/502 simplified chat

### LINKED ISSUE #502: Simple UI for plain chat message

sanjayradadiya:
## 🧰 Issue

Fixes #502 

## 🔗 Link to preview

https://serge-review-feature-50-qq1xxd.herokuapp.com/storybook/?path=/story/local-molecules-chatmessage--demonstration

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Nice start @sanjayradadiya 

The StoryBook controls weren't showing.  I think that's because we're part way through the transition to StoryBook 6.

I've updated the `stories` to follow the new standard. I've also made some `props` compulsory, in order to:
- make the Storybook boolean controls work
- reflect _real life_, where we will either always know the value, or where it's essential that the value is provided.

packages/components/src/local/molecules/chat-message/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+/* Import Stylesheet */
+import styles from './styles.module.scss'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faUserSecret } from '@fortawesome/free-solid-svg-icons'
+import Box from '@material-ui/core/Box'
+import moment from 'moment'
+import Badge from '../../atoms/badge'
+import Paragraph from '../../atoms/paragraph'
+
+const useStyles = makeStyles({
```
IanMayo: This is personal preference @sanjayradadiya - but could you put the styles in `styles.module.scss` rather than using the `makeStyle` call please?   We're not making heavy use of material, so I'm trying to reduce its usage.

sanjayradadiya: okay

sanjayradadiya: but how can i make color dynamic if i am adding into styles.module.scss ?

IanMayo: This is how we do it for non-chat messages:
image

sanjayradadiya: Yes, Got it

IanMayo:
Nice work. I guess the next steps are to create the message editor component. This sits at the bottom of the column, and is used for entering new messages.

The next phase after that will probably be integrating the new component into the game UI.  I _think_ this will be via the `factory.tsx` in `ChannelTabsContainer`.  In there we currently choose whether "this channel" is a map or a normal message channel. We'll have to introduce logic to determine if it's actually a chat channel (i.e. it's not called `mapping` and the message templates for the channel are either empty or just equal to `chat`).    We'll inspect the `ChannelData` items to determine that.

Also note - the ChatMessagesList won't receive the data in the same form as your story.  It will be more like the content in `game-messages.mock.ts`.

But, a lot of that is for the future :-)

IanMayo:
A user has just been reviewing the StoryBook, and made a request. Could we introduce a key combination to send? 
* If the cursor is in the text-box, or the private text-box then <ctrl-enter> would send the message.
* Include a description of the shortcut in the tooltip for "Send".

packages/client/src/Components/ChatChannel.tsx
```diff
@@ -0,0 +1,150 @@
...
+    }
+  })
+  const icons = state.channels[channelId].forceIcons
+  const colors = state.channels[channelId].forceColors
+
+  console.log('state.channels', state.channels);
+
+  return (
+    <div className={channelTabClass} data-channel-id={channelId}>
+      <ChatMessagesList
```
IanMayo, line 63: We shouldn't be hard-coding the messages in here, we should be passing the `messages` const array.

sanjayradadiya: Your right, just adding a hardcoded message here for just checking UI either it is working as expected or need some styling like this.

IanMayo:
Ok @sanjayradadiya - I have this working. Changes pushed.
image

I have a couple of layout issues that I'd like you to fix, please:
* [x] the Chat Channel by default is taller than the available space -  so it introduces a scroll bar.  There should not be a scroll bar
* [x] when playing as White/Umpire the private field is hidden off the bottom of the page:
image

IanMayo:
@sanjayradadiya - there's a conflict in `factory.tsx` - I'm fixing it.

PR #640, opened 2021-03-29
Feature/604 rfi channel

lilitkarapetyan:
## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Hello @lilitkarapetyan - I've made good progress in here, now I need to pass it back to you.

I've made these changes:
* [x] Updated RFI message template to include `Reference` and `Title` fields
* [x] Update RFI Channel rendering to display the `Reference` and `Title`
* [x] Update the Mock RFI data to match above structure 

I need these changes from you, please
* [x] I don't think we need an RFI channel.  We just have normal channels, and when we're rendering the messages we use the RFI Message component instead of the normal message component if `message.details.messageType === 'RFI'`.
* [x] Players from the umpire force will see the extra fields in the RFI Message component.  When they change the state of the message (by clicking one of the buttons at the bottom) - we need to push the message back to PouchDb with the same `_id` but with a new `_rev`.
* [x] When we load messages from the database, we need to filter them to just have the most recent version. This should use the `mostRecentOnly()` helper function

I am less clear on this last step.  The RFI messages should all have an id-value (`Red-1`, `Red-2`, etc.).  Even though there may be more than one RFI channel for a force, there is just one counter for that force.  When the player submits a new message, we should check if it has a field called `Reference` that has a value of `` `` ``.

This reference be the force name, plus a counter for that force.  I am not sure how we maintain this counter.   I have these thoughts:
- when we open, we load all messages. Loop through them to find any with a `Reference` field.  Keep track of the highest value, and store it in `PlayerUI` state.
- when we send a new message, and it has an empty `Reference` field, load all game messages & loop through as above
- introduce a [local document](https://pouchdb.com/guides/local-documents.html) in the database, titled `ref-counter-blue`.  When sending a message with an empty reference field, check if that document exists.  If it doesn't create it, and put in `1`.  If it does exist, take the value, increment it to use as reference, push the new document back.

Can you think of another way of handling this reference number?

IanMayo:
Could you please help me out with:
* [ ] we remember which messages are open using an integer index. But, when change a message the new version is inserted at the top of the listing. The integer index means a different message is now open. I _think_ we need to remember open/close using `message.Reference` (if present) else `_id`).  Hmm, or maybe just collapse the current message when a new one is sent.
* [x] Could you make the RFI message detail boxes the smallest necessary to show their contents please?
* [ ] Could you give the code a general review?  The `mostRecentOnly` stuff was a bit of a challenge.  I also refactored the code to remove the use of MessageTypes as an "in-between" storage container - now we just pass in the MessageCustom
* [x] We have a problem where we show a `Turn Marker` for every wargame document update.  We should reduce them to just show one for each turn.
* [x] Some general tidying on the RFI Detail & RFI Header, particularly the UI & _gloss_/

I still need to:
* [x] Produce the code to generate a new reference number for that force. I'll do it by going through all the messages and finding the highest reference number for this force
* [x] RFI status board is currently blank - I must have broken it

PR #671, opened 2021-04-14
Added condition selection in visibility component

### LINKED ISSUE #648: Umpire should be able to edit `Condition` in Planning phase
Each type of platform in a wargame has a range of conditions. These detail how capable the unit it is.  They can range from "fully working" through "partially disabled" to "destroyed".  By convention the first and last condition represent the platform being able to do anything it wants to, and the platform being unable to do anything.  For the in-between conditions the players & umpire will follow the rule-set for that game.

During the adjudication phase the Umpire will decide on the condition of an asset.  They may decide that there has been some machinery failure that restricts speed, or that the platform suffered damage in conflict.  At the end of the adjudication phase the new state is sent back to the players.

On occasion the umpire(s) could realise too late that they have failed to correctly update the condition of a platform, so we need them to be able to change platform condition during the playing phase. We already support this for visibility.  During the planning phase, if the umpire clicks on a red/blue platform the visibility dialog opens.  Or if they select a Green platform from the `visibility` tab they can change its visibility.  We need to add `Condition` to this dialog.

In case an error is made in the Adjudication phase, the Umpire should be able to change asset condition in the planning phase.

An example of allowing the Game Control to set `Condition` can be seen in the Adjudication Form:
image

It probably needs to get added to the `Visibility` panel that game control sees in the Planning Phase:
image

Oh, and change the `Visibility` panel title and tab title to `visibility and condition`
image

This issue will need these steps:
* [x] Rename `VisibilityForm` to `VisibilityAndConditionForm`
* [x] Modify `collateVisibilityFormData` method to return visibility and condition values. The code to get conditions for the selected platform can be found in `collateAdjudicationFormData`.
* [x] Provide list of conditions in vis and condition form
* [x] Change `MessageVisibilityChanges` to include optional field for new condition
* [x] On visibility-form.onSubmit, include new condition if it's different to the original one
* [x] Rename `handleVisibilityChanges` to `handleVisibilityAndConditionChanges`.
* [x] Modify `handleVisibilityAndConditionChanges` to update the platform condition, if a change is present.
* [x] Rename `handle-visibility-changes.spec.ts`, and update test to verify correct handling of condition

Note: this issue will introduce you to the non-messaging side of the app.  Here is a file describing the forces in a wargame:
https://github.com/serge-web/serge/blob/develop/packages/mocks/forces.mock.ts

It's a list of forces (white, blue, red), and for each one there is an optional list of assets (ships, helicopters, etc).   For each one there is a platformType.   There is also a condition - that's what is being edited in this issue.  

This other document (`platform-type`) provides the details for each platform. It includes a list of conditions that apply to that platform:
https://github.com/serge-web/serge/blob/develop/packages/mocks/platform-types.mock.ts

When it comes to testing the vis changes, this flow has proved useful:
* Open unit test (_spec.ts_) file, such as `hand-visibility-changes.spec.ts`
* Open command line, navigate into `components\helpers`
* Type `yarn test:unit visibility --watch`. In this case`visibility` wildcard matches against the test your working on.
* Each time you save, the test will re-run.

sanjayradadiya:
## 🧰 Issue

Fixes #648 

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

image

## Confirmations


IanMayo:
All looking good so far @sanjayradadiya  - well, one minor change. I'll test it within a running version of the game tomorrow.

packages/helpers/src/tests/handle-visibility-condition-changes.spec.ts
```diff
@@ -6,6 +6,7 @@ import findAsset from '../find-asset'
 
 const payload: MessageVisibilityChanges = {
   messageType: VISIBILITY_CHANGES,
+  assetId: 'C06',
   visibility: [
     {
       assetId: 'C06',
```
IanMayo: you should remove the above assetId - since the `visibility` interface has changed

IanMayo:
there is a deep scan error reported in hex grid, but I'll fix that.

packages/helpers/src/tests/handle-visibility-condition-changes.spec.ts
sanjayradadiya: Not getting you, as you have asked me to add assetId there.

IanMayo: The new line 9 is valid. We don't need the duplicate `assetId` in the new line 12.

IanMayo (approved):

PR #672, opened 2021-04-15
Bugfix/578 snapshot generation

### LINKED ISSUE #578: Fix snapshots
Whenever I do a `git push`, the git-hook does a full build, which updates lots of `.snap` snapshots.

These then show as new changes to be committed.

But, the only items that change are id values and date-time parameters.

[Here](https://jestjs.io/docs/snapshot-testing#property-matchers) is an article that explains how to ignore differences in id values or dates.

[Here](https://jestjs.io/docs/snapshot-testing#interactive-snapshot-mode) is an article that introduces _interactive updates_ where the developer can decide whether to update tests (if they're still valid) or to fail tests (if it's a sign of broken code).

Our use of snap-testing seems broken.

I would expect it to work like this:
- in testing, compare element to stored snapshot.  If they don't match - then either fail the test or update the snapshot (depending upon if it's an indication of something broken).

For the test comparisons to work, I think the comparison has to avoid comparing ids or date/times.  Or we have to change the snapshots so that they don't include sensitive dates or times.

But, I think we have this:
- don't compare snaps during testing
- in build, rebuild all snapshots

> Why do we have snapshots for every component?

I _think_ this is the intended behaviour:

1. During initial development of a component, the developer creates a `spec` file for it. This creates an instance of the component, plus a snapshot, and compares the instance to the snapshot.  The component, the `spec` and the snapshot get submitted as part of the PR
2. During subsequent development, builds should create a new instance of the component and compare that to the snapshot, and fail if there is a difference. A developer can then inspect the difference(s) and decide if the changes he is making have caused a _knock-on_ problem with that component, or if some other regression error has arisen.  If there is no real problem the developer would update the snapshot (somehow), but if there is a problem it should get addressed as part of that PR.

sandro768:
## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/components/__mocks__/@fortawesome/react-fontawesome.tsx
```diff
@@ -0,0 +1,30 @@
+/**
+ * Mocks @fortawesome/react-fontawesome.js to return a simple, empty <svg> element
```
IanMayo, line 2: hey - well spotted!

IanMayo (changes requested):
Thanks, nice work.

Note: when I run jest coverage I get snapshot errors like this:
image

I think it's clear what's happening there.  The `dateTime` value is 15:08...Z.  The `Z` means it's UTC.  But, moment is switching it to your timezone, which is `19:08`.

I think that for our tests, we need to tell moment to always use UTC - then whereever the developer is, the times will be formatted the same.

It looks like we can tell moment what timezone to use:
https://momentjs.com/timezone/docs/#/using-timezones/default-timezone/

So, we _probably_ just need to set that default in each `.spec` that needs it.

Hmm, how can you find where it's needed.  You could change your timezone and re-run your tests, then ones with time would fail.

Or, I can update the snapshots so they're all UTC - then when you run your tests they will fail until the default timezones are set.

IanMayo:
Hello @sandro768  - do you think you'll have chance to make the above timezone-related fix?

sandro768:
> Hello @sandro768 - do you think you'll have chance to make the above timezone-related fix?

I think the one way of doing is to mock the data the same way it's done for font-awesome icons.

IanMayo:
You don't think the fix to set the local timezone to UTC in all instances won't fix it?

sandro768:
> You don't think the fix to set the local timezone to UTC in all instances won't fix it?

Do we want to fix the local timezone in the components or tests only?

IanMayo:
It would only have to happen in the tests.  If we did it in the deployed app then daylight-savings-time won't be respected.

sandro768:
@IanMayo It's ready for testing

IanMayo (approved):

Issue #646, opened 2021-04-01
Need to flexibly load JSON data-file


IanMayo:
For the storybook titled `Atlantic Mapping` and the wargame titled `Task Group Initialised`, we load a very large datafile of hexagon cells (`atlantic-cells.ts`).  This gets produced by scientists, using Python.  They need to be able to provide an update to the file.

The file we're using is a TypeScript object, which gets transpiled and built into the app.

But, the new versions will be delivered as a single file of json code.  They won't get built into the app, but we can get them to put the file in a particular folder location - replacing the previous copy.

To do that, we probably need to load it in from the `client\public` folder.  Since we're loading it at runtime, we probably need to load it as JSON, rather than TS.

 So, to load `json` at run-time it may be an async call.

Ideally, we will find a way to load the same data-file from the `Storybook` for `Mapping`.

IanMayo:
Fixed in #679

PR #679, opened 2021-04-19
Flexible atlantic json file

sanjayradadiya:
## 🧰 Issue

issue #646 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Nice react coding, @sanjayradadiya .

IanMayo:
~~We should remove the `atlanticCells.ts` file, please. We'll still have a safe copy in the repo history.~~

No - let's keep it, it may be useful for testing.

packages/server/bin.js
```diff
@@ -20,7 +20,8 @@ server(
       'https://serge-dev.herokuapp.com',
       'http://localhost:3000',
       'http://localhost:8080',
-      'http://localhost:8000'
+      'http://localhost:8000',
+      'http://localhost:6611'
```
IanMayo, line 24: Can you explain why we've registered this port?

Oh, and do you know why we have `main.js` _and_ `bin.js` ?

I know you didn't create them - I'm just wondering if you understand why.

sanjayradadiya: This port is register to get out from CORS issues while request from other apps like storybook and web app.

Sorry, no idea about the difference between that file but I just see we had added this URL in two files so I added it for reference.

It would be good if you give more explanations here if I had added wrongly 😂

IanMayo:
I've tried this in storybook, but it fails to get the tiles:
image

I wonder if you need to include some mock in the story - aaah, like in the article I posted to the issue yesterday.

sanjayradadiya:
> I've tried this in storybook, but it fails to get the tiles:
> image
> 
> I wonder if you need to include some mock in the story - aaah, like in the article I posted to the issue yesterday.

Maybe your server is not started or just check the network request

you can check running example here
https://serge-review-646-flexib-7ps3t4.herokuapp.com/storybook/?path=/story/local-atlanticmapping--natural-earth

image

packages/server/server.js
```diff
@@ -94,6 +94,10 @@ const runServer = (
     res.status(200).send({ ip: req.ip })
   })
 
+  app.get('/atlantic-cells.json', (req, res) => {
```
IanMayo: This requires us to use this particular file-name.  They may need to have a number of different files of data.

I appreciate we need to define some kind of route.  How about if the request begins with `/cells/` - and we provide the contents of the file at `/cells/xxxxx`?

IanMayo:
> Maybe your server is not started or just check the network request

Could you add the instructions you used to start the server in the PR description please?

Hmm, I think we need a solution that doesn't require another server running.

packages/server/server.js
sanjayradadiya: would be a great, good idea 💡

sanjayradadiya:
> > Maybe your server is not started or just check the network request
> 
> Could you add the instructions you used to start the server in the PR description please?
> 
> Hmm, I think we need a solution that doesn't require another server running.

there could be an issue only in local env but on prod or test env server is always up

packages/server/bin.js
IanMayo: Aah, it's the port StoryBook is registered to use:
image

Could you also do some digging (file-searching) to see where the two files are used, please?

We can then put explanatory comments in the files themselves.

sanjayradadiya: ok sure

IanMayo:
Aah, I guess we could _mock_ the `fetch` call, and just return the file with the supplied filename.

sanjayradadiya:
> Aah, I guess we could _mock_ the `fetch` call, and just return the file with the supplied filename.

yeah, right

IanMayo:
@sanjayradadiya  - I've pushed some changes to this branch.  Sorry, I intended to create a branch off it, to demonstrate something to some users.

My change involves switching to a much larger hex cells file.  I don't think it has any other affect on your work/branch.

I still welcome you removing the necessity for the server process to be running, so that I can run local Storybook tests.

IanMayo:
@sanjayradadiya - you've introduce a new `component` for `data`.  It doesn't contain a `package.json`.  It would _feel_ more tidy to include one, even if it's just for the `description` of what it's doing. Could you investigate if there are best-practices about adding new components.  Maybe this is a mono-repo issue, maybe not.

sanjayradadiya:
> @sanjayradadiya - you've introduce a new `component` for `data`. It doesn't contain a `package.json`. It would _feel_ more tidy to include one, even if it's just for the `description` of what it's doing. Could you investigate if there are best-practices about adding new components. Maybe this is a mono-repo issue, maybe not.

I am not sure about mono-repo so can you give me some idea what is expected behaviour ?

IanMayo:
Sure. I'm not expecting a particular behaviour, I'm just thinking about good engineering.

Here is the `package.json` for `packages\components`:
image

I _guess_ the first line allows us to refer to it from other packages.  Including the others seems _tidy_.  You're welcome to include your email as author if you want a little fame, else use mine `ian@planetmayo.com`.

IanMayo:
@lilitkarapetyan - we're stuck in this branch.

For one of our StoryBook tests (`index3.stories.tsx`), we need to mock a call to `whatwg-fetch`.  The call actually occurs in `mapping\index.tsx`. 

You told jest to mock a function in `dbWargames.test.js`, like this:
image

Neither @sanjayradadiya  nor I can mock this function . We initially tried defining it in `__mocks__`, but the inline mocking like you did above would be acceptable.

Are you able to help us please?

**Note:** aah, hold on Lilit.  Since this is in `StoryBook` rather than a `jest` unit test,  the above probably isn't the right strategy.  It now looks like we need to use `__mocks__` like described here: 
https://storybook.js.org/docs/react/workflows/build-pages-with-storybook#mocking-imports

It looks seriously complicated though.

packages/components/src/local/mapping/index3.stories.tsx
```diff
@@ -16,7 +16,7 @@ import { HexGrid } from '../hex-grid'
 // import data types
 import { Phase, serverPath } from '@serge/config'
 
-import data from './data/atlantic-cells';
+import data from './data/atlantic-cells'
```
IanMayo, line 18: @sanjayradadiya - if you put a package.json in `data`, you can call it `@serge/data`, then this import becomes `@serge/atlantic-cells` (I think...).  That's the way we do it in the rest of the app.

sanjayradadiya: ohhh, yes i forget that

sanjayradadiya: This are different one it is the sample data inside `mapping` component `atlantic-cells.ts` file and file inside data folder is access via fetch call so it can't be import anywhere.

IanMayo (approved):

PR #681, opened 2021-04-20
resolved version error on lerna for pkg-build

### LINKED ISSUE #676: Deployment packager not working
In the root we have an npm command to package up a deployment: `pkg-build`.

The command _should_ create a deployable instance of Serge that can be deployed to a PC/network without an Internet connection.

It's failing:
```
yarn run v1.22.10
$ lerna run start --scope @serge/executable --stream
lerna notice cli v3.22.1
lerna info versioning independent
lerna notice filter including "@serge/executable"
lerna info filter [ '@serge/executable' ]
lerna ERR! EFILTER No packages remain after filtering [ '@serge/executable' ]
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### LINKED ISSUE #999: Need to be able to edit existing copy of document in Collab Edit channel
If a collab-working channel is setup for Collaborative Editing, we need to view (and edit) the document using the relevant JSON-Editor template.

So, if the document is in `Edit Document` state, the JSON editor is enabled, else it is disabled, in read-only mode.

Dhamu143:
fixes #676

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "storybook": "lerna run storybook --scope @serge/components --stream",
     "build-storybook": "lerna run build-storybook --scope @serge/components --stream",
     "build": "lerna run build --scope @serge/components --scope @serge/config --scope @serge/helpers --scope @serge/store --scope @serge/client --stream",
-    "pkg-build": "lerna run start --scope @serge/executable --stream",
+    "pkg-build": "lerna run start --stream",
     "component-build": "lerna run build --scope @serge/components --stream",
     "component-dev": "yarn component-build & lerna run watch --parallel --scope @serge/components",
     "diff": "lerna diff",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "storybook": "lerna run storybook --scope @serge/components --stream",
     "build-storybook": "lerna run build-storybook --scope @serge/components --stream",
     "build": "lerna run build --scope @serge/components --scope @serge/config --scope @serge/helpers --scope @serge/store --scope @serge/client --stream",
-    "pkg-build": "lerna run pkg-build --scope @serge/executable --stream",
+    "pkg-build": "lerna run start --scope @serge/executable --stream",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "storybook": "lerna run storybook --scope @serge/components --stream",
     "build-storybook": "lerna run build-storybook --scope @serge/components --stream",
     "build": "lerna run build --scope @serge/components --scope @serge/config --scope @serge/helpers --scope @serge/store --scope @serge/client --stream",
-    "pkg-build": "lerna run start --scope @serge/executable --stream",
```
IanMayo: Hmm, it did look like we were previously telling lerna to just build the executable.  Wouldn't that give better performance?

Dhamu143: not sure but issue was causing due to --scope flag.

IanMayo:
Thanks for getting it to build.  But, it doesn't look like a Lerna version issue. It looks like solution just relaxes the target to cover the whole repo.

IanMayo:
@Dhamu143 - I don't know that much about Lerna. I gave the package.json a name, then the `scope` call worked.

But, the build step still doesn't work.  I think the build step was from when this was still a plain single repo.  I _think_ the `pkg` instructions may need to be modified to collate the components from the monorepo.

IanMayo:
At @Dhamu143 recommendation I ran original command  on ms-windows.   Well, I named the `executable` package in its `package.json` and the original build command ran.  It ran with these warnings/errors.
[build.txt](https://github.com/serge-web/serge/files/6350643/build.txt)


It did produce an executable though.   When I ran the executable I got these errors/warnings:
[res.txt](https://github.com/serge-web/serge/files/6350646/res.txt)

Note: if I run `pkg-build` without a scope value, it runs start in all repos.  This give the attached build errors.
[build3.txt](https://github.com/serge-web/serge/files/6350875/build3.txt)

IanMayo (approved):
Fix package.json

PR #684, opened 2021-04-22
646 fetch as props

IanMayo:
Minor change to allow the `fetch` function used in `<Mapping>` to be over-ridden. We needed to do that so StoryBook can retrieve the hex cell definitions.


sanjayradadiya (approved):
Looks good to me 👍

PR #685, opened 2021-04-22
update pkg-build

### LINKED ISSUE #698: Packaged serge needs to expose `data` folder at runtime
The `data` package contains files that the users may wish to edit/overwrite at runtime.

That's easy enough in a cloned repo where the `packages\data` folder can be easily found. But, it's not possible for an Serge deployment that has been packaged into an executable with `pkg`.

So, when the packaged Serge is run for the first time it should create (and use) the `data` sub-folder.

When run on other occasions, it should not overwrite this folder, since it may contain user data.

lebaphi:
fixes #698


packages/executable/package.json
```diff
@@ -13,15 +13,19 @@
...
+    "targets": [
+      "node14-macos-x64"
+    ],
+    "outputPath": "build"
   },
   "scripts": {
-    "start": "node helpers/pkgResolveErrors.js && pkg . --out-path ./builds && node helpers/pkgCopyNodes.js",
-    "pkg-build": "pkg . --out-path ./builds"
+    "start": "node helpers/pkgResolveErrors.js && rm -rf build && pkg . && node helpers/pkgCopyNodes.js",
+    "pkg-build": "rm -rf build && pkg ."
```
IanMayo: @lebaphi - we don't need the script at line 28, do we?

packages/executable/package.json
```diff
@@ -13,15 +13,19 @@
...
       "../client/build/**/*",
-      "../server/default_img/**/*"
+      "../server/**/*"
     ],
     "scripts": [
       "../server/addons/**/*"
-    ]
+    ],
+    "targets": [
+      "node14-macos-x64"
```
IanMayo: This is fine for testing @lebaphi , but MS-Windows is the real desired target. I _believe_ that `pkg` can build MS-Windows executables under Macos.

packages/executable/package.json
lebaphi: yep, I think we don't need it. I see it existed before so I update it again :D

packages/executable/package.json
lebaphi: yep @IanMayo , it should be "node14", I will commit the change now.
image

IanMayo:
Thanks for this Phi,
could you investigate the cause of these 2 errors please?
image

(They may not affect `pkg-build`, but I'd like us to get down to zero errors/warnings.  This will make it easier to spot any new errors in the future).

IanMayo:
Hello @philb89  - the current script only builds for macos.   Could you modify it so that it builds for macos, linux and Win64 please?

IanMayo:
@lilitkarapetyan is the original author of this code, I've invited her to review it.

@lebaphi - there are quite a few changes to `server.js`.  Could you explain what the changes are for please?

lebaphi:
> @lilitkarapetyan is the original author of this code, I've invited her to review it.
> 
> @lebaphi - there are quite a few changes to `server.js`. Could you explain what the changes are for please?

Hi @IanMayo , actually I don't change the server, just do some refactor code of moving the following:
```
 const onAppInitListeningAddons = []
  const onAppStartListeningAddons = []

  addons.forEach(addonId => {
    const addon = require(`./addons/${addonId}`)
    switch (addon.info.on) {
      case 'app-start-listening':
        onAppStartListeningAddons.push(addon)
        break
      default:
        onAppInitListeningAddons.push(addon)
    }
  })
```
to `main.js`. Place this code here will show a warning error `Can not resolve the dynamic path of "./addons/${addonId}"` when the pkg compiles the script. The flow start server is still the same but we can avoid this error.

Another one I remove the following:

```
const opn = require('opn')
```

The `opn` package also makes the pkg compile show a warning/error of can not bundle the library into the executable file. I replace this by the following:

```
const start = (process.platform === 'darwin' ? 'open' : process.platform === 'win32' ? 'start' : 'xdg-open')
require('child_process').exec(start + ' ' + `http://localhost:${port}`)
```

It uses the `child_process` of the node to open the URL then the compile error has gone.

IanMayo:
Thanks for this Phi,
can you explain to me how the `_sqlite.node` files get integrated into the build?  Or, does each executable go into the relevant folder for that operating system?

lebaphi:
> Thanks for this Phi,
> could you investigate the cause of these 2 errors please?
> image
> 
> (They may not affect `pkg-build`, but I'd like us to get down to zero errors/warnings. This will make it easier to spot any new errors in the future).

Hi @IanMayo, it seems likes an issue of yarn and you can ignore it, more details at https://github.com/yarnpkg/yarn/issues/7803. You can downgrade yarn to version `1.19.2` to avoid this warning.

lebaphi:
> Thanks for this Phi,
> can you explain to me how the `_sqlite.node` files get integrated into the build? Or, does each executable go into the relevant folder for that operating system?

Hi Ian, you should distribute the `sqlite3.node` along with the final executable because this issue https://github.com/vercel/pkg/issues/141. it is required by the adapter `websql` in `PouchDb Options` and make sure that you build your project with pkg with the same node version that you built sqlite3 with.
To make the executable go into the relevant folder, I will write a custom script to handle this.

IanMayo:
> make sure that you build your project with pkg with the same node version that you built sqlite3 with.
This did cause a real issue a couple of years ago - memory leaks that eventually crashed the process.

Are we able to either:
* force a sqlite rebuild - to ensure the same version of node. I _guess_ this could be done my yarn uninstall sqlite, then yarn install, then pkg-build
* Introduce some test to verify they are the same version. I _think_ we can do that from the folder name in `node_modules`

IanMayo:
Hello @lebaphi - I've got `yarn install` working in a VM, and have run `pkg-build` - hey, it worked!

One issue. When the app starts, it's opening the Admin login page (below), rather than the normal root folder.
image

I _don't think_ the issue is in `executable`, probably in `server`.  But, I'd appreciate you investigating how to change the default URL.

IanMayo:
One other issue. Could we put the contents of `server\db` in the `db` folder that is created the first time the executable is run, please?

lebaphi:
> Hello @lebaphi - I've got `yarn install` working in a VM, and have run `pkg-build` - hey, it worked!
> 
> One issue. When the app starts, it's opening the Admin login page (below), rather than the normal root folder.
> image
> 
> I _don't think_ the issue is in `executable`, probably in `server`. But, I'd appreciate you investigating how to change the default URL.

Hi Ian, it seems you have not rebuild the app yet, please run `yarn build` then `yarn pkg-build` to make sure the pkg packaged the latest code.

IanMayo:
This was on a vanilla VM, @lebaphi .  I ran:
* yarn install (and saw sqlite got built)
* yarn build
* yarn pkg-build

Then I navigated to `executable\build` and ran the windows exe.

I'm pretty sure I wouldn't have had a working interface if it wasn't built ;-)

lebaphi:
> This was on a vanilla VM, @lebaphi . I ran:
> 
> * yarn install (and saw sqlite got built)
> * yarn build
> * yarn pkg-build
> 
> Then I navigated to `executable\build` and ran the windows exe.
> 
> I'm pretty sure I wouldn't have had a working interface if it wasn't built ;-)

which branch did you run? It can not be because I already fixed the URL as the image below:
image

lebaphi:
> One other issue. Could we put the contents of `server\db` in the `db` folder that is created the first time the executable is run, please?

I did, when you run the first time, the executable will create all content of the `server/db` into this location, on the window you will see the `serge/db` folder stay at the same directory of the executable file.
Screen Shot 2021-04-30 at 7 34 34 PM

IanMayo:
Have pulled new branch. Getting this error:
image

lebaphi:
> Have pulled new branch. Getting this error:
> image

this is `rimraf` error I think

IanMayo:
This is mostly sorted now, thanks @lebaphi 

I'd like to add one extra feature please:
- once the executables are built, could we package up a zip file for each operating system?

I guess there are standard naming conventions for these, but I'd like something with the date and operating system, like:
`SERGE_20210505_win64_x64.zip`

With that feature in place, a later modification can upload these with new Pull Requests & Releases - to give a playable demo version.

lebaphi:
> This is mostly sorted now, thanks @lebaphi
> 
> I'd like to add one extra feature please:
> 
> * once the executables are built, could we package up a zip file for each operating system?
> 
> I guess there are standard naming conventions for these, but I'd like something with the date and operating system, like:
> `SERGE_20210505_win64_x64.zip`
> 
> With that feature in place, a later modification can upload these with new Pull Requests & Releases - to give a playable demo version.

okay Ian, will update soon

IanMayo:
Thanks @phileba 

Codacy gave this feedback:
image

lebaphi:
> Thanks @phileba
> 
> Codacy gave this feedback:
> image

Hi @IanMayo , I'm not sure the version of PDM is being used and it seems does not support the new code style of ES6:
```
const promises = nodeFiles.map(async (nodeFile, idx) => {
  const { name, pathExec, pathSqlNode, fileExec, fileSqlNode } = nodeFile
  const prefix = `[${idx + 1}/${nodeFiles.length}] ${name}:`
  await findAndCopy(fileExec, pathExec, prefix, finalDir[idx])
  await findAndCopy(fileSqlNode, pathSqlNode, prefix, finalDir[idx])
})
```
it is a destructuring feature of ES6 (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring), so the codacy should not throw an error Unnecessary block, the unnecessary block is:
```
if (bar) {
    {
        // Bad
    }
}
```

IanMayo:
That's a surprise. We use unboxing in the rest of Serge.

Aah, but this is `executable` package. Maybe we need some package level config to allow it?

The `eslint.rc` is much longer for `components` than for `executable`. 

Are you able to provide enough `eslint.rc` to fix the issue?

lebaphi:
> That's a surprise. We use unboxing in the rest of Serge.
> 
> Aah, but this is `executable` package. Maybe we need some package level config to allow it?
> 
> The `eslint.rc` is much longer for `components` than for `executable`.
> 
> Are you able to provide enough `eslint.rc` to fix the issue?

Sure, I will check it.

packages/executable/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  "extends": [
+    "prettier",
+    "standard" 
+  ],
+  "rules": {
+    "no-lone-blocks": 0
```
IanMayo: isn't that just a workaround @lebaphi ?

If there was a lone block, we would want `eslint` to tell us about it.

Don't we need to tell eslint that we're using modern JS?

IanMayo:
Thanks :-)

I wasn't expecting changes outside the `executable` package.  Afraid I have to finish now for the day. Will be back on it in the morning.

lebaphi:
> Thanks :-)
> 
> I wasn't expecting changes outside the `executable` package. Afraid I have to finish now for the day. Will be back on it in the morning.

the commit hook runs the lint check and I can not push the code if I don't fix lint on some class outside the `executable`. They are all change cause of the lint fix (indent, naming convention, the function returned type....). For the codacy error on the `executable` package, I tried  with the 
```
"parserOptions": {
    "ecmaVersion": 12
  }
```
but it does not work. Need to investigate more :D

IanMayo:
> commit hook

Aah, my commit hook is broken, that's why the errors are there. Thanks for fixing them :-)

Issue #680, opened 2021-04-20
Manage RFIs from RFI Status board


IanMayo:
The umpire force are able to keep an overview of RFIs from the RFI Status Board.

But, it's a read-only view.

We could let the umpire force do all of their RFI management tasks if we switch the RFI Status message format to the one (`ChannelRFIMessageDetail`) we use in the channels.

I _think_ this is just about swapping one component for another.

But, note: when we send an RFI message I'm pretty sure we need to insert the channel name.  If this is being done from the RFI Status Board (special channel) then the channel name to use is the one the message was actually sent to, not the current one.  This will be in the message details block.

Note: replacing the component is the _ideal_ solution since it gives all roles their relevant buttons.  If it's really difficult to swap the components then an acceptable workaround would be to just provide the RFIManager with `Reject` and `Release` buttons.  It's the RFI Manager who _sits_ in the RFI Status Panel - so we deliver the most benefit by letting this user do their job(s) from within the status board.

I've just seen that the `RFIStatusBoard` is in `client`, rather than `components`.  I'd like the first step be to move it over there, and create it as a component in its own folder, with its own `stories` and `spec` tests, please.  Actually - let's do this issue as two PRs.  First to refactor to components, second to change detail form.

sanjayradadiya:
@IanMayo I understand the first concerned that we need to decompose into component but i didn’t get much idea about existing logic and purpose for changes.

IanMayo:
> purpose for changes

The Blue/Red force players submit their questions in the relevant channel, and the experts typically reply in that channel.  But, the RFI Manager has to review all responses before they are sent (to check the response isn't clashing with advice given elsewhere, or breaks the game-play).  The RFI manager can see from the new RFI Status Board that a response is ready for review. But, currently he has to go to that channel & find the message in order to Reject/Release it.   It would be easier if he could do this from the RFI Status Board.

We could meet this requirement by adding Reject/Release buttons to the RFI Detail in the RFI Status Board.  But, since we support all of the buttons/roles in the component that's used to display the RFI in the original channel, then switching the status board to this component would give us the buttons for all of the roles/actions.

IanMayo:
Fixed in #689

PR #687, opened 2021-04-23
RFI status board component

sanjayradadiya:
## 🧰 Issue

issue #680

## 🚀 Overview: 

PR contains only creation of RFI status board component and integration of that component into client side app.
we just added component and removed existing `RfiStatusboard` from client and added newly create component there.

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/components/src/local/rfi-status-board/index.tsx
```diff
@@ -1,22 +1,27 @@
...
-import { CollaborativeMessageStates} from '@serge/config'
+import { CollaborativeMessageStates } from '@serge/config'
 import { ChannelData } from '@serge/custom-types'
 
-const RfiStatusBoard = ({ rfiData}: { rfiData: {rfiMessages:MessageCustom[], roles: string[], channels: Array<ChannelData>} }) => {
+/* Import Types */
+import Props from './types/props'
 
+/* Import Stylesheet */
+// import styles from './styles.module.scss'
```
IanMayo, line 11: aren't we importing the styles?

sanjayradadiya: style will be imported in the same component but in next PR because we did more refactor on that PR #689

packages/client/src/Components/RfiStatusBoardChannel.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
```
IanMayo, line 1: @sanjayradadiya  - this component doesn't add any functionality to RFIStatusBoard. 

It just passes the data through.

We don't need it, do we?

sanjayradadiya: we are passing all states related data through this component so we make it separate from other channel components.

Instead of maintaining all logic inside factory helper we can delegate some part here.

IanMayo: Do we need this component though?  Can't we just use RfiStatusBoard in factory.tsx?

PR #712, opened 2021-05-12
Fix win executable which built on mac does not work on windows

lebaphi:


IanMayo:
This doesn't fix the `pouch_all_dbs` issue.

In the MS-Win `db` folder the size is `68Kb`.

In the macos `db` folder the size is '90Kb'

image

If I copy the `macos` version to `win` then the wargames are all shown.

The other files in the `db` folder are of the correct size.

IanMayo (approved):
Fixed. Thanks

PR #720, opened 2021-05-13
Investigate levels of detail

### LINKED ISSUE #719: Adjust zoom detail levels
As the player zooms in/out, the Serge map adjusts the visibility of these items:
* shaded polygons
* hex outlines
* hex labels

The demo attendees found some of the transitions "jarring". We need to adjust the zoom levels when we decide what to render.

We could interpret the issue as 'detail levels': low, medium and high.  Then, decide the cut-off point at which we transition through them. The cut-off point could be in terms of a LeafletJS zoom level.  Alternatively it could relate to the number of hex cells currently visible. The latter would allow adjustment to different cell sizes or map extents.   i.e.  We could show all detail for all cells if the map had large cells in a relatively small region.

For level of detail vs # cells, this could be a starting point:
* high detail < 500 cells
* medium detail < 2000 cells
* low detail > 2000 cells

Will produce instance of Serge that is `instrumented` to show # of visible cells.

IanMayo:
Fixes #719 
TRIAL of reducing level of detail according to number of visible cells.

This seems more versatile than doing it according to zoom level.


IanMayo:
We have a compiler error in this branch.

See if integrating the `husky` hooks would cause error to be spotted before we push.

Issue #709, opened 2021-05-10
Delete role from Force Admin


IanMayo:
## Issue

I'm defining the roles in a Force.

I've created one too many.  I wish to delete it, but I can't.

## UI 

The UI precedent for this would be a `trash` icon next to the role title that appears on hover.

## Steps to reproduce
* Open Serge
* Go into `Admin` (user icon near top-right of welcome page)
* Press `Enter` to skip entering password
* Select a Game
* Go to `Forces` page
* Select a `Force`
* Try to delete a `Role`.

![image](https://user-images.githubusercontent.com/1108513/117682935-7aa3a100-b1ab-11eb-9a65-f329dd40cbbc.png)

IanMayo:
Fixed in #745

PR #731, opened 2021-05-20
Added delete role feature

sanjayradadiya:
## 🧰 Issue

issue #709 

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

image

## Confirmations


IanMayo:
This is a good piece of work @sanjayradadiya . You've understood the app logic and the data structures.

But, this solution is more complex than is necessary.  Sorry I didn't give more clear direction on what was required.

It would be ok to delete the role on the client - the database would get updated on `Save Force` - just like if a role was added, or role permissions were modified.

I _think_ the correct way to handle the new strategy is in a new PR, I'm afraid.  What do you think?

sanjayradadiya:
> This is a good piece of work @sanjayradadiya . You've understood the app logic and the data structures.
> 
> But, this solution is more complex than is necessary. Sorry I didn't give more clear direction on what was required.
> 
> It would be ok to delete the role on the client - the database would get updated on `Save Force` - just like if a role was added, or role permissions were modified.
> 
> I _think_ the correct way to handle the new strategy is in a new PR, I'm afraid. What do you think?

Totally agree, even I was also thinking about it but my mistake I didn't confirm with you.

IanMayo:
Closing this PR, so a simpler (client-only) one can be created.

PR #813, opened 2021-06-02
Feature/two icon drag drop issues 784

### LINKED ISSUE #784: Two icon drag/drop issues
We have two issues related to moving platforms into / onto others.

There are two actions:
* **Hosting** this is like a helicopter being put onto a ship
* **Comprising** this is like two ships becoming a team (a task group)

Both are only allowed when the platforms are in the same cell.  This bit is working well.

I've modified your implementation - so there is a more green shade when the operation is allowable.

A new requirement means we should not let a `Task Group` _host_ another platform.  The task group is just the name of an organisation - so you can't put a helicopter onto it, for example.

In this example, SSN2 cannot be placed onto SSN3.  The drop-target is shown in Red.  But, if I still drop it, it gets added to `CTF-300`.  If something is dropped on a disabled drop-zone, nothing should happen.

screenshot

Second issue.

I drag `SSN1` over `CTF 300` . I should be allowed to drop it onto CTF-300 (comprising), but not to the `Move onto` drop-zone.  The `move onto` drop-zone is shown as red.  But, I can still drop onto it.  When I do, it gets added as `comprising`.  I think that nothing should happen.

screenshot

The above interactions can be produced in this deployment: 
https://serge-demo-task-group-d-ekn98z.herokuapp.com/?wargame=wargame-koiquada&access=pkoiqzx9e

The deployment is from this branch: `demo_task_group_drag`

lilitkarapetyan:
## 🧰 Issue

closes #784

## 🤔 Reason: 

When dragged, the item being dragged was magnetic. Plugin have no option to turn off magnetizing. I added check do do not fire change (onSet) callback when dragged item outside of dropzone

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
@lilitkarapetyan - the fix isn't working for me.  
screencast

1. I can't move any members of the `Task Group` to the root level.
2. I have two SSN instances in the same cell.  I can't combine them into a `Task Group`.

I can attach a copy of the wargame with the ships all in the same cell:
[wargame-koiquada.sqlite.zip](https://github.com/serge-web/serge/files/6585661/wargame-koiquada.sqlite.zip)

IanMayo:
Closing this PR.  It's no longer relevant after our switch to monolith.

PR #820, opened 2021-06-05
Issue 817: Load game url in address bar

### LINKED ISSUE #817: Always put game/role in address bar
For developer convenience, the capability was added to the game to allow the game/role to be embedded in the URL:
image

In this way, URLs could be sent to stakeholders/devs that took them straight into a particular game, as a particular role.  Also, when the page refreshed as part of the compilation process, the page would stay logged in as that role.

Clicking on the `Role` badge (`CO` below) re-opened the page, looking at the current game/role:
image

It would actually be useful if this was always the case.  So, at the login, clicking `Enter` actually redirects the page to the current URL with wargame/accessCode appended to it.

image

lebaphi:
Fixed #817


IanMayo (approved):
Great, thanks.

PR #821, opened 2021-06-08
System heartbeat checker

### LINKED ISSUE #798: Heartbeat to check system still up.
From the Game Control PC, the `executable_win.exe` crashed on maybe 5 times.

It wasn't obvious that this had happened.

I think the client should be able to check a `heartbeat` on the server every few (5?) seconds.  This very low bandwidth call would just ping the server and get an "ok" answer back.

If it doesn't get an answer, then it will show a `Server down - please check comms` message.

Maybe this heartbeat would be another way of supporting `System down for maintenance`, able to switch back to `Normal processing` once the heartbeat responses return.

The `Server down` could be displayed for Umpire Force players` but `Check connection - please check with admin` if for a player force.

Here is a possible data/UI flow:
Heartbeat _ maintenance mode (1)

sanjayradadiya:
## 🧰 Issue

Fixes #798 

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

image
image

## Confirmations


packages/client/src/ActionsAndReducers/dbWargames/wargames_ActionCreators.js
```diff
@@ -146,6 +151,13 @@ export const populateWargameStore = () => {
   }
 }
 
+export const pingServerWithInterval = () => async (dispatch) => {
+  setInterval(async () => {
+    const serverStatus = await wargamesApi.pingServer()
+    dispatch(saveServerStatus(serverStatus))
+  }, 10000)
```
IanMayo: Could you store the 10000 as a constant in the global consts file please?

sanjayradadiya: sure

packages/components/src/local/atoms/heartbeat-checker/index.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  },
+  argTypes: {
+    enableHeartbeat: {
+      control: {
+        type: 'boolean'
+      }
+    }
+  }
+}
+
```
IanMayo, line 32: @sanjayradadiya - do we have a background server running for StoryBook tests?  So, does the heartbeat work for this?

IMHO, we should mock the backend call, and instead let a toggle control if  (or what) is returned from the `healthcheck` URL.

IanMayo:
Thanks @sanjayradadiya  - I've just had a play.

1. Could the heart icon just "beat" after each server healthcheck?  If currently beats every second, even though the link may actually be down
2. If the healthcheck fails, could you paint the icon grey, instead of red please?
3. I saw the popup message.  It flashed up twice, then disappeared.  It should _probably_ stay visible until the user clears it.

sanjayradadiya:
> Thanks @sanjayradadiya  - I've just had a play.
> 
> 
> 
> 1. Could the heart icon just "beat" after each server healthcheck?  If currently beats every second, even though the link may actually be down
> 
> 2. If the healthcheck fails, could you paint the icon grey, instead of red please?
> 
> 3. I saw the popup message.  It flashed up twice, then disappeared.  It should _probably_ stay visible until the user clears it.

Sure, I will update based on this

packages/components/src/local/atoms/heartbeat-checker/index.stories.tsx
sanjayradadiya: this is plain component it will only animate when flag enabled

IanMayo:
This seems a bit picky @sanjayradadiya  - but....  Could the heart be small, and grow on the "beat"?  It currently shrinks on the beat - which looks wrong.

Could you also change the interval to 5 seconds, please?

sanjayradadiya:
> This seems a bit picky @sanjayradadiya - but.... Could the heart be small, and grow on the "beat"? It currently shrinks on the beat - which looks wrong.
> 
> Could you also change the interval to 5 seconds, please?

which interval you mean ? in keyframe animation or inside component settimeout function ?

IanMayo:
We should call the heartbeat once every 5 seconds, not every 10 seconds.

PR #829, opened 2021-06-10
Fixed position of the context menu

### LINKED ISSUE #826: Wargame context menu not visible with only one wargame
In the Admin page, if only one wargame is present (such as in current `develop` branch), then only the last two items on the context menu are visible.

We _could_ force the list of wargames to take a minimum height, but this may not fix the issue for the lowest issue.

image

This is what the full list should look like:
image

lebaphi:
Fixes #826


IanMayo (approved):
Fix works great, thanks.

Issue #832, opened 2021-06-11
Crash on delete wargame


IanMayo:
If a wargame is deleted from the admin page, the app crashes.

![screencast](https://i.gyazo.com/1293349993a7e0aad5558561521e57e3.gif)

![image](https://user-images.githubusercontent.com/1108513/121690091-4f1f1980-cabd-11eb-8ffc-dc0bcf41c4f9.png)

Note: the wargame does get deleted - since on reload it has gone.

IanMayo:
Fixed in #835

Issue #830, opened 2021-06-10
Newly created wargame gets listed twice


IanMayo:
In the `admin` page we can create a new wargame.

But, when we go back to the admin page, the new wargame is listed twice:

![screencast](https://i.gyazo.com/fcba9089d8623c16bcf562b153115daa.gif)

If we refresh the page, only one game is listed. 

_Maybe_ the new game is added to the list, then Serge adds all existing wargames to that list, which will include the new one.

IanMayo:
21st June: Issue still present.

IanMayo:
Verified as fixed in `develop` branch.

PR #835, opened 2021-06-12
fix clean copy bug

### LINKED ISSUE #827: Client crash on `Create clean copy`
A Game Control is trying to create a duplicate of a wargame from the Admin page.

The `Clean copy` process should copy a wargame, but delete all messages, and just have the wargame description, in a non-initiated state.  Note: Ian doesn't think the flaw is in this logic.

He used `Clean copy` from a wargame's context menu.

A blank screen was shown.

Here is the client-side error log:
image

There is a _chance_ that this issue is related to us starting to append `.sqlite` file suffix to wargames.  When we join a wargame as a player, or when we open a wargame for editing, then the correct filename is found.  But, when we run `Clean Copy` it fails.

Or, maybe... The duplicate wargame is created, but we try to access it before it is physically available.

lilitkarapetyan:
## 🧰 Issue

closes #827,  #830, #832

## 🤔 Reason: 

Fixed Clean Copy bug, but one more bug still there, same crash present for Duplicate

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo (changes requested):
`Clean Copy` no longer crashes client.

But, the new copy can't be opened.

If I create a clean copy, and open the copy, the client crashes.  If I refresh the page, the new copy will open.
screencast

IanMayo:
Hi, there's another issue. If I create a duplicate, I can't save changes to it.

I did `Duplicate`, then went into the new wargame. When I tried to save a change to the wargame title it threw this error:
image

packages/client/src/api/wargames_api/index.ts
```diff
@@ -140,11 +140,11 @@ export const populateWargame = (): Promise<Wargame> => {
   return fetch(serverPath + 'allDbs')
     .then((response: Response): Promise<string[]> => response.json())
     .then((dbs: string[]) => {
-      const wargameNames: string[] = wargameDbStore.map((db) => db.name)
+      const wargameNames: string[] = wargameDbStore.map((db) => db.name)      
       const toCreateDiff: string[] = _.difference(dbs, wargameNames)
       const toCreate: string[] = _.pull(toCreateDiff, MSG_STORE, MSG_TYPE_STORE, SERGE_INFO, '_replicator', '_users')
 
-      toCreate.forEach((name) => {
+      toCreate.forEach((name, i) => {
```
IanMayo: @lilitkarapetyan - the `i` is unused. Should we delete it?

PR #838, opened 2021-06-15
Invalid time issue in wargame timings

### LINKED ISSUE #795: Invalid turn time made planning impossible
At one point no force could plan their turns.

I logged in as the relevant (`Can submit orders`) role for each force, and confirmed it wasn't possible.  It was possible to choose a state, then speed, then click on `Plan Turn`.  But, no planning marker appeared.

On investigation, it was, however, possible to plan the turn for an aircraft.

Aircraft can travel an unlimited distance which lead me to thinking the issue could be related to the turn time allowance.

When I opened the Admin Page, I saw that the `Game Turn Time` just contained spaces - no digits at all.

Shortly before the issue arose, I did change the `planning allowance` from `one hour` to `twenty minutes`.  I'm almost certain this interaction caused the issue, but I don't know how.

In the admin pages we shouldn't allow the overview to be saved if there aren't recognisable (non-zero) times for game turn and planning allowance.  Plus, in the game we should throw an error if we try to plan a turn but there is a zero turn time.

sanjayradadiya:
## 🧰 Issue

Fixes #795 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Thanks. But, I'm still able to clear the time values:
screencast

If the text empty, or the values are all zero, then the previous value should be displayed.

IanMayo:
It's possible to use `backspace` to delete a character:
image

sanjayradadiya:
> It's possible to use `backspace` to delete a character:
> image

Sorry, Not understanding your concern here, what needs to be change ?

IanMayo:
I guess it won't store a valid time if it can't parse the string (because) there is a missing character.

I _think_ it should replace with the last valid value if focus leaves the field when there is a missing character.

IanMayo (approved):
Working great, thanks.

PR #875, opened 2021-06-26
Better rendering of messages

### LINKED ISSUE #865: Better rendering of messages
Besides plain chat, some messages get sent in a structured format.

Here is an example of a structured message:
image

Serge does it's best to render this content in a tidy way, but it's not great:
image

We have these issues:
* [x] doesn't render `true`, it's blank. We should use a tick or cross character if the value isn't a string, but is a boolean
* [x] omits title of composite components. In the data we have `Consumption` and `Location` elements. These aren't rendered. They should be. 
* [x] Object hierarchy is hidden.  For parent objects (arrays or objects) we should show the name larger, more bold, something more distinctive.
* [x] Rendering fails on a boolean value (see `Spare field 2`)
* [ ] label can get separated from the value, sometimes on the next line. We should try to "wrap" the label & value with a line border - and see what that looks like.  The label & value are currently separate spans, we'd have to wrap them in a parent span/div.

Step to reproduce above issue:
* open Serge
* open `tall msg` game
* login as `Game Control`
* expand `Weather Forecast`

The logic for the above is in `channel-message-detail\index.tsx`. There is a StoryBook to demonstrate the actions.

Run `yarn storybook` then open the `Channel Message Detail` story.  The `Weather Forecast` contains the detail for this issue.

lebaphi:
Fixes #865

### DIFF
--- a/packages/components/src/local/atoms/message-label/index.tsx
+++ b/packages/components/src/local/atoms/message-label/index.tsx
@@ -11,7 +11,9 @@ const useMessageLabelStyle = makeStyles((theme: Theme) =>
     }),
     sizeSmall: {
       height: '18px',
-      fontSize: '.75rem'
+      fontSize: '.85rem',
+      fontWeight: 550,
+      marginLeft: '16px'
     },
     labelSmall: {
       paddingLeft: theme.spacing(0.5),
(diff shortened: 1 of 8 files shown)


IanMayo:
Thanks @lebaphi - but that layout may be _too_ verbose.  If the one-line-per-element strategy consumes more than a screen-full of space, then they can't view a whole message at one time.

lebaphi:
> Thanks @lebaphi - but that layout may be _too_ verbose. If the one-line-per-element strategy consumes more than a screen-full of space, then they can't view a whole message at one time.

okay, I just thought to display it as the same as JSON is easier to view. So I will try another go to save the space of the screen 🤔

IanMayo:
Ok @lebaphi - based on the wargame that has run this week, it's important that the formatted messages look more like the version when they were being edited.

So, we're going to close this PR, and instead we'll focus on using a read-only instance of `jsoneditor`

PR #881, opened 2021-06-28
Fix delay issue

### LINKED ISSUE #877: Delay in choosing new message format
Some players can send multiple types of message into a channel (will provide link with demo: https://serge-review.herokuapp.com/?wargame=wargame-kq9ozhe6&access=pkq9oxs4h).

The choices of message type are in a combo-box at the top of the `New Message` dialog.  When switching between message types there can be a delay of up to 4 seconds for the new message template to open.

screencast

I don't recall this happening in past wargames (around a year ago). Here is a recording of a build from December 19:
screencast 2

We should investigate the cause of the delay.

I've just checked the `network` tab in `developer tools`, and the delay isn't because of a server round-trip.

lebaphi:
Fixed #877


IanMayo:
Hello @lebaphi  - could you explain what the issue was please?

IanMayo:
Aah, @lebaphi - I think I understand it.

Rendering Serge was taking several seconds.  We were asking for two re-renders, now we're not.

The fact that Serge rendering is (almost) instanstaneous fixes the issue, but it's right to remove the renders.

Issue #883, opened 2021-06-28
Performance impact with over 150 messages


IanMayo:
The users have a wargame with around 150 messages. Some are up to a page long.

The client can be unresponsive at times.  The clock only moves forward once every 9 seconds.  It can take 5 seconds to switch tabs.

When they type, it can take a few seconds before text starts to appear.

I've mocked up a similar scenario: `CMDWK-HIGH DATA`

Here is the clock. Notice sometimes is jumps two seconds:
![screencast](https://i.gyazo.com/5d9e4bf7304fc31bb8c9fbc22e23c587.gif)

My _gut-feeling_ is that the UI is re-rendering on each clock-tick.

Here is the delay when switching between tabs:
![switch between tabs](https://i.gyazo.com/7f78bcb7f6f7a91ab243f25a82f41f80.gif)

I've just tried typing in a long message, one second at a time. There is a pause in the typing once every 5 seconds.
![delay in processing text chars](https://i.gyazo.com/c87bf52ed934b3ddc5752c9a101b8efb.gif)

Aah, that delay may be related.  When we open the app, there is a delay of 1-2 seconds while it renders the channels on the left-hand side.  If the _heartbeat_ causes it all to re-render, then that would introduce a further delay.  _maybe_ once there are enough messages it can take more than 5 seconds, causing the re-renders to stack up.
![pause on load](https://i.gyazo.com/59eaf91c4574ac19676f866198cba144.gif)

Here is a link to white game control, notice the lag in changing channels: https://serge-review.herokuapp.com/?wargame=wargame-kqgy546g&access=pkoiqu5ev

Here is a link to red-2, which isn't in any high-volume channels: https://serge-review.herokuapp.com/?wargame=wargame-kqgy546g&access=p3d3n

PR #885, opened 2021-06-29
fix ui re-render on heartbeat ping

### LINKED ISSUE #876: "Random" client crash while entering large form
A number of users have to complete a form that has around 6 textarea fields.

At a random point in time (at a moment when they may not actually be typing) the Serge client crashes, leaving them with a plain blue screen.

Looking at the console in developer tools, it's clear the issue is in the `json-editor` component.

```
Uncaught TypeError. "Cannot ready type 'Property' of null" - jsoneditor.js:5361
Uncaught TypeError. "Cannot ready type 'remove_empty_properties_' of null" - jsoneditor.js:3558
```

I've looked at the JS code at that point, and it's when the JS is referring to the schema.   It looks like the schema is null.

The above is a challenging issue.

But, we may be able to address the "Random" aspect.  It may actually be related to the heartbeat.

I think that last week I learned that the whole UI re-renders on the heartbeat.  This is unnecessary.

Could we please isolate the heart-beat so that only the heart icon gets re-rendered?

lilitkarapetyan:
## 🧰 Issue

closes #876
partially: #883

## 🤔 Reason: 

In our case we don't need to save heartbeat ping status in our global store, enough save it in useState inside component, so it will fix our problem.

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo (approved):
Fix working perfectly, thanks.  No delay in typing every 5 seconds

PR #896, opened 2021-07-01
Form editor gets reset on UI refresh

### LINKED ISSUE #892: Form editor gets reset on UI refresh
If a user is in a formatted message editor, the editor gets reset each time a new message is received.

Steps to reproduce:
* enter game `Tall Msg`
* login as game admin
* expand` new message` form
*  select `daily intentions` message type
* write `4` as `turn`
* write `have a party` as overall intentions
* put cursor in Game Admin (on right hand side)
* enter `test message`, press send.
* note that different template is now displayed.  When changing back to `Daily intentions` all the text is gone.

Note: we should only close/recreate the JSONEditor if the user no longer has access to that template type.

This should probably be implemented after #893

sanjayradadiya:
## 🧰 Issue

Fixes #892 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Thanks for this @sanjayradadiya 

Just to help my react education - could you explain how the fix works please?

Either in here as a comment, or in JS-comments, please?

sanjayradadiya:
> Thanks for this @sanjayradadiya 
> 
> 
> 
> Just to help my react education - could you explain how the fix works please?
> 
> 
> 
> Either in here as a comment, or in JS-comments, please?

Sure, mostly i just prevent temple selection from reset while chat input occurred in right tab so i just capture template in prev object so i can ensure that component is render first time and second time we will prevent from value reset to zero position.

IanMayo:
> Sure, mostly i just prevent temple selection from reset while chat input occurred in right tab so i just capture template in prev object so i can ensure that component is render first time and second time we will prevent from value reset to zero position.

I see we never use the value of `prevTemplates` - we we're just using it as a flag to indicate this isn't the first time the page has been rendered for this set of data.  Right?

Aah, and is there a ready why we don't use `usePrevious` from react-hooks?

sanjayradadiya:
> > Sure, mostly i just prevent temple selection from reset while chat input occurred in right tab so i just capture template in prev object so i can ensure that component is render first time and second time we will prevent from value reset to zero position.
> 
> I see we never use the value of `prevTemplates` - we we're just using it as a flag to indicate this isn't the first time the page has been rendered for this set of data. Right?
> 
> Aah, and is there a ready why we don't use `usePrevious` from react-hooks?

we had already used `usePrevious` to check like page is render first time or its getting loaded second time , if we maintain other any kind of flag inside that component then it causing the same issue because its value get also reset while re-rending

IanMayo (approved):

IanMayo (approved):
Thanks @sanjayradadiya  - working perfectly, including:
* send Game-Admin while edit form open
* send in-channel message while edit form open
* remove template from player with edit form open (template successfully removed from drop-down, but message not mangled).

PR #915, opened 2021-07-02
Display formatted messages in read-only editor

### LINKED ISSUE #868: Display formatted messages in read-only editor
Templates are provided for some messages:
image

These structure the information, and lead the player through what to enter.

But, when we display the message for other players, it's just `word-soup`.  The JSON is sorted in alphabetical order, rather than the original order:
image

We may have a better way of presenting the information.  If we presented it in a read-only version of the original editor then it will keep the original information layout.

So, for `ChannelMessageDetail` - we'll look for the message type. Then we'll create an instance of the editor form, insert the message payload, and mark the editor as read-only.

lebaphi:
Fixes #868


IanMayo:
Yes, it works :-)

But, I think in order for it to be acceptable we (you) need to work on improving the format/content.  It takes up so much space the Player(s) can't see the whole message at once.

But, note that the Game Designer may be able to improve the layout in the template layout (through use of table/grid)

image

lebaphi:
> Yes, it works :-)
> 
> But, I think in order for it to be acceptable we (you) need to work on improving the format/content. It takes up so much space the Player(s) can't see the whole message at once.
> 
> But, note that the Game Designer may be able to improve the layout in the template layout (through use of table/grid)
> 
> image

yep, it seems to take too much space. I think we should have another way to display the message the way we want, view the message in vertical doesn't seem to work. Should be displayed in a horizontal line, I will try to put it in a table, take advantage of the empty space on the right to display more data 🤔

IanMayo:
@lebaphi  - it may be more compact if we remove all that whitespace / padding / margins.

lebaphi:
image
remove white space, margin, padding, and add more editor to save the space on the right side

IanMayo:
Thanks @lebaphi - can we do any CSS "tricks" to further reduce the space.  Are we able to convert any `div` to `span`?  Or other steps to reduce the with of those text boxes?

lebaphi:
> Thanks @lebaphi - can we do any CSS "tricks" to further reduce the space. Are we able to convert any `div` to `span`? Or other steps to reduce the with of those text boxes?

I think we can reduce the width of those text boxes or reduce the space, but I'm not sure we can convert any `div` to `span` 🤔

IanMayo:
Aah, I have an issue.  We're not showing all of the content for text fields:
screencast

https://serge-868-4v3qzfxgbl0mmp1j8bk9.herokuapp.com/?wargame=wargame-kqgy546g&access=pkoiqu5ev

All of the submitted text needs to be visible.

IanMayo:
Another thought @lebaphi ,
what if we manipulated the element returned from `json-editor` and replace text boxes with spans?

lebaphi:
> Aah, I have an issue. We're not showing all of the content for text fields:
> screencast
> 
> https://serge-868-4v3qzfxgbl0mmp1j8bk9.herokuapp.com/?wargame=wargame-kqgy546g&access=pkoiqu5ev
> 
> All of the submitted text needs to be visible.

hmmm,  think it should be a text area with CSS: word break 🤔

lebaphi:
image
I converted all input elm to span and set height auto, it looks better

IanMayo:
Thanks - will wait for deployment, and have a look.

IanMayo:
Having a play...

I've gone into `CmdWeek - HighData` as `Red-1/CO`.  In `Red HHQ` I have a `Daily intentions` template.  It takes data in table rows:
image

But, after I send, it's not displayed in a table:
image

lebaphi:
> Having a play...
> 
> I've gone into `CmdWeek - HighData` as `Red-1/CO`. In `Red HHQ` I have a `Daily intentions` template. It takes data in table rows:
> image
> 
> But, after I send, it's not displayed in a table:
> image

which data does not display in the table @IanMayo ?

IanMayo:
`Item 1` in the second image is the first row in the top image, and `Item 2` in the second image is the second row in the top image. 

@lebaphi - you may not have made a mistake, this may just be a part of us learning about `json-editor` in read-only mode.

lebaphi:
> `Item 1` in the second image is the first row in the top image, and `Item 2` in the second image is the second row in the top image.
> 
> @lebaphi - you may not have made a mistake, this may just be a part of us learning about `json-editor` in read-only mode.

ah, it caused by we are displaying the whole message in json editor, its very long in vertical and we do a css trick with column-count:3. We do not display it in the table from the beginning :D

IanMayo:
>  We do not display it in the table from the beginning :D

In the screenshot I gave, it's JSON-Editor that is deciding to use a table, not our code.    Have a look at how `jsoneditor` works with JSON schemas:
https://github.com/json-editor/json-editor

The schema for `daily-intent` specified that `jsoneditor` should display the `array` as a `table`:
https://github.com/serge-web/serge/blob/develop/packages/client/src/Schemas/DailyIntentions.json#L53

IanMayo:
Hello @lebaphi - let's pause this for now.

In the new features we're adding this summer, I'm expecting us to have to do more work with `json-editor`, including the ability to edit existing documents in the editor control.

I'm going to ask @lilitkarapetyan to take on some of that complex work, and include `read-only jsoneditor` in that package of work.  Your work will stay **safe** in here, for reference.

PR #920, opened 2021-07-03
Remove maximize button

### LINKED ISSUE #902: Remove maximise button
Occasionally or accidentally, players clicked on the `maximise` button.

Since our tabs are semi-transparent (so the shiny background texture is shown), players were confused to see tabs on top of other tabs.

screencast

We can overcome this by using `"enableMaximize": false` in our flexlayout config code.

lebaphi:
Fixes #902

### DIFF
--- a/packages/client/src/consts.js
+++ b/packages/client/src/consts.js
@@ -64,10 +64,11 @@ export const SERVER_PING_INTERVAL = 5000
 // were failing CORS test
 export const baseUrl = () => {
   const { hostname, protocol } = window.location
-  const host = (new URL(window.location)).searchParams.get('host')
+  const host = new URL(window.location).searchParams.get('host')
 
   // NOTE: for all non-heroku deployments, we need to append the port number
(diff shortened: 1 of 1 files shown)


IanMayo:
Thanks @lebaphi , but I'm not going to close it right now.

Some players were monitoring around 20 channels - and had a real challenge keeping track of new messages.  In recent days I've realised that "maximise" may just be essential to them:
* arrange channels into 3 zones
* left-hand pane full height, useful for filling in forms
* right hand has upper/lower panes - with long list of channels (tabs) in each one.

The above may allow a user to view all of their tabs, without needing the `Overflow` menu.  But, the two right-hand panes are quite small.  The `maximise` button may be useful in making a message fill the screen.

If the above is valid - then we actually just have to treat this as a user training issue (maybe including it in the walkthrough).

IanMayo:
Closing issue.

We still have this branch/code -  but we'll wait for a stronger reason to remove the maximise button.

PR #922, opened 2021-07-03
Not possible to view Channel details when selecting lower down channels

### LINKED ISSUE #901: Not possible to view Channel details when selecting lower down channels
We need a scrollbar for the list of channels.

If we scroll to the bottom of the list of channels, and select an item, we no longer see the detail, and have to scroll back up.

So, I _guess_ the maximum height of the list of channels needs to the available height, introducing a vertical scroll bar if necessary.

screencast

lebaphi:
Fixes #901


IanMayo (approved):
Graet, thanks.

PR #928, opened 2021-07-07
Feature/hard coded platform icon names 814

### LINKED ISSUE #814: Hard-coded platform icon names
We currently create the platform icon shapes using SVG.

We do this in range of places, but each instances uses one of these two implementations:
* perception form
* asset-icon

These are quite different. The perception form icon sits in the HTML page. The asset icon is rendered/managed by Leaflet Code.

We support platform icons using this trick:
```
$platforms: agi, auxiliary, boghammer, carrier, coastal-radar-site, datum,
  destroyer, fast-attack-craft, fishing-vessel, fixed-wing-aircraft, frigate,
  frigate-ta, helicopter, mcmv, merchant-vessel, mine, missile, mpa, pirate, ssk,
  ssn, task-group, torpedo, unknown, unmanned-airborne-vehicle;

@each $platform in $platforms {
  .platform-type-#{$platform} {
    background-image: url("./counters/#{$platform}.svg");
  }
}
```

But, it is wrong for the list to be hard-coded in our TS code.  The SVG icons are specified in the `platform-types` editor.  We should extract the icon from that data, and specify it in a way that works for both perception-form and asset-icon.

lilitkarapetyan:
## 🧰 Issue

Closes #814

## 🤔 Reason: 

Updated iconDropzone and iconUploader to use api based uploading (if api enabled)
Updated Map and other components withc ones using Icons
Checking, if image loading give  404 error then try to load preserved icon
PlatformTypes objcet instead of PlatformTypes array witch one will allow to easy get platformType object for current asset
Updated Icon Uploader and api to support svg images

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/components/src/local/hex-grid/index.tsx
```diff
@@ -17,24 +17,17 @@ import multiPolyFromGeoJSON, { TerrainPolygons } from './helpers/multi-poly-from
...
 import { LAYDOWN_TURN } from '@serge/config'
 
 /* Render component */
 export const HexGrid: React.FC<{}> = () => {
+  const props = useContext(MapContext).props
+  if (typeof props === 'undefined') return null
   const {
     gridCells, planningConstraints, setNewLeg, setHidePlanningForm,
     selectedAsset, viewAsRouteStore, viewport, polygonAreas
-  }: {
```
IanMayo, line 28: @lilitkarapetyan - I thought I had to include the next block to get TS information for the props.  Don't I need to any more?

IanMayo:
Great piece of work @lilitkarapetyan 

I know it's still draft, but we have a couple of missing icons.  The first one is of type `Carrier`.  We do have a platform type for it, and it the `carrier.svg` seems to be in the correct place(s).

image

IanMayo:
@lilitkarapetyan - while you're in this PR, is it an option for you to remove one of the duplicate sets of icons?
image

(if not, we'll slip it to another issue)

IanMayo (commented):
Thanks for your fixes to my code, too :-)

packages/components/src/local/hex-grid/index.tsx
lilitkarapetyan: You can define there any types what you want but those props may can not mutch with real (inputed data -> MappingContext) types

https://github.com/serge-web/serge/pull/928/files#diff-42f0cdc7af2680f52905fe1b8f7f9429147aaa25c14346581c5544bc78c7269fL2
If you check this part you will se that we used any type, so if we will define props in right type (MappingContext) it will automatically works with already defined props, and will match inputed data, and when you will change MappingContext or imput data you will get right ts validation messages on erros

IanMayo: Aah, brilliant. Thanks :-)

lilitkarapetyan:
> 
> 
> @lilitkarapetyan - while you're in this PR, is it an option for you to remove one of the duplicate sets of icons?
> image
> 
> (if not, we'll slip it to another issue)

Done

IanMayo:
@lilitkarapetyan - I was expecting these hard-coded elements to be removed:
image

IanMayo:
One other issue @lilitkarapetyan 

In the `map-bar` the icon is in black, but on the `mapping` it's white.  I think the white background looks best.

image

IanMayo:
Ooh, @lilitkarapetyan  - I've found another issue.

We _currently_ use `SVG` to specify the icons in `asset-icon\counters` .  But, in the new `Platform Types` admin editor page, it invites us to provide a `png`.  I _think_ we should be providing a `svg`.  Is that right?

IanMayo:
Hmm, you may know about this.  I'm logged into `CmdWk` as `White Game Control`.  I'm on the `mapping` tab.  I get these 404 errors:
image

lilitkarapetyan:
> 
> 
> Hmm, you may know about this. I'm logged into `CmdWk` as `White Game Control`. I'm on the `mapping` tab. I get these 404 errors:
> image

This is because many resources don't have a valid url for images, I added a boolean that checks for images if the image url returns 404 and then we switch it to the old (hardcoded by class name) mode. In this way old code also works and we have working icons, until we will upload valid icons for all platformTypes

IanMayo:
> > Hmm, you may know about this. I'm logged into `CmdWk` as `White Game Control`. I'm on the `mapping` tab. I get these 404 errors:
> > image
> 
> This is because many resources don't have a valid url for images, I added a boolean that checks for images if the image url returns 404 and then we switch it to the old (hardcoded by class name) mode. In this way old code also works and we have working icons, until we will upload valid icons for all platformTypes

None of these icons are ones that have been provided by users in the admin page. They are all valid files, but it appears that Serge can't find them.

Hmm, for images like `datum.svg` we do have the files in the repo.  I guess we're just not looking for them in the right location. Aah, or the build process isn't putting them into the correct location.

Hmm, I _think_ the ones listed may be a new set of SVG icons that got added more recently.  No, that's true for `frigate-ta.svg` but not for `datum.svg`.  `datum.svg` has been in for a long time.

lilitkarapetyan:
> 
> 
> Ooh, @lilitkarapetyan - I've found another issue.
> 
> We _currently_ use `SVG` to specify the icons in `asset-icon\counters` . But, in the new `Platform Types` admin editor page, it invites us to provide a `png`. I _think_ we should be providing a `svg`. Is that right?

Updated

IanMayo:
I like the HSP work @lilitkarapetyan - we need to do the SVG color selection (light/dark) on the map, aswell as the map-bar:
image

IanMayo (approved):
Great, thanks.

PR #930, opened 2021-07-08
Added special channel dropdown

### LINKED ISSUE #605: Minor - UI to add "Special" channels.
Steps to reproduce:
- at home page, go into Admin pages (see below)
- select any wargame
- go to `channels` tab
- here we need UI to create a new `channel` with one of two specific titles

Here is an example of a set of channels: https://github.com/serge-web/serge/blob/develop/packages/mocks/game-channels.mock.ts

Note: Admin pages is accessed via "person" icon near top-right:
image

We have two "Special" channels:
- Mapping - for players that see/use the mapping control
- RFI Status - for players that will be managing/tracking RFIs.

The Serge application gives specific UI for these channels when it the current role is in that channel.  It does this by checking if the name matches one of these two strings:
image

Note: constants for the two special channels are in `consts.js` file.

These channels are currently generated by creating a Channel with the exact correct title.  But, it is fragile to expect the Game Designer to enter those specific strings for the new channel.

It would be more robust for Game Designers to be able to indicate if they're creating a "normal" channel or one of these "special" ones.  

Just like with "normal" channels, in these "Special" channels Roles get added (to indicate who sees them). But, for the "special" channels but the Game Designer does not indicate which messages (templates) get sent - so the Admin page for these channels could be slightly simpler.
image

Somehow - if we have one of these special channels we should hide the `templates` column.

_Maybe_ the `Add Channel` would become button with drop-down that lets the designer add a "Special" channel instead of a standard channel.
image

sanjayradadiya:
## 🧰 Issue

Fixes #605

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

image

## Confirmations


IanMayo:
Thanks @sanjayradadiya  - I know this is still in draft, but there is a rendering issue for the drop-down:
image

IanMayo (changes requested):
Thanks @sanjayradadiya  - I've had a play.

I want the "normal" channels to be what gets created by default.  So, I think we need Split Buttons, like here:
https://getbootstrap.com/docs/4.0/components/dropdowns/#split-button-dropdowns

Clicking on the main element would create a `Normal` channel, but clicking on the drop-down would have a menu with a `Special channels` [header](https://getbootstrap.com/docs/4.0/components/dropdowns/#menu-headers), the `Mapping` and `RFI Status Board`.  Clicking on those would give a channel with the relevant name.

IanMayo:
Hello @sanjayradadiya  -  could we please also have a disabled menu item that we can use as a menu header: `Special channels`.

IanMayo (approved):

PR #931, opened 2021-07-09
Store icon to server side and return url for preview

### LINKED ISSUE #899: Store URL for icon for force, not data
Game Designers can use the admin pages to specify a logo for a force.

But, in the wargame document the data is stored as a base64 string,  not a URL.  Serge should store the icon on the server, and store a URL in teh 

image

We should just have the iconURL for the force in the channel.

This fix involves both changing how the `admin pages` store the force icon, and how the `player pages` retrieve/display the icon.

* [ ] In TypeScript `force-data.d.ts` switch `icon` to `iconURL`
* [ ] In `Forces` tab, store icon to server, and store URL in wargame
* [ ] In player-UI, load icon from URL not base64 string
* [ ] Check above flow works when running application built with `yarn pkg-build` (to including assigning icon and display icon in UI)

Note: we should be backwards compatible.  Hmm, actually...  We don't need to display a icon if it's stored in `base64`, but the app shouldn't crash.

Note: once we have a working version, we'll need to check that the `pkg-build` version works (and stores/retrieves the icon correctly). @lebaphi is the expert on `pkg-build`.
Note: an image to use for testing can be found in: `packages\client\public\images\layers-2x.png`

sanjayradadiya:
## 🧰 Issue

Supports #899 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations

## 📝 Developer Notes:

I had also fixed the image upload and receive an issue when we create a pkg build so this would be work the same on both builds.


IanMayo:
@sanjayradadiya - I _don't think_ this completely resolves #899 

So, you need to use `Supports #899` rather than `Fixes #899` -  since we don't want merging this PR to close the issue.

sanjayradadiya:
> @sanjayradadiya - I _don't think_ this completely resolves #899
> 
> So, you need to use `Supports #899` rather than `Fixes #899` - since we don't want merging this PR to close the issue.

What's the issue you are facing? 

One more thing, Heroku is not supporting file upload in the server so what we need to do for that?

IanMayo:
> Heroku not supporting file upload

That's ok.  I'm happy for us to just test locally.  

BTW - did you understand the bit about `yarn pkg-build`?  We need to do a build, and test we can provide an icon, that it gets stored into `serge\img`, and rendered in the player UI.

IanMayo:
I'd like to "tag on" a couple more bits to this PR, please @sanjayradadiya 

I've changed the icon for Blue force: 
image

The icon is shown at the bottom-right of the PlayerUI, but against the wrong shade of blue:
image

IMHO, that icon should be larger, too.

We use icons to show the forces in each channel.  But, the icon doesn't contain the custom force icon:
image

IanMayo:
Ok, I've just tested in when running in the executable, and it fails.  When I changed the force logo, the new logo got pushed to the correct (`serge\img`) sub-folder, but the icon wasn't found by the node process:
image

sanjayradadiya:
> Ok, I've just tested in when running in the executable, and it fails. When I changed the force logo, the new logo got pushed to the correct (`serge\img`) sub-folder, but the icon wasn't found by the node process:
> image

We need to fix path issue while working with executable files

sanjayradadiya:
@IanMayo 
One more issue i am facing related to icon field because exiting db contain icon fields and from now we are using `iconURL` so do we need to add some migration logic there or we can skip that part and only accept iconURL from onwards ?

IanMayo:
Thanks for that Sanjay - good point.

How about we keep `icon` in the TypeScript definition, but use `jsdoc` to mark it as deprecated?

To give backward compatibility I suggest read the `iconUrl` value if it's present, else we will look for an `icon` value.

Does that sound ok?

sanjayradadiya:
> Thanks for that Sanjay - good point.
> 
> How about we keep `icon` in the TypeScript definition, but use `jsdoc` to mark it as deprecated?
> 
> To give backward compatibility I suggest read the `iconUrl` value if it's present, else we will look for an `icon` value.
> 
> Does that sound ok?

yeah, absolutely

IanMayo:
Aah, @sanjayradadiya - I've just seen that Lilit is also working on icon-handling code:
https://github.com/serge-web/serge/pull/928#issuecomment-878799321

I think I should merge her changes before we finalise this PR.

packages/components/src/local/molecules/admin-message-creator/index.tsx
```diff
@@ -30,7 +30,7 @@ export const AdminMessageCreator: React.FC<Props> = ({ from, channel, role, post
           force: from.name,
           forceColor: from.color,
           role: role,
-          icon: from.icon ? from.icon : ''
+          iconURL: from.iconURL || (from.icon || '')
```
IanMayo, line 33: Aah, I think there is an issue here @sanjayradadiya 

In the legacy games I _think_ that the `icon` is the Base64 string representation of the icon itself.  But, `iconURL` is always the URL.  In the above, is there a change that we may put the Base64 icon into the `iconURL` field, and cause later problems?

sanjayradadiya: It doesn't cause any problem because it will accept base64 first time but later on when user change it icon from admin side it will be replaced  by URL so it will work on both side.

IanMayo: Great, thanks.

IanMayo (approved):

PR #932, opened 2021-07-09
Reduce transparency of channels

### LINKED ISSUE #927: Reduce transparency of channels
Ian realised that `maximise view` could be a useful tool for players keeping track of lots of channels.

But, the semi-transparent channels show the content underneath, which gives a confused UI.

image

We should reduce the transparency of channels, so the channels behind the maximised one are less (or not) visible.

lebaphi:
Fixes #927


IanMayo:
Thanks @lebaphi . In the future could you put some description of the issue into the branch name too, please.  It helps us to navigate the repo.
image

IanMayo:
Not quite there yet @lebaphi 

The `below` tabs and are still visible, which could make a player try to click on them:
image

lebaphi:
> Not quite there yet @lebaphi
> 
> The `below` tabs and are still visible, which could make a player try to click on them:
> image

Hi @IanMayo , it's just a draft PR, so I can hide the content of the tab is below the maximize tab (mapping tab). I think we can add some CSS to hide the tab header but it still does not work 🤔 
image
 
image

lebaphi:
Hi @IanMayo , I think it's ready to test 👍

IanMayo:
Thanks @lebaphi 

The deployed version works perfectly :-)

With the change to `factory.tsx`, do we still need the changes to `tabRender`? 

If other tabs aren't being rendered, it doesn't seem necessary to change the CSS/z-order.

lebaphi:
> Thanks @lebaphi
> 
> The deployed version works perfectly :-)
> 
> With the change to `factory.tsx`, do we still need the changes to `tabRender`?
> 
> If other tabs aren't being rendered, it doesn't seem necessary to change the CSS/z-order.

Hi @IanMayo, yes we still need the changes in tabRender, because the changes in factory does not render the node content, but the node header and and minimize button and the background color is still rendered. We need to hide it in the tabRender too.

IanMayo:
> > Thanks @lebaphi
> > The deployed version works perfectly :-)
> > With the change to `factory.tsx`, do we still need the changes to `tabRender`?
> > If other tabs aren't being rendered, it doesn't seem necessary to change the CSS/z-order.
> 
> Hi @IanMayo, yes we still need the changes in tabRender, because the changes in factory does not render the node content, but the node header and and minimize button and the background color is still rendered. We need to hide it in the tabRender too.

Perfect answer, thanks :-)

IanMayo (approved):
Working fine, thanks.

PR #935, opened 2021-07-12
Fix issue `Unread message count in Tab overflow menu`

### LINKED ISSUE #906: Unread message count in Tab overflow menu
Some users need to track new messages across lots of channels.

If there are too many channels to view at once, the other fall into an overflow menu:
image

This menu doesn't include an indication of if there are unread messages in that channel.

It would be great if the channel name could be followed by ` (4)` - to indicate 4 unread messages.

_Maybe_ if the overflow tabs used a _formatting function_ instead of just the tab name. We could look up the unread message count, and provide that in the formatting function.

Aah, this looks like what we're looking for:
image

(above game [here](https://serge-review.herokuapp.com/?wargame=wargame-kqgy546g&access=pkoiqu5ev) )

lebaphi:
Fixes #906 
Screen Shot 2021-07-12 at 10 34 32 PM

### DIFF
--- a/packages/client/src/Views/ChannelTabsContainer/helpers/tabRender.tsx
+++ b/packages/client/src/Views/ChannelTabsContainer/helpers/tabRender.tsx
@@ -3,12 +3,28 @@ import { PlayerUi } from '@serge/custom-types'
 import _ from 'lodash'
 import findChannelByName from './findChannelByName'
 
-const tabRender = (state: PlayerUi): (node: TabNode) => void  => {
+const tabRender = (state: PlayerUi): (node: TabNode) => void => {
   return (node: TabNode): void => {
-    let channel;
+    let channel: any;
+
+    const addMenuItemMsgCount = (className: string) => {
+      const overflowBtn = document.getElementsByClassName('flexlayout__tab_button_overflow')
+      if (overflowBtn.length) {
+        overflowBtn[0].addEventListener('click', () => {
+          setTimeout(() => {
+            const menuItems = document.getElementsByClassName('flexlayout__popup_menu_item')
+            Array.from(menuItems).forEach((menuItem: Element) => {
+              if (menuItem.textContent === node.getName()) {
+                menuItem.classList.add(className)
+              }
+            })
+          });
+        })
+      }
+    }
 
     const setUnreadClassName = (className: string): void => {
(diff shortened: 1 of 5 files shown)


IanMayo:
Thanks for this @lebaphi  :-)

Could you explain why you had to use CSS styling for the numbers, instead of the `onRenderTab()` method?

It's a clever strategy, but I'm just double-checking the complexity is justified.

lebaphi:
> Thanks for this @lebaphi :-)
> 
> Could you explain why you had to use CSS styling for the numbers, instead of the `onRenderTab()` method?
> 
> It's a clever strategy, but I'm just double-checking the complexity is justified.

hi @IanMayo , I just follow the same way as the function `setUnreadClassName` (in `onRendereTab()`) did. We have the CSS class name `unread-<number-unread-message>`, and we update the node unread message count by calling: `node.getModel().doAction(...)`. In this issue, it's just the menu item, (not a node element), so I get each item and add the class to style to unread message count. I can do this by javascript but I think using the CSS styling is better.

packages/client/src/Views/ChannelTabsContainer/helpers/tabRender.tsx
```diff
@@ -20,6 +36,10 @@ const tabRender = (state: PlayerUi): (node: TabNode) => void  => {
 
       channel = matchedChannel && matchedChannel.length > 1 ? matchedChannel[1] : undefined
 
+      setTimeout(() => {
+        addMenuItemMsgCount(`unread-${channel.unreadMessageCount}`)
+      })
```
IanMayo: does the above line need to be modified to return `9plus` if there are 9 or more unread messages?

Aah, or we modify the method that calculates `channel.unreadMessageCount`

lebaphi: ah not yet Ian, do you want to show 9+ or the real number unread message count here? e.g: (12) or (23)?

IanMayo: 9+ is fine, thanks.

lebaphi:
image
Fixed @IanMayo, the unread message count work stable now 👍

IanMayo (approved):

PR #936, opened 2021-07-15
fix player ui permissions issue

### LINKED ISSUE #848: Channel memberships update live, but not role permissions
During a game we can make live changes to which channel a user in.

But, when we decide channel membership, we only use the role permissions that were loaded when the game started.

So, we will need to either:
1. retrieve the `live` role information and pass it to `handleMessageUpdates`
2. update role details on new wargame update

Actually, the second makes more sense - since other parts of the UI will probably need to know the updated role permissions (such as `canSubmitOrders`, `insightViewer` or `rfiManager`)

Steps to reproduce:
* edit wargame, don't give Game Control permission to release RFIs (RFI Manager)
* log in as `Blue Force` \ `CO`, create new RFI message in `Blue RFIs`
* log in as `Game control`. Go to `RFIs` channel
* take ownership of new RFI
* provide response
* press `Send for review`
* see no `Release` or `Reject` buttons present
* edit wargame, give Game Control `isRFIManager` permission. Save wargame
* note that `Release` & `Reject` buttons still not present
* refresh page
* note that `Release` & `Reject` buttons now present

The issue is that the client received the new `Forces` object when it got saved, but the Player State object didn't get updated.

I _guess_ that `SET_ROLE` needs to be called when `SET_FORCE` does.  Hmm, maybe not.

lilitkarapetyan:
## 🧰 Issue

closes #848 

## 🤔 Reason: 

Used find function instead of filter as result we get only first matched "Participants and messages" messageTemplates for selected Role

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

изображение

## Confirmations


packages/helpers/src/participant-states.ts
```diff
@@ -52,12 +53,19 @@ export const getParticipantStates = (channel: ChannelData, forceId: string | und
...
   const {
     isParticipant,
-    participatingRole,
+    participatingRoles,
     allRolesIncluded
   }: CheckParticipantStates = checkParticipantStates(channel, forceId, role, isObserver)
 
-  if (participatingRole) {
-    chosenTemplates = participatingRole.templates
+  if (isParticipant) {
```
IanMayo, line 70: You've done really well with this logic @lilitkarapetyan - there's some complex stuff going on.

In the next block, it "looks like" `chosenTemplates` will contain the templates for all of the participating roles in the channel.  I _think_ it should only contain the templates for _my_ role.  So, `Blue` could have some templates and `White` could have others.

IanMayo:
@lilitkarapetyan - this is almost certainly a flaw in game design, but I think it's worth removing duplicates from the available templates:
image

(you can see the wargame, force and role in the above screenshot in order to reproduce the issue).

Well not exactly a game design flaw - it _could_ have been intentional in some way. Here's the channel design:
image

packages/helpers/src/participant-states.ts
IanMayo: No, I was mistaken.  The chosenTemplates is the union of the ways this role can participate in the channel, which is correct :-)

IanMayo:
@lilitkarapetyan - the original flaw in the logic is still present.  (but thanks for the fixes in this PR :-) ).

Have provided steps to reproduce in the original Issue: #848

lilitkarapetyan:
> 
> 
> @lilitkarapetyan - the original flaw in the logic is still present. (but thanks for the fixes in this PR :-) ).
> 
> Have provided steps to reproduce in the original Issue: #848

Fixed

lilitkarapetyan:
> 
> 
> @lilitkarapetyan - this is almost certainly a flaw in game design, but I think it's worth removing duplicates from the available templates:
> image
> 
> (you can see the wargame, force and role in the above screenshot in order to reproduce the issue).
> 
> Well not exactly a game design flaw - it _could_ have been intentional in some way. Here's the channel design:
> image

изображение
Fixed

packages/client/src/ActionsAndReducers/playerUi/playerUi_Reducer.ts
```diff
@@ -184,6 +181,11 @@ export const playerUiReducer = (state: PlayerUi = initialState, action: PlayerUi
     default:
       return newState
   }
```
lilitkarapetyan, line 183: Log For player ui state update only for dev mode

packages/helpers/src/handle-channel-updates.ts
```diff
@@ -195,21 +202,25 @@ export const handleAllInitialChannelMessages = (payload: Array<MessageInfoType |
     }
 
     if (isObserver || isParticipant || allRolesIncluded) {
+      // TODO: define type for force Icons
+      const forceIcons: any[] = []
+      const forceColors: string[] = []
+      for (const { forceUniqid } of channel.participants) {
```
lilitkarapetyan, line 208: single loop instead of 2 loops

IanMayo (approved):
Great - nice improvements to the general codebase, too :-D

PR #940, opened 2021-07-16
Fixed default date entry in json editor

### LINKED ISSUE #916: Inclusion of ‘default’ entries (eg 01/01/2025 for date boxes)
Discussed with Mark on 30/6/21

The intention for this was to overcome the issue of players mistkenly entering dates that were several days/months away from the intended time.

The solution is to provide the date-time of the start of the current turn as the default value for any date-time field.

For some field representing a time that doesn't occur within the current turn, the control will still have the rough date-time, which should still help with accuracy.

To implement this for messages from templates, we _may_ have to inspect/mangle the form template, to look for a text field formatted as date-time, then to `inject` the current date-time (`gameTurnTime`) as a default value.

sanjayradadiya:
## 🧰 Issue

Fixes #916 

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

image (13)

## Confirmations

## 📝 Developer Notes:

default date introduces to JSON editor template and format will be dd/mm/yyyy hh:mm with validation using flatpickr calendar.

flatpickr calendar uses the "window" object during initialization in the library itself which is generated issues for admin message template preview. avoid this issue window.flatpickr object set to undefined in JsonCreator component.

flatpickr property modified in code such as wrap,time_24hr,dateFormat which will be replaced once we changed this property in database for message template.


IanMayo:
Hello @sanjayradadiya  a jest snapshot failed.

I don't know how familiar you are with them.  We have a snapshot in the repo that gets compared with the rendered code.  In this case the rendered `rfi-status-board` doesn't match the snapshot.  

Hmm, but your PR should not have changed how the status board gets rendered.  Could you have a look at the logging, and see if you can understand why the rendered code doesn't match?
https://github.com/serge-web/serge/pull/940/checks?check_run_id=3086635185

IanMayo:
Good start. But, it's showing the current date.  It should show the current `gameTurnTime`. In this game it's
 (
image

https://serge-916-inclusion-of--plpcil.herokuapp.com/?wargame=wargame-kqgy546g&access=pkoiqu5ev

packages/client/src/Components/JsonCreator.jsx
```diff
@@ -18,7 +18,15 @@ class JsonCreator extends Component {
       selectedSchema: ''
     };
   }
-
+  
+  /*
+  ** avoid loading of flatpickr load in Message template screen for admin
```
IanMayo, line 23: @sanjayradadiya - could you expand this comment a little please?

(Or reply to this comment with more explanation if that's easier :-) )

sanjayradadiya: flatpickr calendar uses the "window" object during initialization in the library itself which is generated issues for admin message template preview. avoid this issue window.flatpickr object set to undefined in JsonCreator component.

packages/client/src/Components/MessageCreator/MessageCreator.tsx
```diff
@@ -83,6 +85,41 @@ const MessageCreator: React.FC<Props> = (props) => {
   }, [props])
 
   const createEditor = (schema: any) => {
+    /*
+    ** Render Default datetime entries in template of json for type datetime-local
```
IanMayo: @sanjayradadiya - this looks like it loops through the top level properties in the schema.

But, the date may be in a child object.  So, I think we actually need to "walk" the schema if the property is of `"type": "object"`:
image

packages/client/src/Components/JsonCreator.jsx
IanMayo: Aah, I see, thanks.

IanMayo:
One more thing, please @sanjayradadiya - it is several data types that need to be given the default value:
https://github.com/serge-web/serge/issues/916#issuecomment-877081843

sanjayradadiya:
image
image

Introduce default date and time of gameTurnTime for JSON schema editor.

packages/client/src/Components/MessageCreator/MessageCreator.tsx
```diff
@@ -82,7 +85,129 @@ const MessageCreator: React.FC<Props> = (props) => {
...
+  /**
+   * Render Default datetime entries in template of json for type datetime-local
+   */
+  const configDateTimeLocal = (schema: any) => {
+    if(!schema || !schema.properties){
+      return
+    }
+    Object.keys(schema.properties).forEach(key => {
+      let prop = schema.properties[key]
+      if(prop.format === 'datetime-local' || prop.format === 'date' || prop.format === 'time'){
```
IanMayo: @sanjayradadiya  - we to string comparison above, then we do it again in the switch block in `configCommonProps `

I'm pretty sure we can remove this `if` construct, can't we?

packages/client/src/Components/MessageCreator/MessageCreator.tsx
```diff
@@ -82,7 +85,129 @@ const MessageCreator: React.FC<Props> = (props) => {
...
+        prop = configCommonProps(prop)
+      }
+      
+      if(prop.type === 'object'){
+        configDateTimeLocal(prop)
+      }else if(prop.type === 'array'){
+        configDateTimeLocal(prop.items)
```
IanMayo: in the above line, we pass in an array of items.  Will this still work with the `configDateTimeLocal` method?  (it looks like that only processes one line at a time)

packages/client/src/Components/MessageCreator/MessageCreator.tsx
```diff
@@ -82,7 +85,129 @@ const MessageCreator: React.FC<Props> = (props) => {
...
+        prop.options = {"flatpickr": {
+          "wrap":false,
+          "time_24hr": true,
+          "dateFormat":"d/m/Y H:i",
+        }}
+        return prop
+      case "date":
+        prop.default = moment(gameDate).format("DD/MM/YYYY")
+        prop.options !== undefined ? prop.options.flatpickr = flatpickr(".calendar") : prop.options = {}
+
```
IanMayo: @sanjayradadiya  - doesn't line 106 overwrite any changes made in line 104?

(same in other catch blocks)

IanMayo:
Aah, I missed one item @sanjayradadiya - you need to look out for `datetime` as well as `datetime-local`, `date`, and `time`:
https://github.com/json-editor/json-editor#format

packages/client/src/Components/MessageCreator/MessageCreator.tsx
sanjayradadiya: Yes, you are right.
I have two options at that time.
1) function should be called every time irrespective of format
2) function should be called only if we have format datetime or date or time
I choose option 2 and I will change it to option 1.

IanMayo:
@sanjayradadiya - we still have a bit to do here, please:
https://github.com/serge-web/serge/pull/940#issuecomment-884932925

packages/client/src/Components/MessageCreator/MessageCreator.tsx
sanjayradadiya: https://github.com/serge-web/serge/blob/develop/packages/client/src/Schemas/DailyIntentions.json#L3
https://github.com/serge-web/serge/blob/develop/packages/client/src/Schemas/DailyIntentions.json#L15

Daily Intension schema template has type array and object as "properties" which will work in both cases.

IanMayo: Yes, but don't we need to provide iterator for `array`?

sanjayradadiya: image
image

Json Schema Editor specification mentioned that {"type":"array"} will take array in "items.properties" which is ultimately an object which is mentioned in image 1.

IanMayo: Aah, I see.  That's fine then :-D

packages/client/src/Components/MessageCreator/MessageCreator.tsx
sanjayradadiya: option 2 changed to option 1 in latest commit.

packages/client/src/Components/MessageCreator/MessageCreator.tsx
sanjayradadiya: yes, it's overwriting. changes related to this problem were applied in the latest commit.

IanMayo:
Hello @sanjayradadiya   - I've modified the `weather` template in `TallMsg` to include a range of date and datetime control types.  You'll need to pull the branch to get the new `weather` template.

`datetime-local` is working, but not `datetime`.   No popup appears.  The others all work fine :-)

image

sanjayradadiya:
> Hello @sanjayradadiya - I've modified the `weather` template in `TallMsg` to include a range of date and datetime control types. You'll need to pull the branch to get the new `weather` template.
> 
> `datetime-local` is working, but not `datetime`. No popup appears. The others all work fine :-)
> 
> image

HTML5 supports only date, time, datetime-local which is mentioned in the below link.
	https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

@json-editor/json-editor has mentioned datetime and datetime-local in the below link.
	https://github.com/json-editor/json-editor#format

both datetime and datetime-local are pointed to the same format like "datetime-local" mentioned in the below link. they didn't use "datetime" format anywhere.
	Code: https://github.com/json-editor/json-editor/blob/master/docs/datetime.html#L156
	View: https://json-editor.github.io/json-editor/datetime.html

IanMayo:
Hey, well-researched @sanjayradadiya  :-)

Could you remove the (mistaken) support for `datetime` in the PR please, and I'll get it merged?

IanMayo (approved):

Issue #873, opened 2021-06-25
If only one wargame present, select it


IanMayo:
A Game Designer may wish to make it easier for the players to get into the game, by deleting other wargames:
![image](https://user-images.githubusercontent.com/1108513/123452236-20816280-d5d6-11eb-9ba6-c3c8aa4963d5.png)

When this happens, we should automatically select that wargame. Selecting it like this would also reveal the access roles (if they're switched on).

PR #941, opened 2021-07-17
As a Game Control, I wish to be able to hide games

### LINKED ISSUE #900: As a Game Control, I wish to be able to hide games
A game control may be working with a number of wargames, and may choose to hide some snapshots (or old games) from the players - to ensure the players go into the correct wargame.

The list of games in the admin page could have a checkbox to indicate which ones are presented to players.

I _guess_ we would store this in the `pouchdb_all_dbs` collection.  The alternative is to have a `visible` element in the wargame, but we'd have to load/parse all the wargames in order to show the welcome page.

This requires a degree of familiarity with PouchDb.

Sadly it looks like it isn't possible via `_all_dbs`.  I suspect we could only achieve this via a custom table of databases.

I guess an alternate strategy is to use a leading underscore as a `hidden` marker.  So, we rename hidden games to have a leading underscore. That's not all that bad - since we won't be expecting users to load/play a hidden game.

So, in the admin UI we could provide a `visible` checkbox in the wargame listing.  Clicking this checkbox (or eye icon) would (secretly) add/remove a leading underscore for that wargame.

lebaphi:
Fixes #900


IanMayo:
Thanks for this @lebaphi :-)

I've put a note in Slack about future PRs (not this one) - about helping my by trying to keep them focussed on the new feature - and include bulk linter/tidying for separate PRs.

IanMayo (approved):

IanMayo (changes requested):
Aah, I do have a change request, please @lebaphi 

In Game Admin, I reduced the games down to just one.

When I refreshed the Player UI, just one game was shown.

But, we've introduced the logic where if just one game is present, the UI _loads_ that game (#873).

Could you investigate what that isn't happening, please?

lebaphi:
> Aah, I do have a change request, please @lebaphi
> 
> In Game Admin, I reduced the games down to just one.
> 
> When I refreshed the Player UI, just one game was shown.
> 
> But, we've introduced the logic where if just one game is present, the UI _loads_ that game (#873).
> 
> Could you investigate what that isn't happening, please?

okay @IanMayo

lebaphi:
@IanMayo I removed some code refactoring that does not relate to this feature. So it's easier for you to review. We may have another PR to do the code refactor or clean code.

lebaphi:
> > Aah, I do have a change request, please @lebaphi
> > In Game Admin, I reduced the games down to just one.
> > When I refreshed the Player UI, just one game was shown.
> > But, we've introduced the logic where if just one game is present, the UI _loads_ that game (#873).
> > Could you investigate what that isn't happening, please?
> 
> okay @IanMayo

Fixed @IanMayo

packages/client/src/Views/PlayerUi/PlayerUi.tsx
```diff
@@ -78,6 +79,10 @@ const PlayerUi = ({ gameInfo, wargame, messageTypes, checkPasswordFail, loadData
 
 
   const handleCheckPassword = (pass: string): void => {
+    if (currentWargame.startsWith(hiddenPrefix)) {
```
IanMayo, line 82: Hello @lebaphi - this test shouldn't fail, should it?

The hidden wargame shouldn't be listed, so it can't be present as `currentWargame`, can it.

I _guess_ this is handling some edge case.  Based on it being an edge case, I suggest the notification at line 21 below be changed to:
`Error: hidden wargame should not be available. Not opening.`

lebaphi: Hi @IanMayo, the hidden wargame shouldn't be listed, right. This handles the case:
- in the client-side, the user loads the list of 3 games
- at this time on the server-side, the admin hide 1 of 3 games
- in the client-side, the user does not reload the page and chooses the hidden game then press enter.

=> we should handle this case and show an error message. I will update the error message.

IanMayo (approved):

PR #942, opened 2021-07-17
Introduce uniqid for Role

### LINKED ISSUE #812: Introduce uniqid for Role
We currently use the role title as the identifier when referred to from channel definitions.

But, the role title can be renamed, which would break the logic.

Instead, when creating a role a persistent uniqid should be generated, which would handle role name changes.

**Update:** Aah, actually we already create a unique id with each role - the passcode/password. So, we could rename `password` to `roleId`

* [x] Change `Role` type by renaming `password` to introduce `RoleId` (`role.d.ts`)
* [x] Change to `Channels` page, to store roleId for channel participations, rather than role name (but still display human-readable name in UI).  
* [x] Change to welcome page, to work with role-id values rather than role name 
* [x] Identify locations where `role-name` still used, and work through them.   One will be the `reducer` than handles new wargame.  It will have to retrieve the role data for the current user, then set `observer`, `canSubmitOrders`, `isRFIManager`.

lebaphi:
Fixes #812 

## Updates:
- add `roleId` when creating a new role
- using `roleId` instead of `roleName`
## Backward compatible: 
- if `roleId` does not exists, it uses the `roleName` instead


packages/client/src/Views/InsightsChannel.tsx
```diff
@@ -31,7 +31,7 @@ const InsightsChannel = (): React.ReactElement => {
       <MessagesListRenderProp
         curChannel={'feedback_messages'}
         messages={state.feedbackMessages}
-        userId={`${state.wargameTitle}-${state.selectedForce}-${state.selectedRole}`}
+        userId={`${state.wargameTitle}-${state.selectedForce}-${state.selectedRoleName}`}
```
IanMayo, line 34: @lebaphi it _looks_ like this is expecting an `id` rather than a role name

IanMayo:
Hello @lebaphi - I'm afraid this PR is too large for us to handle in one go.

I expect the code is right, but there's just too much to review.  We don't want to break any functionality, but there is too much code to do that by eye.

Can you think of a simpler approach?

Maybe a series of PRs:
- make changes to admin page, and put both `role` and `role-id` into wargame document
- make change to player ui reducer, to reproduce existing state by first trying for role-id, then fallback on role
- make change to TypeScript definitions, to _add_ new `role-id` element in relevant structures
- make change to TypeScript definitions to _remove_ legacy `role` element

You should be able to copy/paste code from this PR to allow the above, but it means we can move forward one step at a time.

Oh, and as before, I'd like to separate lint-style refactoring from functionality changes, if possible.

lebaphi:
> Hello @lebaphi - I'm afraid this PR is too large for us to handle in one go.
> 
> I expect the code is right, but there's just too much to review. We don't want to break any functionality, but there is too much code to do that by eye.
> 
> Can you think of a simpler approach?
> 
> Maybe a series of PRs:
> 
> * make changes to admin page, and put both `role` and `role-id` into wargame document
> * make change to player ui reducer, to reproduce existing state by first trying for role-id, then fallback on role
> * make change to TypeScript definitions, to _add_ new `role-id` element in relevant structures
> * make change to TypeScript definitions to _remove_ legacy `role` element
> 
> You should be able to copy/paste code from this PR to allow the above, but it means we can move forward one step at a time.
> 
> Oh, and as before, I'd like to separate lint-style refactoring from functionality changes, if possible.

I understand your point @IanMayo, this PR is too large and very risky to merge. I will find another approach with smallest changes 🤔

IanMayo:
> I understand your point @IanMayo, this PR is too large and very risky to merge. I will find another approach with smallest changes 🤔

Great, thanks @lebaphi

PR #968, opened 2021-07-27
Introduce uniqid for Role

### LINKED ISSUE #812: Introduce uniqid for Role

lebaphi:
Supports #812


packages/components/src/local/organisms/setting-forces/settings-force-roles/index.tsx
```diff
@@ -98,6 +98,7 @@ export const RolesAccordion: FC<PropTypes> = ({ data, handleChangeForce }) => {
 
   const handleCreateRole = (): void => {
     const roles: Array<Role> = [...data.roles, {
+      roleId: 'r' + Math.random().toString(36).substring(4), // make id more uniq
```
IanMayo: @lebaphi - elsewhere in the app we use the `uniqid` library: https://github.com/adamhalasz/uniqid/

The above would be replaced with `uniqid.time('r')`

packages/mocks/atlantic-forces.mock.ts
```diff
@@ -309,6 +310,7 @@ export const forces: ForceData[] = [
     "overview": "Blue force.",
     "roles": [
       {
+        "roleId": "r12345CO",
```
IanMayo: @lebaphi - I _think_ it would be healthy for the three `roleId` values in this data were unique.

packages/client/src/Store/PlayerUi.tsx
```diff
@@ -21,6 +24,21 @@ export const usePlayerUiState = (): PlayerUi => {
   if (context === undefined) {
     throw new Error('usePlayerUiState must be used within a PlayerStateContext.Provider')
   }
+  //process roleid for the legacy data
```
IanMayo, line 27: I could do with some help here please, @lebaphi 

This happens as we're loading a game. We see if the roleId is missing. If it is missing, we create the roleId for that role, then we push the forces to the database.

Do we need to worry about multiple players updating the roles at the same time, and getting a conflict?

lebaphi: Hi @IanMayo , I just wonder if multiple users adding the new role or updating the role at the same time, how do we deal with conflicts?

IanMayo: Aah, we don't want to allow multiple admins to create roles at the same time - I _Think_ that's to be treated as a training issue.  The issue is when multiple players load an old wargame at the same time, and separately try to generate the roleId entries.

lebaphi: ah okay, so I think we need to handle the conflict error for this thing, no worries 😙

IanMayo: I don't know for sure - I'm just thinking it through with you.

IanMayo: Let me look at the network traffic & see how long it takes to process with two players logged in.  If it's well under a second, then I don't think the conflict will happen.

IanMayo: It _seems to take 0.91ms to run - so I don't think there will be conflicts.

lebaphi: yep, it seems unlikely to occur, but if you want to make sure there is no conflict, I think we should handle the conflict error, if it does, just update this document again https://pouchdb.com/guides/conflicts.html

IanMayo: Oh yeah - PouchDb will tell us if there is a conflict. Yes, in this instance I think we should just `put` it again.

lebaphi: cool, I will update soon 👍🏻

IanMayo: Hello @lebaphi - I could do with closing this soon please, since in my code I want to use `roleId` instead of `name`.

If you're not ready to handle the conflict today, then let's merge the PR and solve it as another issue.

lebaphi: HI @IanMayo , okay you can merge this PR first, then I can have another PR to resolve the conflict issue later.

IanMayo (dismissed):
Approving. Will slip potential conflict resolution to future PR.

package.json
```diff
@@ -13,7 +13,8 @@
     "@types/react-flatpickr": "^3.8.4",
     "lerna": "^3.16.4",
     "npm-run-all": "^4.1.5",
-    "snyk": "^1.298.1"
+    "snyk": "^1.298.1",
+    "uniqid": "^5.3.0"
```
IanMayo: @lebaphi - shouldn't `uniqid` be added to the `package.json` for `packages\client` ?

lebaphi: hi @IanMayo, I don't add `snyk `, just add `uniqid `, it shows the code changes :D

IanMayo: It's ok. I'm pretty sure I can remove it, since the dependency is already in the `client` package.

IanMayo (approved):

PR #969, opened 2021-07-27
Swap templates

### LINKED ISSUE #869: Swap templates for templateIds in channels
Early in the project we took a design decision to store the template definitions in the `channel` objects.  This means the `wargame` document contains all the data it needs.

But, we have a wargame with lots of long formatted messages. meaning the `channels` part of the wargame document is 34,000 lines long!

This can make the time to transmit (and parse) the wargame document longer, which could introduce the chance of `collisions` - which would crash SQLite.

There is also the down-side that if a message template layout is updated, the change isn't reflected in the wargame - since it will just use the version in place when the channel was created.

So, we should switch from storing the template to storing the template id (plus possibly the template name).  Storing the name means we can use a single template for different message types if the layout is the same.

If possible, I would like to keep some backward compatibility.  So, we can read/display wargames from the period when the wargame contained the template.  But, the game should ignore the template definition in the wargame and use the one from the `message-types` database.

lilitkarapetyan:
## 🧰 Issue

closes #869

## 🤔 Reason: 

Define types: For all project templates defined as `any` type
Update tests
Convert Reducer and Actions to TS
New object of message TypesByKey and remove loop findes
Currently participant templates have only _id and title fields (same fields have old version if templates, so old version will works too)

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

изображение
изображение

## Confirmations


packages/custom-types/message-types.d.ts
```diff
@@ -0,0 +1,12 @@
+import { TemplateBody } from "./template";
+
+export interface TemplateBodysByKey { [property: string]: TemplateBody }
+
+interface MessageTypes {
```
IanMayo, line 5: @lilitkarapetyan - we have two elements with very similar names here.   Can you explain the difference between them please?

IanMayo:
Great piece of work, thanks @lilitkarapetyan 

Is it backwards compatible for our existing wargames?

packages/custom-types/message-types.d.ts
lilitkarapetyan: TemplateBody[] - it's array of templates. if you want get some template from there you need to do loop find
TemplateBodysByKey - it's object of templates. if you want to get some template from there you need to do TemplateBodysByKey[yourTemplateId] and you will get it without any loop findes

lilitkarapetyan (commented):
> 
> 
> Great piece of work, thanks @lilitkarapetyan
> 
> Is it backwards compatible for our existing wargames?

Yep
изображение

packages/custom-types/message-types.d.ts
IanMayo: I should have been more clear 🤣 

I was talking about `loadingDb` and `isLoading` in the next two lines.

packages/client/src/ActionsAndReducers/dbMessageTypes/messageTypes_Reducer.ts
```diff
@@ -0,0 +1,32 @@
+import * as ActionConstant from '@serge/config'
+import copyState from '../../Helpers/copyStateHelper'
+import { MessageTypes, MessageTypesActionTypes } from '@serge/custom-types'
+
+const initialState: MessageTypes = {
+  loadingDb: false,
+  isLoading: false,
+  messages: [],
+  messagesByKey: {}
```
IanMayo: is this actually `templatesByKey` rather than `messagesByKey`?  It looks like a list of templates, rather than a list of messages.

IanMayo: Hmm I think that `messages` should also be `templates`.

Aah, it's short for _message-types_, which are also called _templates_.  Could we switch to `templates`, to avoid confusion with lists of messages?

IanMayo: Here is an example of where we could get confusion:
image

IanMayo:
Hello @lilitkarapetyan - one last question, please.

Why do we store template `title` _and_ `_id`?  Do we ever access the title?  Aah, is it used for rendering the list of channels in admin?

IanMayo (approved):

PR #971, opened 2021-07-27
RFI Response lost if UI updates [latest]

### LINKED ISSUE #944: RFI Response lost if UI updates
Issue #892 covered an issue where partially completed templates were lost when the UI updated (such as when receiving a new message).

This issue is also present in the RFI Editor page.

So, when a user has "taken ownership" of an RFI, and has partially completed their response, when a new message is received (such as a new Game Admin message), the RFI form updates, and the response is lost.

The form should not be reset.

Steps to reproduce:
* Login to game `CMDWK-HIGH DATA` as `Game Control`
* Navigate to RFIs page
* Open the `In Progress` RFI titled `Third Blue RFI`
* Write some text into the `Answer` box, such as `sample response`
* Write text into `Game Admin` channel, press enter
* The `Sample response` is lost

Note: the data for the above scenario won't be present until PR #945 merged

lebaphi:
Fixes #944


lebaphi:
The same codebase but the deployment works @IanMayo 👍

IanMayo (approved):

PR #981, opened 2021-07-30
977: Drop passcode in role, use role-id

### LINKED ISSUE #977: Drop passcode in role, use role-id
We've added role-id for roles.

We can now drop the `password` - and use `roleid` in every place that was used.  I can think of:
* [x] Password box for `forces` admin page
* [x] Injecting the password in the quick-login links on the welcome page
* [x] Checking the password is valid for a user at login page
* [x] Short-cutting into the app via the `access` parameter in the URL

We should lastly remove `password` from the `Role` object.

lebaphi:
Fixes #977


lebaphi:
HI @IanMayo , I remember that we want to do this https://github.com/serge-web/serge/issues/812#issuecomment-871275356. Do you think we should make changes to this PR?

IanMayo:
Thanks for spotting that @lebaphi - nice work :-)

Let's cover that in a separate issue/PR, please.   Actually, I _think_ it's partially implemented, but it's currently broken.  Hmm. Let's not make any changes in `Settings - Channels` because @lilitkarapetyan has a large branch open on it.  Issue for broken roles in #982

IanMayo (approved):

Issue #985, opened 2021-07-30
Early steps in Collaborative Editing UI


IanMayo:
Supports #960 

* [x] Copy existing RFI Status Board components, renaming `RFI Status` to `Collab Working`
* [x] Copy existing RFIMessageDetail components
* [x] Use new `format` property to determine channel type
* [ ] create suitable `CollaborationController` entity
* [ ] (rough) logic to drive `CollaborationController` logic
* [ ] test-drive with different roles (probably via different instances in StoryBook)
* [ ] Allow user to create new document if controller provides templates
* [ ] Show correct roles for `Assign-to`
* [ ] Use JSON-editor for collab working
* [ ] Allow JSON-editor to be editable for `collab-edit` channels
* [ ] Make JSON-editor read-only for `collab-RFI` channels
* [ ] Provide `Response` field for `collab-RFI` channels
* [ ] Invite user to provide (optional) feedback when requesting edits
* [ ] Test flow through message states

IanMayo:
Done, in #1012

PR #986, opened 2021-08-02
Feature/985 early collab working

IanMayo:
Supports #985 

See task list in #985


IanMayo:
Close - Lilit did it her own way 🤣

PR #987, opened 2021-08-06
Wargame turn presentation change

### LINKED ISSUE #955: As Game Manager I want to control how turns are presented
Past games have used a `natural` turn numbering system: `0, 1, 2, 3, 4` (where turn-zero is spent preparing to play).

For the next wargame, unusually the turns will be run in pairs, so it will be `0.1, 0.2, 1.1, 1.2, 2.1, 2.2`.

In the `Overview` tab of `Admin` pages, we should introduce an optional `Turn Presentation` drop-down list, with this values/labels:
* `natural`: `Natural (0, 1, 2, 3)`
* `turn-pair numbers`: `Turn-pair numbers (0.1, 0.2, 1.1, 1.2)`
* `turn-pair letters`: `Turn-pair letters (0.a, 0.b, 1.a, 1.b)`

The data will be stored in an optional `turnPresentation` element that takes one of the two above values.   I guess they should be enums in TS, so we can grow them in a controlled way in the future.

I don't think we will have to change the actual turn numbering, we just cosmetically change what is displayed if there is a value in `turnPresentation` that is not `natural`.

This logic will convert `i` from natural to turn-pair numbers: 
```
Math.floor(i / 2) + '.' + (i % 2 ? 1 : 0)
```

This logic will convert `i` from natural to turn-pair letters: 
```
Math.floor(i / 2) + '.' + (i % 2 ? b : a )
```

The number rendering will happen in these places:
* Status bar at top-right
* In-channel turn markers
* Wargame Export (check for non-'natural' value in `overview.turnPresentation`, and render turn number column if necessary).

**Note:** in this modification we should also allow the Game Manager to set the starting turn number. This functionality would also be used to set the _current_ turn number part-way through a game.  The turn number will use the `natural` numbering above.

sanjayradadiya:
## 🧰 Issue

Fixes #955 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/client/src/consts.js
```diff
@@ -48,6 +48,8 @@ export const PLAN_REJECTED = 'rejected'
 export const CHANNEL_RFI_STATUS = 'rfis'
 export const CHANNEL_MAPPING = 'mapping'
 
+export const NATURAL = 'Natural'
```
IanMayo, line 51: @sanjayradadiya - I don't like us declaring these constant values twice.  Is it possible to just use the constants declared in `@serge/config` below?

packages/custom-types/wargame-overview.d.ts
```diff
@@ -5,6 +5,8 @@ export default interface WargameOverview {
   name: string,
   /** high level description of wargame */
   gameDescription: string,
+  /** type of turn presentation */
+  turnPresentation?: string,
```
IanMayo, line 9: @sanjayradadiya - shouldn't we specify that this can take one of the two constants?

IanMayo:
Close. Superceded by #990

PR #1020, opened 2021-08-27
Duplicate passcode in wargame forces

### LINKED ISSUE #1018: Duplicate passcode in wargame forces
## 🐞 Bug description

1. Duplicate password allowed in multiple force in wargame
2. Duplicate password allowed in same force in wargame
image

## 💻 System information

 - Operating System: [e.g. iOS, Windows, etc...]
 - Browser Version [e.g. 22]
 - Browser resolution 
 - Are cookies enabled
 - Is Javascript enabled
 - Do you have an ad blocker

sanjayradadiya:
## 🧰 Issue

Fixes #1018 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Thanks for this @sanjayradadiya  - but can't we do the check earlier?

When we're in the `forces` tab we have the force data object.  So, when the user clicks on `Add role` - we could just repeat:
1. create new role id
2. check if role id already in use
3. if in use, go back to `1`

I don't think the above would require a database round-trip.  Obvs we'd be in trouble if multiple admins were creating roles, but we don't expect that to happen.

sanjayradadiya:
> Thanks for this @sanjayradadiya - but can't we do the check earlier?
> 
> When we're in the `forces` tab we have the force data object. So, when the user clicks on `Add role` - we could just repeat:
> 
> 1. create new role id
> 2. check if role id already in use
> 3. if in use, go back to `1`
> 
> I don't think the above would require a database round-trip. Obvs we'd be in trouble if multiple admins were creating roles, but we don't expect that to happen.

Yes, we can do. 
It will not prevent the below scenario.
1. Duplicate passwords exist and users try to save that force.

IanMayo:
> Duplicate passwords exist and users try to save that force.

I think the above logic will prevent duplicate passwords from existing - so we won't have problem saving.

IanMayo:
Close & reopen, to wake up static analysis

packages/client/src/Views/GameSetupNew.jsx
```diff
@@ -122,6 +124,11 @@ const AdminGameSetup = () => {
     const forceName = newForceData.name
     newForceData.overview = forceOverview === 'string' ? forceOverview : forces.forces.find((force) => force.uniqid === selectedForceId).overview
 
+    if (isUniquePasscode(newForceData, forces.forces)) {
+      dispatch(addNotification('Duplicate password in force roles.', 'warning'))
```
IanMayo: thanks @sanjayradadiya - sorry if it makes the logic more untidy, but we really need to tell the Game Designer which two roles share a password.  So, I guess `isUniquePasscode` needs to return an array of role names that share the pass-code.

Oh, and the message has to be `Duplicate passcode in force roles` - not `Duplicate password in force roles`.

Thanks :-)

sanjayradadiya: Message rename to `Duplicate passcode in force role`

Game Designer will get message `Duplicate passcode in force role CO,Game Control` for sharing same password.

IanMayo:
I know this is dragging on, but I'm afraid we need a bit more @sanjayradadiya 

1. For create new role, I think we should use `isUniquePasscode` to ensure we're generating a unique code, rather than reporting the duplication to the Game Designer.  For this, we can't use the current `isUniquePasscode`.  If (by accident) there are duplicate pass-codes already in the game, `isUniquePasscode` will keep returning false, even though the newly generated one is actually unique.
2.  When report on the duplicate from the `Save Force` in admin pages, we actually need to provide the `force` and `role` for the duplicate entries, since the duplicate could be in another force.  Note: for this I _think_ we just need to put `force` and `role` into the `dupRoleNames` array
3. In `isUniquePasscode` we only check for duplicates in other forces.  I _think_ we should be checking for a duplicate in the current force, too (so we don't need the `newData.uniqid !== force.uniqid &&` test)

sanjayradadiya:
> I know this is dragging on, but I'm afraid we need a bit more @sanjayradadiya
> 
> 1. For create new role, I think we should use `isUniquePasscode` to ensure we're generating a unique code, rather than reporting the duplication to the Game Designer.  For this, we can't use the current `isUniquePasscode`.  If (by accident) there are duplicate pass-codes already in the game, `isUniquePasscode` will keep returning false, even though the newly generated one is actually unique.
> 2. When report on the duplicate from the `Save Force` in admin pages, we actually need to provide the `force` and `role` for the duplicate entries, since the duplicate could be in another force.  Note: for this I _think_ we just need to put `force` and `role` into the `dupRoleNames` array
> 3. In `isUniquePasscode` we only check for duplicates in other forces.  I _think_ we should be checking for a duplicate in the current force, too (so we don't need the `newData.uniqid !== force.uniqid &&` test)

1. `getUniquePasscode` used to generate password everywhere. force roles has unique password also check at time of Save Force to avoid accidental duplicate passcode.
2. New message set. `Duplicate passcodes for: White-CO, White-Game Control, Red-CO, Red-Game Control`.
3. duplicate passcode check in current force as well as other forces to avoid duplicate passcode which is stored in database.

sanjayradadiya:
1. test case for duplicate passcode in current force and other force also.
2. compare passcode for expected value from test case.
3. test case for legacy wargame duplicate passcode.

packages/helpers/src/tests/get-unique-passcode.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { forces } from '@serge/mocks'
+import { ForceData } from '@serge/custom-types'
+import { getUniquePasscode } from '../'
+
```
IanMayo, line 4: @sanjayradadiya  - you can probably delete this test. As you said, it's really hard to test for the creation of a unique passcode.

Ah no, let's keep it.  At least it's proving `getUniquePasscode` still works, without any compiler errors.

Actually, I suggest adding: `expect(res).toBeTruthy()`   ... just to check _something_ is returned.

PR #1032, opened 2021-08-31
1028: Fix Play and Enter buttons not visible

### LINKED ISSUE #1028: Play and Enter buttons not visible
A change elsewhere in Serge has caused the `Play` and `Enter` buttons to lose their background.

image

I don't think we can revert the change since that will probably break something somewhere else.

So, I _guess_ we have to give these buttons an extra class, and to style that class as necessary.

Note: this is also an issue in @lilitkarapetyan 's `Create document` PR:
image

Maybe we need to investigate the cause, and apply a different fix to the change that caused this issue.  That's because there may be other _broken_ places we haven't found yet.  So, it's probably better to find the cause, revert the change, then make a specific change for thing originally being fixed.

Thanks for spotting it @lebaphi

lebaphi:
Fixes #1028

### DIFF
--- a/packages/themes/_flex-layout.scss
+++ b/packages/themes/_flex-layout.scss
@@ -44,6 +44,7 @@
       width: 100%;
 
       .btn-action--primary {
+        color: $white;
         margin-top: 16px;
       }
 


IanMayo (approved):
Great, thanks.

PR #1044, opened 2021-09-01
New strategy for selecting countries.

### LINKED ISSUE #1040: As a player I need to be able to provide multiple locations for my Course of Action
We currently follow `standard-practice` and offer region and country as dependent drop-downs.

I've just found out the players may need to specify multiple countries.

So, we need to change the implementation to allow for multiple countries in regions to be added.

First array is "Europe", containing the two sample countries we're using.

Second array is "Asia", and so on.

IanMayo:
Fixes #1040


IanMayo:
Close. Ian (the idiot) fixed it in `develop`.

PR #1067, opened 2021-09-03
trial stack upgrade

IanMayo:
Trial upgrading the Heroku stack from 18 to 20, to see if it overcomes the intermittent build timeout issue


IanMayo:
It didn't overcome the intermittent build issue. We overcame that by reducing the number of storybook tests.

PR #1070, opened 2021-09-04
Country not shown for some regions

### LINKED ISSUE #1036: Country not shown for some regions
Note: this issue can be reproduced by entering the `Papa Foxtrot` game as `Blue Senior`, going into the `Blue COAs` channel, and creating a new COA document.

We're using JSON-Schema `dependencies` to make one drop-down control dependent upon the value of another.

But, this is breaking when we edit an existing document:
screencast

It works fine for new documents:
screencast 2

Aah, note how the control moves to the right for some values.  That _may_ be down to the length of the USA country name. Will experiment.

Note: resizing the browser window makes the missing control appear:
screencast 3

lebaphi:
Fixes #1036


IanMayo:
Could you explain how the fix works please @lebaphi ?

It's not _good engineering_ for `json-editor` to know about the structure of both the schema _and_ the message.

I wonder if there are alternative solutions?

lebaphi:
> Could you explain how the fix works please @lebaphi ?
> 
> It's not _good engineering_ for `json-editor` to know about the structure of both the schema _and_ the message.
> 
> I wonder if there are alternative solutions?

yep, I found when I change the drop down, it will trigger this code line:
```
handleChange(nextEditor.getValue())
```
but the `nextEditor.getValue()` only return the current selected value on the -> update the `message.LOCATION` schema -> re-render -> lack of data of the schema -> the country dropdown has gone.

let say with example for you to get easy understand: 

the LOCATION schema:
```
LOCATION: {
          title: "*Location of activities in this turn",
          type: "object",
          properties: {
            region: {
              title: "Region",
              type: "string",
              enum: [
                "Asia Pacific",
                "Europe",
                "Americas"
              ],
              default: "Europe"
            },
            Europe: {
              title: "Country",
              type: "string",
              enum: [
                "United Kingdom",
                "Germany"
              ],
              options: {
                dependencies: {
                  region: "Europe"
                }
              }
            },
            Americas: {
              title: "Country",
              type: "string",
              enum: [
                "United State of America",
                "Brazil"
              ],
              options: {
                dependencies: {
                  region: "Americas"
                }
              }
            },
            AsiaPacific: {
              title: "Country",
              type: "string",
              enum: [
                "India",
                "China"
              ],
              options: {
                dependencies: {
                  region: "Asia Pacific"
                }
              }
            }
          }
        },
```

the schema of the `message.LOCATION` in edit mode should be:
```
region: <selected_value>,
Europe: <selected_value>,
Americas: <selected_value>,
AsiaPacific: <selected_value>
```
but currently it is (because it is selecting on the form)
```
region: <selected_value>,
Europe: <selected_value>,
```
That is why it does not render other countries dropdown. 

This fix is only for the `LOCATION` , I think we should have a general solution for this. Maybe when creating a new instance of the Editor, we should verify that all schema's KEYS are available 🤔

IanMayo:
Closing this issue. We no longer rely on dependent arrays.

PR #1092, opened 2021-09-07
Show full content of RFI answer

### LINKED ISSUE #1079: Show full content of RFI answer
When an RFI answer is shown, it's in a text-area of fixed height, showing a scroll-bar when necessary.

image

Since the answer is such an important part of the RFI, we should show the whole RFI.

lebaphi:
Fixes #1079


lebaphi:
maybe we should update the TextArea component by adding extra prop like: `fitContent` or `dynamicHeight`, once it set it to true, the textarea height should be fit the content.

lebaphi:
Usage: `<Textarea fitContent={true} ....`

image

IanMayo (approved):
Great, thanks.

PR #1120, opened 2021-09-10
1116: Allow data table to be sorted by Reference or Updated columns

### LINKED ISSUE #1116: Allow data table to be sorted by Reference or Updated columns
Some users would like the COA table to be sorted by document ID (so they're in persistent locations), and others would like them to be sorted in updated order.

We could support this with `Sort` icons on the relevant columns in the header.  The `updated` column currently shows the value produced by "moments" - which is difficult to sort.  Instead we should store the age (prob in millis) but convert that to `moment` on rendering.

I _guess_ we need to pass additional props to the data-table, maybe with a boolean on `sort`.

Obvs we need to re-run the Active Sort on table updates.

lebaphi:
Fixes #1116

### DIFF
--- a/packages/components/src/local/atoms/table-head-cell/index.tsx
+++ b/packages/components/src/local/atoms/table-head-cell/index.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState, ReactElement, useMemo } from 'react'
+import React, { useRef, useState, ReactElement } from 'react'
 
 /* Import Types */
 import Props, { ContentFilterType } from './types/props'
@@ -62,9 +62,8 @@ const FilterIcon = (): React.ReactElement => (
 )
 export const TableHeadCell = (props: Props): (React.ReactElement | null) => {
   const [open, setOpen] = useState<boolean>(false)
-  const [sortDirection, setSortDirection] = useState<SortDirection>('asc')
   const buttonRef = useRef(null)
-  const { content, id, filters, onFilter, sort, onSort, sortingColId } = props
+  const { content, id, filters, onFilter, sort, onSort, sortingColId, sortDirection = 'asc' } = props
   const contentFilter = content as ContentFilterType
   const classes = useStyles()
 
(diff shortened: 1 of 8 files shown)


IanMayo:
Nice work, @lebaphi 

Could we have an indication in the header for which sort is in place please?  Maybe color in the relevant arrow?

IanMayo:
Aah, we lose the sort order on new (or updated) message. I guess we need to re-run the current sort on each change.

lebaphi:
> Aah, we lose the sort order on new (or updated) message. I guess we need to re-run the current sort on each change.

yep, Im working on it.

IanMayo:
@lebaphi - note the conflict on the branch.  We've changed how we generate the `key` for the `data-table`.  The `random()` key was causing the table to re-generate on every new message. We now use the `message._id` and the flicker has gone :-)

This caused a few changes in the file.  You may find it easier to re-apply your changes to an updated version of `data-table\index.tsx`.

lebaphi:
> @lebaphi - note the conflict on the branch. We've changed how we generate the `key` for the `data-table`. The `random()` key was causing the table to re-generate on every new message. We now use the `message._id` and the flicker has gone :-)
> 
> This caused a few changes in the file. You may find it easier to re-apply your changes to an updated version of `data-table\index.tsx`.

thank you @Ian, I will resolve it

IanMayo:
Nearly there :-)

Note below: the items are in ascending order of `ID`.

From another window, I create a new message.

When the table updates, the items appear to be in ascending `updated` order.
Screencast

IanMayo (changes requested):
Fixes for sorting on receipt of new message, please.

lebaphi:
> Fixes for sorting on receipt of new message, please.

yep, I'm working on this

IanMayo:
> > Fixes for sorting on receipt of new message, please.
> 
> yep, I'm working on this

No rush @lebaphi  - I'm about to finish for the weekend.  

Oh, and I won't be merging it until at least Monday, because I've excitedly sent an analyst a link to the Heroku deployment for this PR.  If I merge it, the link will break.  Duh, Ian!

IanMayo:
Thanks @lebaphi 
I'm sure the logic is complex, but we're still not quite there yet.

I sorted them so that the oldest is at the top, and the newest at the bottom.  Then I modified an item.   It should have moved to the bottom - but it stayed where it was.

screencast

lebaphi:
> Thanks @lebaphi
> I'm sure the logic is complex, but we're still not quite there yet.
> 
> I sorted them so that the oldest is at the top, and the newest at the bottom. Then I modified an item. It should have moved to the bottom - but it stayed where it was.
> 
> screencast

wow, let me check this @IanMayo

lebaphi:
Hi @IanMayo , in this case, when you edit one item, the data table will sort again but it occurs on the column ID as default. It should do based on the current sorting-column, I will update it 👍

lebaphi:
I pushed the fix @IanMayo , now it tracks sort direction and sorting-column when data changed.

IanMayo:
Working fine :-)
Will wait for tests to finish.

IanMayo (approved):
Working fine

PR #1133, opened 2021-09-14
Fix reference dublication on message creation

### LINKED ISSUE #1113: Blue-NaN reference number created
For some reason, we have Four COAs with the reference number `Blue-NaN`.

They mistakenly get treated as the same COA, but they should not have a `NaN` identifier.

### LINKED ISSUE #1115: Duplicate force reference numbers created
I guess there were about 40 COAs created in the game, though the management process meant the database held multiple versions of many of them.

_Maybe_ around 5 COAs (and maybe some RFIs) had duplicate IDs.  Because of that, the older RFI effectively got lost.

I think we need to examine the logic around creating a new COA/RFI, and ensure it occupies the shortest possible time before sending the new document to the database.   If this is already almost instantaneous then we need another approach.  _Maybe_ we'll need to create a new GUID for the document *And* the reference number. We'll manage documents according to the GUID, but display the reference number.  When processing the initial load of documents (and new documents) the logic would look for GUIDs sharing a reference number - and modify the reference numbers to attach some hash of the GUID.

lilitkarapetyan:
## 🧰 Issue

Fixes #1113
Fixes #1115

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo (approved):
Fix tested in two "bulk send" sessions.

PR #1134, opened 2021-09-14
Filter on Location

### LINKED ISSUE #1126: Collab working channels - filter on `From` and `Location`
Some users have expressed the wish to be able to filter on the `From` field.  This seems straightforward.

They have also asked to filter on `Location`.  This is more challenging, since it's a comma-separated array of values.

I _guess_ we could handle it by passing an array of strings to the data-table, and the data-table rendering them as comma-separated. But, it would generate the drop-down and perform the filter based upon the arrays of values.

So, the filter on "From" is essential. The filter on "Location" is nice-to-have, please.

image

lebaphi:
Fixes #1126

### DIFF
--- a/packages/components/src/local/coa-status-board/index.tsx
+++ b/packages/components/src/local/coa-status-board/index.tsx
@@ -77,8 +77,13 @@ const getListOfExtraColumn = (messages: MessageCustom[], columnName: string): st
       case 'LOCATION':
         fields = message.message[columnName].LOCATION
         for (const key of Object.keys(fields)) {
-          const location = fields[key].map((item: any) => `${key}-${item.Country}`)
-          fieldData.push(location)
+          if (!fields[key].length) {
+            filters.push(['[Empty]'])
+            return filters
+          } else {
+            const location = fields[key].map((item: any) => `${key}-${item.Country}`)
+            fieldData.push(location)
+          }
         }
         filters.push(fieldData.join(' '))
         return filters
--- a/packages/components/src/local/coa-status-board/index.tsx
+++ b/packages/components/src/local/coa-status-board/index.tsx
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from 'react'
 import { DataTable, ROW_WITH_COLLAPSIBLE_TYPE } from '../organisms/data-table'
 import { Badge } from '../atoms/badge'
 import { MessageCustom } from '@serge/custom-types/message'
-import { CollaborativeMessageStates, SpecialChannelColumns } from '@serge/config'
+import { CollaborativeMessageStates, EMPTY_CELL, SpecialChannelColumns } from '@serge/config'
 import DataTableProps, { Column, RowWithCollapsibleType } from '../organisms/data-table/types/props'
 
 /* Import Types */
@@ -79,9 +79,9 @@ const getListOfExtraColumn = (messages: MessageCustom[], columnName: string): st
           const location = fields[key].map((item: any) => `${key}-${item.Country}`)
           fieldData.push(...location)
         }
-        const newItems: string[] = fieldData.length ? fieldData : ['Empty']
+        const newItems: string[] = fieldData.length ? fieldData : [EMPTY_CELL]
         filters.push(...newItems)
-        return filters.sort((a, b) => a === 'Empty' ? -1 : a - b)
+        return filters.sort((a, b) => a === EMPTY_CELL ? -1 : a - b)
       }
       default:
         return filters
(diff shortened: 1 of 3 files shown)


IanMayo:
@lebaphi - I'm not a fan of the software knowing about the structure of the messages, including special behaviour for the `Location` column.

Isn't it an option to pass an array of values in the `data` props, and to handle the array of data?

lebaphi:
Hi @IanMayo, I see we already implemented the filter on `From`, so this PR just support us to filter on the `Location` column. I consider in case if location is empty like this (last row)
image
How we handle the filter? Do we have a empty item on the dropdown to filter those items? or just ignore it? 🤔

IanMayo:
> Hi @IanMayo, I see we already implemented the filter on From,

Good point.  _But_ - for a document that is being handled through collaborative working, we can accept the special case that there will be a `From` field - since the document must have come from someone.  But, for other fields I'd rather we had a more "reflection" like strategy where we work from the data.

> if Location is empty

Hmm, another good point :-) I guess that if we have a row containing arrays of data, but a row is empty, then we offer an [Empty] value in the drop-down, to match rows with zero elements.

Would that work?

lebaphi:
> Good point. _But_ - for a document that is being handled through collaborative working, we can accept the special case that there will be a `From` field - since the document must have come from someone. But, for other fields I'd rather we had a more "reflection" like strategy where we work from the data.

You mean we should have a common function to build the filter for all all columns instead of implementing each field? 
I think of it when add the filter for the location, so I add the new function like:
```
const getListOfExtraColumn = (messages: MessageCustom[], columnName: string)....
```
we input the `columnName` and get corresponding list of filter. We just implement the filter for some special field, and for common field, we can use the same logic to build the filter.

> Hmm, another good point :-) I guess that if we have a row containing arrays of data, but a row is empty, then we offer an [Empty] value in the drop-down, to match rows with zero elements.

Yes, it should work

IanMayo:
> You mean we should have a common function to build the filter for all all columns instead of implementing each field?
I think of it when add the filter for the location, so I add the new function like

Aah, I see we already have special processing for LOCATION channels:
image

But, I _think_ I would prefer this processing:
* in CoaStatusBoard, `data.map` collates the data to go into the table.
* the items in the list can be a component or a string.  Where the data in `extraCols` is an array, it is collapsed into a single strong.
* Modify the above to that the items in the list can be a component, a string, or an array
* If <DataTable> receives an array of data, then it will render the array as a comma-separated list, but it will register the individual items in the `filter` drop-down.

IanMayo:
Aah, it seems we're already passing in array of items for `Location`:
image

So, I _guess_ this just needs changing in `DataTable` - but `useMemo()` just needs to check if the object is an array - and if so, handle it as individual items.

lebaphi:
> Aah, it seems we're already passing in array of items for `Location`:
> image
> 
> So, I _guess_ this just needs changing in `DataTable` - but `useMemo()` just needs to check if the object is an array - and if so, handle it as individual items.

yep 👍

IanMayo:
I'm getting this in the drop-down:
image
I have data like this:
image

IanMayo:
Thanks @lebaphi 

But, it's not breaking them down into individual locations:
image

IanMayo:
Hello, @lebaphi - I've had a go at fixing the implementation. Could you have a look at my commit(s), and check the JS is ok?

Oh, and feel free to check it works, too :-)

lebaphi:
> Thanks @lebaphi 
> 
> 
> 
> But, it's not breaking them down into individual locations:
> 
> image
> 
> 

Thanks Ian, but I think the menu item should display the same as the location on each row?

IanMayo:
Here are 3 COAs, some of which have countries.
image

Here is the list of the unique countries present in the column, plus an `Empty` marker since the third COA doesn't have a location.
image

lebaphi:
> Here are 3 COAs, some of which have countries.
> 
> image
> 
> 
> 
> Here is the list of the unique countries present in the column, plus an `Empty` marker since the third COA doesn't have a location.
> 
> image
> 
> 

I got your point, the menu item should list all the unique countries of the location. I build the menu item by adding the location cell text (the same with other column) and filter base on this. okay, I will have a look on your fix and feedback soon 👍🏻

lebaphi:
Hi @IanMayo, your fix works fine 👍

IanMayo:
> Hi @IanMayo, your fix works fine 👍

And is the JS ok? You are welcome to refactor it for "modern JS" or performance reasons.

lebaphi:
> > Hi @IanMayo, your fix works fine 👍
> 
> And is the JS ok? You are welcome to refactor it for "modern JS" or performance reasons.

The code is fine I think :)

IanMayo (approved):

PR #1136, opened 2021-09-15
Game control assign to role

### LINKED ISSUE #1062: Need flag to assign a role as `Game Control` able to step forward
We have an `isGameControl` flag in the wargame to show which role can move the game forward.

We can't set this from the admin pages. It's `probably` another toggle on `roles`.

sanjayradadiya:
## 🧰 Issue

Fixes #1062 

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

image

## Confirmations


IanMayo (changes requested):
The logic is fine about `don't show delete if this is game-control` - but I suspect users won't understand why that is the case.

So, I'd rather you show the `delete` icon, but in the handler check if it's game control.  If it is, popup a message saying:
`User with Game Control permissions cannot be deleted. Please remove Game Control permission`.

packages/components/src/local/molecules/sortable-list/types/props.d.ts
```diff
@@ -2,6 +2,8 @@ import { ReactNode, ReactText } from 'react'
 
 export interface ItemObject {
   name: string
+  /** If this role is `Game Control` - a role which cannot be deleted */
+  isGameControl?: boolean
```
IanMayo: Ooh no, I don't like this @sanjayradadiya 

Sortable list is used in a range of places, including where `isGameControl` isn't present.

I know it's an optional prop - but I'd rather not see it here.

I don't know what the alternatives are, but I _guess_ `canDelete` callback could be provided that return true/false on if the row can be deleted.  It would popup the warning message if it receives a `gameControl` user, and return false.

Or, `maybe` this isn't needed, and we just access `isGameControl` from the item.

sanjayradadiya: I will recheck this and will find out way around to avoid this

sanjayradadiya: `isGameControl` removed from sortable list.

introduce new callback function for role delete with notification `Role EXCON with Game Control permissions cannot be deleted. Please remove Game Control permission.`

IanMayo (approved):

Issue #1053, opened 2021-09-02
Retire RFI Status Board


IanMayo:
The new collaborative working status board is better than the previous RFI one, offering more customisation.

We should retire the legacy working RFI Status Board (`RfiStatusBoardChannel`), and delete the components.

IanMayo:
Done.

Issue #983, opened 2021-07-30
Heroku build failing


IanMayo:
We did have an intermittent Heroku build issue, now it's happening regularly.

Build log attached: 
[build_log.txt](https://github.com/serge-web/serge/files/6906285/build_log.txt)

I _wonder_ if it's related to the size of Serge.  I don't think we've introduced any new technologies - but the size is one thing that has changed - as it steadily grows.

IanMayo:
This build issue has come back.  It was intermittent, now it's failing most builds.  Since Serge is steadily growing, it _may_ be related to the size of Serge.  Maybe try removing something (legacy RFI Status Board?) - and seeing if that fixes it.

IanMayo:
I contacted Heroku support, and got this feedback:
![image](https://user-images.githubusercontent.com/1108513/132480565-47f34678-2e29-4fab-a466-31aac57b6a57.png)

IanMayo:
Builds working reliably.

PR #1139, opened 2021-09-16
1102: Reduce bundle size

### LINKED ISSUE #1102: Reduce bundle size
The webpack build process warns our distributable files are too large.

image

Most usage of Serge is on a LAN, but even on Heroku it seems to load quickly enough.  Well, at least I think it does - I've always thought the initial delay was the Heroku instance spinning up.

Can we tidily reduce the size of the packages by splitting them up?  At least (I guess) they'll be downloaded in parallel.

We could probably also split up the distribution.  We don't need to deliver `admin` _and_ `player` to the client browser.  In lots of cases the huge `mapping` components aren't required.

With the above exercise, we could also ditch some unused code/components.  `RLI Status Board` is certainly one instance of this, but there are probably old `client` modules, too.

lebaphi:
Fixes #1102 

before:
before

after:
after


lebaphi:
I updated some config in webpack, try to split the chunk file to smaller part if its size is too big (then browser) can load in parallel. Also make some improve in build process (build in parallel, don't need to included the source map in production mode).
I take a look on removing the redundant component like `RLI Status Board` but I can't find where it is, did you mean the `RFI Status Board` @IanMayo ?

lebaphi:
The Heroku got an error on deploying, can you help me to check the error @IanMayo ?

IanMayo:
> The Heroku got an error on deploying, can you help me to check the error @IanMayo ?

Hello @lebaphi, it looks like it timed out while compiling StoryBook:
image

IanMayo:
> I take a look on removing the redundant component like `RLI Status Board` but I can't find where it is, did you mean the `RFI Status Board` @IanMayo ?

Yes, this issue: https://github.com/serge-web/serge/issues/1053

lebaphi:
> > The Heroku got an error on deploying, can you help me to check the error @IanMayo ?
> 
> Hello @lebaphi, it looks like it timed out while compiling StoryBook:
> image

hmm, I tried to run the command `yarn build-storybook` on my machine, it only takes 3m30s - 3m45s. But sometimes it takes ~7min on the Heroku and the build failed. So I think we have to do some build improvement to speed it up.

IanMayo:
> But sometimes it takes ~7min on the Heroku and the build failed. So I think we have to do some build improvement to speed it up.

Yes, I agree.  I think your comment mentioned running things in parallel - though maybe StoryBook can't do that.  Heroku commented on the performance issue [here](https://github.com/serge-web/serge/issues/983)

lebaphi:
> > But sometimes it takes ~7min on the Heroku and the build failed. So I think we have to do some build improvement to speed it up.
> 
> Yes, I agree. I think your comment mentioned running things in parallel - though maybe StoryBook can't do that. Heroku commented on the performance issue [here](https://github.com/serge-web/serge/issues/983)

I think we can reduce time on other packages for the time of storybook :D. I found something at https://dev.to/rivertwilight/use-esbuild-to-speed-up-your-creat-react-app-project-2375 but not sure if we get it work, at least we can reduce ~20sec on the packages client 🤔

IanMayo:
I guess that _Temporarily_ we could remove a large block of storybook tests, at least until we find a better solution.

The storybook tests only tend to be used for new feature.

We'd do it by renaming them. Shall we try it?

lebaphi:
> I guess that _Temporarily_ we could remove a large block of storybook tests, at least until we find a better solution.
> 
> The storybook tests only tend to be used for new feature.
> 
> We'd do it by renaming them. Shall we try it?

I think we should not delete tests because it is not only used for new feature but also when we fix bug or make change somewhere we also need those tests to verify it's not broken others.

IanMayo:
Oh @lebaphi - I'm not proposing deleting all tests - just the storybook stories. I don't think the StoryBook tests give us any regression testing - they're only usable by humans.  The unit tests and jest snapshots are what helps us ensure old code doesn't get broken by new code.

lebaphi:
> Oh @lebaphi - I'm not proposing deleting all tests - just the storybook stories. I don't think the StoryBook tests give us any regression testing - they're only usable by humans. The unit tests and jest snapshots are what helps us ensure old code doesn't get broken by new code.

Hi @IanMayo, so if we are going to delete all storybook stories, I think we can modify the `.storybook/main.js` like:
```
module.exports = {
stories: [],
.....
```
I tried with `yarn build-storybook`, it takes ~20sec instead of ~250s before, but `yarn storybook` will show nothing.

IanMayo:
I was hoping for a way to just include some stories - such as the one being edited. I guess we could modify `main.js` to include:
 ```
      '../src/**/*.stories-inc.mdx',
      '../src/**/*.stories-inc.@(tsx)'
```

That will cancel all stories, but we can use that format for a new component in development.

lebaphi:
Hi @IanMayo, I pushed the change and it includes some Webpack config improvements.

IanMayo:
@lebaphi - when you get chance, could you have another look at the changes in this PR, and verify they are still correct?

lebaphi:
> @lebaphi - when you get chance, could you have another look at the changes in this PR, and verify they are still correct?

okay 👍🏻

lebaphi:
Hi @IanMayo, I checked and it works well.

IanMayo:
Hello @lebaphi 

This is another PR that we should probably get merged.  Would you mind having a look at the changes, and verifying they are still valid please?

Promise I'll merge it this time ;-)

IanMayo:
Reconsider this once we've moved to monolith #1720

IanMayo:
Overtaken by our dropping mapping modules.

PR #1143, opened 2021-09-16
always render collab channels

### LINKED ISSUE #1110: Contents of new document lost when switching to another tab
A Blue player started adding a new COA.  Then he switched to Blue RFIs to check one of his RFI answers.

When he switched back to the COA channel, his new COA had gone and the `New message` panel was closed.

It is our intention that non-visible channels are not rendered - as a way of increasing performance.

But, to keep the open COA while the RFI is being edited, we need to keep the rendering of _underneath_ channels.  Or, we remember the state of COA (message contents, open/closed status, selected field) - and recreate the form as it was.

The second channel seems more fragile, potentially forgetting text that has just been typed, or the cursor position within a block of text.

Aah, maybe we do keep _Special channels_ , but don't render hidden other channels.

lilitkarapetyan:
## 🧰 Issue

Fixes #1110 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations

### DIFF
--- a/packages/client/src/Components/Channel.tsx
+++ b/packages/client/src/Components/Channel.tsx
@@ -65,7 +65,7 @@ const Channel: React.FC<{ channelId: string }> = ({ channelId }) => {
         {isCollabWorking ? (
           <CoaStatusBoard
             templates={state.allTemplatesByKey}
-            messages={messages}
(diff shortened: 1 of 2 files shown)


IanMayo (approved):
Works perfectly, thanks.

PR #1149, opened 2021-09-17
1145: Bug when showing RFIs

### LINKED ISSUE #1145: Bug when showing RFIs
I'm pretty sure this is happening in the RFI channel. RFIs do not have location.

Warning: `message have no field LOCATION`. This is being fired from line 70 of COA Status Board.  The message is then shown.  We can see it's an RFI (`messageType: "RFI"`), and that there isn't a location field.

We only need to look for LOCATION fields if that field is present in the message.  If it isn't present then we don't need to throw a warning, IMHO.

Note: have commented out warning line in #1147, but we could prob do with tidier solution.

lebaphi:
Fixes #1145


lebaphi:
Hi @IanMayo, if the message does not included the `LOCATION` field or any extra field, we do not show the warn message.

IanMayo:
Aah, I didn't see the duplicate filter.

Thanks for spotting it.

IanMayo (dismissed):
Great, thanks.

IanMayo:
@lebaphi - I've found another issue related to what we're looking at here.

You removed the duplicate item from the `Status` filter. But, I saw we're missing some other states.  Plus, there are some states that apply to COAs and not RFIs (and vice versa).

So, I think we need to adopt a different approach.  I think we need to generate the list of states based on the data present in the column (like we do for `from`, `location`, etc.

Could you make that enhancement please?

lebaphi:
> @lebaphi - I've found another issue related to what we're looking at here.
> 
> You removed the duplicate item from the `Status` filter. But, I saw we're missing some other states. Plus, there are some states that apply to COAs and not RFIs (and vice versa).
> 
> So, I think we need to adopt a different approach. I think we need to generate the list of states based on the data present in the column (like we do for `from`, `location`, etc.
> 
> Could you make that enhancement please?

okay @IanMayo 👍

lebaphi:
done @IanMayo

IanMayo (approved):
Good solution, thanks.

PR #1171, opened 2021-09-29
Reduce Game Admin Panel

### LINKED ISSUE #1168: Serge needs to work on smaller monitors - reduce Game Admin
A set of players are accessing Serge in a meeting room via a projector. They're doing this so that they can discuss their answers.

But, the projector is old & clunky.

Could we please reduce the width of the Right-Hand "Game Admin" channel by 25% please?

There may be a better long-term fix, but this will improve things for this week.

lebaphi:
Fixes #1168


IanMayo:
Thanks - you've done exactly what I asked for.
image

What I should have asked for was to shrink it to `250px`.  I'll make the change by hand and test.

lebaphi:
> Thanks - you've done exactly what I asked for.
> 
> image
> 
> 
> 
> What I should have asked for was to shrink it to `250px`.  I'll make the change by hand and test.

yep 250px is fine but it likes a fixed value, I preferred a dynamic value 😂

IanMayo (approved):

PR #1174, opened 2021-09-29
Show warning message on cancel

### LINKED ISSUE #1170: Warning message on use of `Cancel`.
A user believes that quite a long COA document was lost when he pressed `Submit`.  That was frustrating.

A colleague suspects he actually pressed `Cancel`. 

We should popup a dialog check the user is sure they wish to `Cancel` the document.

This is specific to the `COAStatusBoard`, really.

*Note:* this will probably go into the NewMessage logic. But, we should add a prop to indicate whether to show the warning on Cancel --since there are lots of times when the form is simple and we don't need to check.

lebaphi:
Fixes #1170


IanMayo:
@lebaphi -- we only need to show the dialog in one situation. So, we need to use a prop to control if the dialog is shown.

> Note: this will probably go into the NewMessage logic. But, we should add a prop to indicate whether to show the warning on Cancel --since there are lots of times when the form is simple and we don't need to check.

lebaphi:
> @lebaphi -- we only need to show the dialog in one situation. So, we need to use a prop to control if the dialog is shown.
> 
> > Note: this will probably go into the NewMessage logic. But, we should add a prop to indicate whether to show the warning on Cancel --since there are lots of times when the form is simple and we don't need to check.

Yep @IanMayo, I create a new component `Confirm` to handle dialog, input the `isOpen` prop to control if the dialog is shown. Another changes is to make the code clean.

IanMayo:
> Another changes is to make the code clean.

I'll leave you to tidy that code @lebaphi

IanMayo (approved):
Great, thanks.

PR #1190, opened 2021-10-20
Remove animation from expand message

### LINKED ISSUE #1187: Remove animation from expand message
We're experiencing some performance issues with collab-edit channels with lots of messages.

This is on virtual machines that may have little processing power.

It appears that we animate the expansion of JSON-editor forms in the collab-editing channels.  We should do a test build that doesn't animate the expansion of a COA - and test it on target networks.

lebaphi:
Fixes #1187


IanMayo:
Works great. But, will test on target VMs before I merge it into `develop`.

PR #1192, opened 2021-10-21
1101 - Toggle to display archived documents

### LINKED ISSUE #1191: Include toggle to display "Archived" documents
In the current wargame users can have hundreds of COAs & RFIs.

This is causing a performance problem on some poorly specced virtual machines.  Improving the spec of the VMs is outside our control.

But, we could archive old COAs & RFIs, and by-default, not display them.

"Archived" documents will have a `details.archived` value of `true`.

So in the `colab-edit` channel, we should provide a `Display archived` checkbox next to the `Mark all as read` button.  The default value would be false.

In the collab-edit processing, we would not display `archived` documents, unless the above checkbox was ticked.

Note: I'll have to modify the data via the backend to create the `archived` flags.  But, I guess we can develop/test the above using the COA/RFI storybooks.

IanMayo:
This is the start of work on #1191 

Fixes #1191


IanMayo:
Drop this. Work done in separate PR

PR #1195, opened 2021-10-27
Change pouch db to ravendb

lilitkarapetyan:
## 🚀 Overview: 

Link for ignore conflict files in heroku 

https://docs.codacy.com/repositories-configure/ignoring-files/

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/config/src/globals.ts
```diff
@@ -108,3 +108,14 @@ export const serverPath = (
...
+
+// PouchDb paths
+export const allDbs = 'allDbs'
+
+// RavenDb paths
+export const rdbPrefix = 'ravendb/'
+export const socketPath = 'http://localhost:4000'
+export const replicate = 'replicate/'
+export const deletePath = 'delete/'
+export const dbSuffix = '.sqlite'
```
IanMayo: `.sqlite` is the suffix for SQLite databases, I guess we don't create a physical file for `RavenDB` databases.

packages/client/src/api/db/index.tsx
```diff
@@ -0,0 +1,253 @@
...
+}
+interface RavenFetchDataArray {
+  msg: string,
+  data: (Wargame | Message)[]
+}
+
+type initPDB = (n: string) => PouchDB.Database<Message | Wargame>
+const initPouchDbDefault: initPDB = (databasePath: string) => new PouchDB(databasePath)
+
+export class DbProvider implements DbProviderInterface {
```
IanMayo, line 17: @lilitkarapetyan - I _think_ that a tidier version of this would be to create two instances of `DbProviderInterface`.  One implements PouchDb, and the other implements RavenDb.

They would be created in separate files, and each one will only know about its own database API.

This `index.tsx` file would create an instance of the specified DbProviderInterface.

packages/client/src/api/db/index.tsx
```diff
@@ -0,0 +1,253 @@
...
+import { Message, Wargame } from '@serge/custom-types'
+import PouchDB from 'pouchdb'
+import { ProviderTypeType, ProviderRavenDB, ProviderPouchDB } from './types'
+import { io } from "socket.io-client"
+
+export const POUCH_DB = 'pouchdb'
+export const RAVEN_DB = 'ravendb'
+const defaultProvider = RAVEN_DB
+
+export const availableProviders: ProviderTypeType[] = [POUCH_DB, RAVEN_DB]
```
IanMayo: `availableProviders` isn't used.

packages/client/src/api/db/types.d.ts
```diff
@@ -0,0 +1,50 @@
+import { Message, Wargame } from "@serge/custom-types"
+import PouchDB from 'pouchdb'
+import  RavenDB  from 'ravendb'
+import { Socket } from "socket.io-client"
+import { POUCH_DB, RAVEN_DB } from "."
+
+export interface ProviderPouchDB {
```
IanMayo: Hmm, if we do end up with two separate instances of `DbProviderInterface` it would be _neat_ to put these types into the two separate instances.

Here is my _ideal goal_.  `PouchDb` will only be mentioned in the Pouchdb isntance of DbProviderInterface and `RavenDb` will be mentioned in the `RavenDb` instanceof DbProviderInterface.

IanMayo:
Note: I have connected to remote hosted RavenDb using this code:
image

I can provide the certificate

IanMayo:
There appears to be an issue related to the server round-trip time.  I make a wargame change, and press `Save Overview`. If I then make a change (such as adding a user), the change gets lost when the page updates on new document received.

screencast

IanMayo:
One other thing.  I _Think_ it's worth introducing a `messages` collection. It will keep the data (slightly) easier to navigate in the management app.  We currently have one collection per document:
image

packages/config/src/globals.ts
lilitkarapetyan: In case if we will compleatly remove pouch db we dont need it, currently you can switch databases 
https://github.com/serge-web/serge/pull/1195/files#diff-66c5602cae11b6d25bdcb24fc3ff243a6a94346e857b2e5cd7211fc1101189c3R21

IanMayo: Sure @lilitkarapetyan - but this constant is under the `RavenDb` section - it should be under the `PouchDb` section.

IanMayo:
@lilitkarapetyan - there are some outstanding comments, including refactoring into two separate database instances.

_If_ we end up having to try _another_ backend technology - then having each instance in its own class will keep things easier to maintain, IMHO.

IanMayo:
@lilitkarapetyan - separating the two database implementation almost how I expected.  But, I think we still don't need the `db\index.tsx` module.  I _think_ that in `server.js` we could check the value of `DB_TYPE` and either create an instance of PouchDb or RavenDB - stored into an object of type `DbProviderInterface `.  (Note: that approach may be more `Java`-like than `Javascript`-like)

IanMayo:
@lilitkarapetyan - could we tell the user if the backend database isn't found?  Now that we're not building it into `nodejs` it's more likely that there could be a problem.  So, we should tell the user about it.

I _guess_ we could handle this via a heartbeat response, but I'm pretty sure we do database calls at the start of the app - when we may be able to tell the user about the issue.

IanMayo:
@lilitkarapetyan - I think we have some unresolved issues for this.

1. did you mean to add the empty `.env` file
2. where do we provide the Raven URL?  I presume this PR is connecting to an Internet-hosted instance
3.  most of our users will run Serge on a PC without an Internet connection.  Ideally, I'd like to distribute Serge with an embedded RavenDB instance that gets started alongside (well, just before) Serge

Could you look into the above too, please :-)

IanMayo:
@lilitkarapetyan - I think we're going to have trouble distributing RavenDB within the Serge release - partly because of the licensing issue.

It is _really convenient_ for people interested in Serge to be able to just download the app and run it.

So, I _currently_ think we should continue to support PouchDb by _default_, but allow a user to upgrade to using RavenDB if they wish to.  We'll need to provide a way of telling Serge which database to use.  Maybe accept command-line parameter?  Then provide two `.bat` files - one indicates PouchDb and the other indicates RavenDB.

What do you think?

lilitkarapetyan:
We can do it with command line, also in release version can be
game-puchdb-x64.exe
game-ravendb-x64.exe
game-pouchdb.sh
...

IanMayo:
Hello @lilitkarapetyan - I'm pretty sure we're going to need `RavenDB` for an Internet-based wargame next month.

We'll host the game in Heroku, but will need RavenDB to make the data persistent (rather than getting wiped after an hour of unuse).

I did just look at solving the conflicts, but struggled. I'd appreciate you doing that, please.

IanMayo:
Hello @lilitkarapetyan - I _think_ that we need SQLite to be the default database.  That's so the `heroku` deployment still works.  It also runs without the RavenDB instance needing to be `initialised` with a RavenDB login.

PR #1197, opened 2021-10-29
1180: As Game Control I wish all users to be using current version of Serge to ensure fixes present

### LINKED ISSUE #1180: As Game Control I wish all users to be using current version of Serve to ensure fixes present
For one wargame, users are playing from Virtual Machines.

The game lasts several months, and the VMs can stay alive between turns.

Yesterday I encountered a user playing Serve version dated 24th Sept. He experienced an issue that I know was fixed on the release dated 29th Sept.  I know an F5 refresh would update his version of Serge.

This is an enduring issue that may hurt us/them in the future.

One potential solution to this is for the server to reply to heartbeat with the current version. If this is newer than the client version, then a dialog can open to ask the user to refresh.

For this to happen we'd require the build version in both the server (node) code and the client code. We're currently not incrementing the version on updates - just the build date.

The build date comes from this client-side code:
image

There is a chance that use of that code could produce different times if separately declared in `client` package build vs `server` package build steps.  So, the code needs to go into a single central repository. Maybe in `config` package?

We do have a challenge here.  The issue this week was with an old copy of Serge.  If we add this feature, these old copies of Serge will not know how to handle the version (build) data coming back from the server.

So, I propose this modification:
- change the URL used for heartbeat
- updated server won't reply to old URL. Clients will show error. They will be advised to refresh
- server will return build version from new URL

Note: I don't necessarily think this is an opportunity to move forward the `last-interaction` logic at #910

lebaphi:
fixes #1180


packages/client/src/api/wargames_api/index.ts
```diff
@@ -157,22 +157,26 @@ export const listenForWargameChanges = (name: string, dispatch: PlayerUiDispatch
...
-    console.log(err)
-    return "NOT_OK"
-  })
-} 
+  return fetch(serverPath + 'health_check')
+    .then((response: Response): Promise<any> => response.json())
+    .then((data: any) => {
+      // check the version returned from the server with the current client version
+      if (!data.version || data.version !== process.env.REACT_APP_VERSION) {
+        return "NOT_OK"
```
IanMayo: @lebaphi  - we're probably going to have to check for `"NOT_OK"` in the client code too.  It occurs in line 171, too. So, I think we should create a const somewhere for it - to ensure a type doesn't break the app.

lebaphi: sure, I will update now @IanMayo

packages/config/src/globals.ts
```diff
@@ -108,3 +108,7 @@ export const serverPath = (
 export const databasePath = `${serverPath}db/`
 export const iconUploaderPath = `${serverPath}saveIcon`
 export const hiddenPrefix = '_#_'
+export const SERVER_PING_STATUS = {
+  SUCCEED: 'OK',
```
IanMayo, line 123: Hello @philb89 - it looks like we don't use `SERVER_PING_STATUS.OK`.  We should, shouldn't we?

lebaphi: yep, but I added for the future, may we need to use it :D

IanMayo:
Hello @lebaphi ,
I'm trying to test this.

I think it's not working as expected.  The server is returning the version defined in `.env\REACT_APP_VERSION` (a hard-coded value in `\server\.env` ).
But, the client is comparing that to `$npm_package_version` - which comes from `client\package.json`.  These can get out of sync.
I _think_ we need a solution that takes the version from the root-level `package.json`.

IanMayo:
Aah, I have another request too. Could you introduce `SERVER_PING_STATUS.WRONG_VERSION` enum?  When `wargames_api` returns this value to `version.jsx`, we can display a message indicating `Please refresh browser (F5) to update SERGE`.

IanMayo:
@lebaphi - we have some more thinking to do on this - to ensure we're doing a valid comparison.  Could you share your thoughts on it please?

> Hello @lebaphi , I'm trying to test this.
> 
> I think it's not working as expected. The server is returning the version defined in `.env\REACT_APP_VERSION` (a hard-coded value in `\server\.env` ). But, the client is comparing that to `$npm_package_version` - which comes from `client\package.json`. These can get out of sync. I _think_ we need a solution that takes the version from the root-level `package.json`.

IanMayo:
Thanks for the update @lilitkarapetyan - we're getting closer.

Couple of issues:
- This feature is to check the client is running the same version of Serge as the server (in case the software has been updated, but the page hasn't refreshed).  I can't find the code where the client compares version(s) & reports "Currently running old version. Please refresh browse page".  Hmm, can you help me find it @lebaphi ?
- The build version (currently 0.3.0) doesn't get incremented in our build process.  But, the build date does.  So, IMHO we need to either increment the patch number in the build process (e.g. to 0.3.1), or use the build date to compare versions of client & server.

IanMayo:
Oh dear @lebaphi , we've (well, I've) left this one too long without updating it.  There are some terrible conflicts.

I think we should re-apply the changes in this PR to a fresh branch off develop, in a new PR.

lebaphi:
Hi @IanMayo, there is a lot of code changed in develop branch, also I don't see `healthcheck` endpoint, do we need this feature?

IanMayo:
Close. Revisit this once we have monolith.  Actually, having a monolith should make this go away.

PR #1205, opened 2021-11-20
Fix issue unread message count

### LINKED ISSUE #1204: Unread message count works incorrectly

lebaphi:
Fixes #1204


IanMayo:
Hello @lebaphi - could you have another look at this PR, and resolve the conflicts please?

lebaphi:
> Hello @lebaphi - could you have another look at this PR, and resolve the conflicts please?

sure @IanMayo, I resolve it now

IanMayo:
Aah, this isn't quite done.  When the player marks a message as `unread` we need to regenerate the count of unread messages.

Or, maybe that's an unrelated issue 🤣 

screencast

lebaphi:
> Aah, this isn't quite done. When the player marks a message as `unread` we need to regenerate the count of unread messages.
> 
> Or, maybe that's an unrelated issue 🤣
> 
> screencast

I think so, let me take a look at this :D

IanMayo (approved):
Working perfectly :-)

packages/components/src/local/coa-status-board/helpers/gen-data.tsx
```diff
@@ -209,7 +209,7 @@ export const genData = (
     }
 
     const row: Row = {
-      id: message.message.Reference || message._id,
+      id: message._id || message.message.Reference,
```
IanMayo, line 212: @lebaphi - do you know the reason for the above change?

All PouchDB documents have an `_id` field, so the `Reference` will never get used.  I _think_ the previous way was the correct way around.

lebaphi: @IanMayo Ah I get it from the old code base, but the same with you, I think we should use `message._id` instead of the `message.message.Reference`.
image

I change it to make the row id is the same with the message id, then we can filter the opened message in the `ExpandedComponent` component.

PR #1219, opened 2021-11-23
1198 collab data - DO NOT MERGE

IanMayo:
Prepare backend data for V3 of collaborative editing.

Note: compiler failing - so I'm creating this PR as a way of incrementally applying changes.


IanMayo:
Drop this branch.  I couldn't get it to work.  With hindsight, it appears the issue was because I was declaring an `enum` outside the `enums.ts` file.

PR #1224, opened 2021-11-28
1223: Update Location column for the alternative table

### LINKED ISSUE #1223: Update Location column for the alternative table

lebaphi:
Fixed #1223


IanMayo:
Ok. Two issues:
1. We need to show more of the locations, if we can.  You said in Slack that it breaks the layout. Could you provide a screenshot of how the un-trimmed list of locations breaks the layout please?
2. We need to provide filter on Location, Status, From, Owner.  Hopefully these capabilities will come from the `data-table` component.

lebaphi:
Hi @IanMayo , we can filter the location and display full location also.

location_filter
location

IanMayo:
Thanks @lebaphi - that layout of countries is good for me.

I see the location column get sorted, but how do I apply a filter? (to select rows for a particular country?)

Aah - I see the free text search.   Yes, that _may_ be a substitute for the other filters.  I'll try to get some feedback.

PR #1233, opened 2021-12-08
[1231] Private message in Edit RFI page not visible

### LINKED ISSUE #1231: `Private message` in  `Edit RFI` page not visible.
The `Private message` fields aren't legible, because of wrong colors.

Here is a link:
https://serge-review.herokuapp.com/?wargame=wargame-kskl1i3t&access=rkskl61c3

Look in room `Blue RFI 2`.  We're logged in as J2A, we have write access to `BLUE-9`.  When it expands, we can edit response.

We should also be able to edit the contents of the`Private Message` text-area.  It is actually there, and we can edit it, but it's not visible.

I expect this is just a styling change.

image

lebaphi:
Fixes #1231


IanMayo (approved):
Perfect, thanks :-)

Issue #1238, opened 2021-12-09
Consider use of Fonts in Serge UI


IanMayo:
We use a range of font-sizes within Serge. We have some text in all caps, and others in title case.

We should develop a consistent strategy for these elements, and document it in a wiki page here in the repo.

PR #1269, opened 2021-12-16
1215: Font Size and Type Uniformity

### LINKED ISSUE #1215: Font Size and Type Uniformity
Throughout many pages and windows of the game, the font size and type are somewhat inconsistent. It would be nice to have font uniformity through the different pages and windows.

Here's an example of different sizes of fonts in the same window:
Different Size-Type Fonts

lebaphi:
Fixes #1215


lebaphi:
Screen Shot 2021-12-16 at 8 46 29 PM

ShendMikullovci:
That's excellent @lebaphi. If you could apply the same font and color to "Show Archived" feature as well, please. Thank you!

lebaphi:
> That's excellent @lebaphi. If you could apply the same font and color to "Show Archived" feature as well, please. Thank you!

sure @ShendMikullovci 👍

lebaphi:
Please help to review it @ShendMikullovci :D
image

IanMayo:
There may be an issue with the use of case in the controls, but @ShendMikullovci is going to consider that separately, in this issue: https://github.com/serge-web/serge/issues/1238

ShendMikullovci:
The color looks great. As for the font size, I think sticking to 12px capitalized fonts would be good. I'd have to see it, however, considering how the "MARK ALL READ" came out, it looks promising. I'll provide a mean count shortly.

IanMayo:
I've just run this branch.

image

In my opinion it looks like the controls at the top of the channel are disabled.

Are you able to contribute any further @ShendMikullovci ?

ShendMikullovci:
> I've just run this branch.
> 
> image
> 
> In my opinion it looks like the controls at the top of the channel are disabled.
> 
> Are you able to contribute any further @ShendMikullovci ?

I think it appears to be disabled due to the lack of size uniformity. If we could get the font at the same size and only have initials capitalized, it should look smoother. If not, we can also go with a lighter gray.

IanMayo:
Yes, @ShendMikullovci - I agree that the controls are inconsistent.

i've just been googling "toggle or switch" - and it appears a "Switch" control would be more suited to "Show archived lessons".  I _think_ we also need to make "Mark all as read" look more like a button.   I propose we give "Mark all as read" a border.  I also propose (as Shend said) we make the case consistent for both.

Let's remove the `Filter` control.  We really need per-column drop-down filters and `order` controls.  @lebaphi is going to experiment with this in the `Player log table` feature.

Could you produce an updated UI please @ShendMikullovci ?

Note: I know @ShendMikullovci is trying to reduce the visual clutter, but I'm pretty sure we can't use grey for these controls, since it's the standard way of indicating something is disabled:
image

I _guess_ another way of reducing the clutter would be to use smaller/simpler controls.  _Maybe_ we can come up with icons for "Show archived" and "Mark all as read" - with tooltips to give further explanation.

ShendMikullovci:
I'm very much pro turning these features into buttons. I had a go at it: 

Inbox Icons

3 icons for all 3 features. The search "spyglass" icon is switched to a funnel icon since it more appropriately matches the term "filter". The archive button (middle) is the standard button for most messaging apps and is easily recognizable. Finally, the mark-read button we've used previously to maintain uniformity. 
In this screenshot, none of the features would be active.

In the screenshot below, all 3 features are active: 

Inbox Icons 2

Here, the user has selected to mark all messages as read (sorry that I couldn't do it for a better illustration, bug), has selected to show archived (thus archive button is filled) and has clicked on the filter button to type what they're searching for. 
Personally, I'd consider using terms like "search" or "sort" for the latter feature.

IanMayo:
@lebaphi  - could you also work on the move to icons that @ShendMikullovci suggested above please?

lebaphi:
> @lebaphi - could you also work on the move to icons that @ShendMikullovci suggested above please?

sure @IanMayo 😁

IanMayo (approved):

ShendMikullovci:
Hi guys! Just checked this. The icons look great. We could do with 2 small improvements:

1. The "MARK ALL READ" button should also mark all unread.
2. The buttons, when not activated, should be dimmer in color, that way once one is activated, it stands out in white next to the dim grey ones.

ShendMikullovci:
I see I've been requested for a review here. The error is still relevant. I'll check as soon as it's fixed.

PR #1274, opened 2021-12-20
add duplicate button

lilitkarapetyan:
Fixes #1221

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

Screenshot

## Confirmations

- [ ] I have chosen reviewers for my PR.
- [ ] I have assigned myself to this PR.
- [ ] I have chosen an appropriate label for the PR.
- [ ] I have completed the mandatory sections of this document.
- [ ] I have deleted any unused sections.
- [ ] I confirm that I have checked for required README updates and acted accordingly.

PR #1275, opened 2021-12-20
Reapply #1274 fixes to legacy branch

IanMayo:
Fixes #1274 

This branch re-applies @lilitkarapetyan 's changes to the relevant legacy branch.

There is a remaining issue.  When a duplicate of a message is created, the existing message is removed from the browser.

A F5 refresh shows the existing message is still in the database, but:
* the initial message has its `Reference` changed to the new message
* as the new message is handled in `handleNonInfoMessage` the code mistakenly thinks there is a previous instance of the message, and it is deleted.


IanMayo:
Oops, I've updated this branch to match `develop`. That was a mistake. I don't want to have all the new (untested) features.  Will re-create branch.

PR #1276, opened 2021-12-20
Duplicate on legacy - DO NOT UPDATE TO DEVELOP

IanMayo:
Fixes #1274 

This branch re-applies @lilitkarapetyan 's changes to the relevant legacy branch.

There is a remaining issue. When a duplicate of a message is created, the existing message is removed from the browser.

A F5 refresh shows the existing message is still in the database, but:
* during the `duplicate` logic, the initial message has its `Reference` changed to the new message reference
* as the new message is handled in `handleNonInfoMessage` the code mistakenly thinks there is a previous instance of the message, and it is deleted.

So,
1. we have `blue-1` in the database with `_id` of `14:54`
2. we click on duplicate
3. `blue-1` is renamed to `blue-2`, still with `_id` of `14:54`
4. the new `blue-2` is received from PouchDb, with `_id` of `14:55`
5. our `new message` reducer checks for existing instances of `blue-2`. It finds the one at `14:54` and deletes it
6. we now just show `blue-2`.  `blue-1` has been removed.

The problem is in step `3.`. But,  I can't find where it is happening.


IanMayo:
Working perfectly, thanks!

Issue #1234, opened 2021-12-08
Data changes for player log


IanMayo:
This supports #1218

We wish to display a log of active users. The log will display:
* red/green for if they have sent heartbeat request in last minute
* force & role-name
* name of last message sent
* elapsed time since last message sent

To support this data, the backend will track the heartbeat call.  No data will be permanently stored, this will just be an in-memory data object.  Data will be lost if the server is restarted.

These changes will be made:
* change client `heartbeat` call to include wargame-id, force-id, role-id
* embed new functionality in `healthcheck` handler:
    * extract force id, role id, wargame id from call
    * if fields present check for `wargame` entry in `log` dictionary
    * if not present, create it
    * retrieve role-id entry from wargame-log dictionary
    * set current date-time for that entry

The `last message sent` data will be collated in this way:
* on initial connection to the server, the existing set of documents are `get` by the client
    * for each message, retrieve the sender
    * retrieve the `role-messages` entry in the `log` dictionary
    * if not present, create it
    * store the message-type and date-time for that entry
    * overwrite existing entries if necessary
* also apply the above logic for new messages

ShendMikullovci:
I would just like to note that I believe it'd be better to stick to green/grey instead of green/red to display the availability of the players. Grey demands less attentional resources, is standard among messaging apps where the feature is most used and we'd avoid the potential priming of the players towards the grey of the heartbeat icon on the bottom of the screen.

IanMayo:
Thanks for contributing @ShendMikullovci :-)

If the game facilitator is looking out for someone with an issue (or is absent) then wouldn't it be easier to scan the data for a red marker, rather than a grey one (or maybe scanning for the absence of a green marker)?

In this case it does seem significant enough to use a red _warning_ marker.

ShendMikullovci:
From a cognitive standpoint, I don't think red would be significantly more salient in this specific case (though I can't say for sure without experimenting). I suggested we use grey because you had described a playing situation to me once where one (or more) player(s) went out to lunch without notifying anyone. In that case, there's not software issue to handle, but only a matter of availability. Here, the user would scan the log to view if all players are active or how many are/aren't. 

However, you make a good point in saying that IF they were to look for someone with an issue, then a grey marker seems inadequate. A red warning marker (e.g., a red exclamation point) would be fitting for that case.

IanMayo:
Hmm, I've got stuck on this.

[...]

PR #1280, opened 2021-12-22
1234 server player log

IanMayo:
Supports #1234 

Provides server components to support providing player log.


packages/client/src/Views/Version.jsx
```diff
@@ -45,6 +45,32 @@ const Version = ({ showNotification, notifications, hideNotification, wargameId,
...
+    window.location.search
+      .substr(1)
+      .split('&')
+      .forEach((item) => {
+        tmp = item.split('=')
+        if (tmp[0] === 'wargame') result = decodeURIComponent(tmp[1])
+      })
+    return result
+  }
+
```
IanMayo, line 73: Thanks @lilitkarapetyan  - I can see how this works.   @lebaphi  also suggested doing it this way.  Is there really not a way of getting role/wargame from the client state?

IanMayo:
Merging this branch - to stop it diverging too much

PR #1288, opened 2021-12-22
1282: Clickable wargame title

### LINKED ISSUE #1282: Clickable Wargame Title
Similar to issue #1261 we should make the wargame titles appear clickable. An underscore would do.

Wargame Creation Screen 1

lebaphi:
Fixes #1282 

image


IanMayo:
@lebaphi  - something is going wrong with the snaphots. Are use using different versions of some linter/formatter?

This PR causes lots of snapshots to update.  But, the snapshots don't match those created at Heroku - forcing the build to fail.

IanMayo (approved):

PR #1292, opened 2022-01-05
Implement new admin controls for Collab Editing

### LINKED ISSUE #1278: Implement new admin controls for Collab Editing
Implement the UI shown at the end of the thread in #1198 

This is the creation of a new admin page, used to edit Collaborative Editing channels.

The `Channels` admin page will have these changes:
* [x] Change the `Add channel` button to just have `Collab' Editing` and `Mapping`
* [x] On `Collab' Editing`, open new `CollaborativeEditing` page
* [x] New page will allow creation/edit of a new `ChannelCollab` object
* [x] UI for object contained in #1198 
* [x] Instead of existing `Participant` editor table, it will edit instances of `ParticipantCollab`.

lebaphi:
Fixes #1278


IanMayo:
I've tested this via deployment, and the functionality seems to work.  Will test further on local branch.

But... Could we please move "Additional data" above the "Permission" block, so it's beneath "Terminology"?  Otherwise I worry that it could slip off the bottom of the page, and be missed.

lebaphi:
> I've tested this via deployment, and the functionality seems to work. Will test further on local branch.
> 
> But... Could we please move "Additional data" above the "Permission" block, so it's beneath "Terminology"? Otherwise I worry that it could slip off the bottom of the page, and be missed.

sure @IanMayo

IanMayo:
Aah, I see we're not setting the controls in the "participations". Here is the "Blue RFI 2" details for "Papa Foxtrot":
Screenshot 2022-01-06 at 15 11 39

You'll see that the two EXCON roles have false for `canCreate` and true for `viewUnreleasedVersions`.  These are the values for the toggles `Create new message` and `See live updates`.  They also have permission 5.  This integer value relates to these permissions: 
image

IanMayo:
Hello @lebaphi ,
I've had a play in the branch.

The `Participations` now show valid fields (see below). But, the fields aren't editable.  If you don't have chance to work on it, I'll come back & look tomorrow.  I (strongly) suspect part of the solution is casting the `Participant` to `ParticipantCollab`.

image

lebaphi:
> #1278

Hi @IanMayo , I keep the old participant but add more fields into it, you can see the file `setting-channels/helpers/generateRowItems.ts`, code line from 47 to 86.
So I should keep the old `Participant` and to create a new interface like `ParticipantCollab` with new fields.

IanMayo:
Hello @lebaphi 

_One day_ we will get rid of `Participant` - and use `ParticipantCore` and its child interfaces.  For now we check `isCollab` and use the `ParticipantCollab` interface when we know it's valid:
image

Note: I _think_ we need to refactor the editor page for `Channels` tab.  I _think_ we need a separate instance for each type of channel.   That will produce much cleaner code - the same class won't have to handle each different type of channel.  

Then it will be easy/tidy for the channel-type to have its own properties at the top, and different `Participant` columns for different channel types. Unique TypeScript interfaces for the `Channel` and `Participant` for this channel-type will keep the code easy to maintain/understand (such as `ChannelCollab` that contains a list of `ParticipantCollab`).

What do you think?

(But, let's get this version working first, anyway 🤣 )

The new controls in the `Participant` row appear to be either locked, or the handler isn't updating state correctly:
[screencast](https://i.gyazo.com/dbcc155d7a3cf1f15f250c20b7876c5d.mp4)

lebaphi:
> Hello @lebaphi
> 
> _One day_ we will get rid of `Participant` - and use `ParticipantCore` and its child interfaces. For now we check `isCollab` and use the `ParticipantCollab` interface when we know it's valid: image
> 
> Note: I _think_ we need to refactor the editor page for `Channels` tab. I _think_ we need a separate instance for each type of channel. That will produce much cleaner code - the same class won't have to handle each different type of channel.
> 
> Then it will be easy/tidy for the channel-type to have its own properties at the top, and different `Participant` columns for different channel types. Unique TypeScript interfaces for the `Channel` and `Participant` for this channel-type will keep the code easy to maintain/understand (such as `ChannelCollab` that contains a list of `ParticipantCollab`).
> 
> What do you think?
> 
> (But, let's get this version working first, anyway 🤣 )
> 
> The new controls in the `Participant` row appear to be either locked, or the handler isn't updating state correctly: [screencast](https://i.gyazo.com/dbcc155d7a3cf1f15f250c20b7876c5d.mp4)

sure @IanMayo , I think we should make it separate, then we get it easy to maintain later 👍

IanMayo:
> sure @IanMayo , I think we should make it separate, then we get it easy to maintain later 👍

Ok @lebaphi - get it working, then refactor it?  Or refactor it now and then get it working?

lebaphi:
> > sure @IanMayo , I think we should make it separate, then we get it easy to maintain later 👍
> 
> Ok @lebaphi - get it working, then refactor it? Or refactor it now and then get it working?

I fix it now with a bit refactor, I will do more refactor on the weekend 😄

packages/components/src/local/organisms/setting-channels/helpers/generateRowItemsCollab.ts
```diff
@@ -0,0 +1,93 @@
...
+      forceSelected = [forceIndex]
+    }
+  }
+
+  // get selected roles
+  const activeRoles: Array<number> = nextParticipant.roles.map(role => {
+    return roleOptions.findIndex(option => option.value.roleId === role)
+  }).filter(active => active !== -1)
+
+  // if this is a collaborative editing channel then the participant
```
IanMayo, line 34: if this helper is just for collab channels, how about changing the type of the `nextParticipant` parameter to `ParticipantCollab`?

(then we don't need to do the cast down here)

packages/components/src/local/organisms/setting-channels/helpers/generateRowItemsCollab.ts
```diff
@@ -0,0 +1,93 @@
...
+  // init row item for see live updates switch
+  additionalFields.push({
+    type: EDITABLE_SWITCH_ITEM,
+    uniqid: 'see_live_updates',
+    // get default value for switcher
+    active: !!collab.viewUnreleasedVersions
+  })
+
+  const permissionOptions: Option[] = []
+  Object.keys(CollaborativePermission).forEach((key: number | string) => {
```
IanMayo: I'm worried this this could be a number or a string. Doesn't our current implementation force the enum value to be a number?

packages/components/src/local/organisms/setting-channels/helpers/generateRowItemsCollab.ts
lebaphi: I think the function `rowToParticipantCollab` should return the `ParticipantCollab` instead of `Participant` as default 😄

packages/components/src/local/organisms/setting-channels/helpers/generateRowItemsCollab.ts
lebaphi:
```
export enum CollaborativePermission {
  /** can */
  CannotCollaborate,
  CanEdit,
  CanSubmitForReview,
  CanApprove,
  CanRelease,
  CanUnClaim
}
```
this enum has type number @IanMayo :D

packages/components/src/local/organisms/setting-channels/helpers/generateRowItemsCollab.ts
lebaphi: I think we just need to update the `rowToParticipantCollab` to return the `ParticipantCollab ` data type, then we don't need to do a cast, it will need to update data type for some handler of saving data like `handleSaveRows`.

packages/components/src/local/organisms/setting-channels/helpers/generateRowItemsCollab.ts
lebaphi:
```
export interface SelectItem {
  readonly uniqid: string
  type: typeof EDITABLE_SELECT_ITEM
  title?: string
  active?: Array<number>
  multiple?: boolean
  emptyTitle?: string
  options: Array<Option>
}
```
yep @IanMayo , the `active` items is an array of number.

IanMayo (approved):
There is an outstanding issue (that I probably introduced) whereby the admin view of some channels is showing the wrong UI.  To be handled as separate issue.

PR #1294, opened 2022-01-10
Rafactor participants & channels

IanMayo:
Extends work in #1278 
Create PR for branch, to make it easier to review @lebaphi 's suggested changes.


packages/components/src/local/coa-status-board/index.tsx
```diff
@@ -28,10 +28,10 @@ export const CoaStatusBoard: React.FC<Props> = ({ templates, messages, channel,
   const [filteredRows, setFilterdRows] = useState<Row[]>([])
   const [inFilterMode, setFilterMode] = useState<boolean>(false)
 
-  const myParticipations = channel.participants.filter((p) => p.force === role.forceName && ((p.roles.includes(role.roleId)) || p.roles.length === 0))
```
IanMayo, line 31: Hello @lebaphi - I'm pretty sure we can just delete/deprecate this component, and any custom children it has. The data model doesn't work for it, but it has been superceded by `coa-status-board-2`.  Deleting it will prevent us wasting time in having to fix/maintain compiler issues.

packages/components/src/local/molecules/channel-coa-message-detail-2/helpers/visibility.spec.ts
```diff
@@ -100,13 +103,13 @@ describe('Visibility tests', () => {
     expect(userCanSeeCollab(coaChannel, blueLogs)).toBeFalsy()
 
     // white user in collab channel with collab rights
-    expect(userCanSeeCollab(rfiChannel, whiteUmpire)).toBeTruthy()
+    expect(userCanSeeCollab(rfiChannel, whiteUmpire)).toBeFalsy()
```
IanMayo: The test results here have been changed. But, are you confident the new test results are correct?  If you're not sure - then please create a `/TODO: check test results` - so we know the logic is still valid.

packages/components/src/local/molecules/channel-coa-message-detail-2/helpers/visibility.ts
```diff
@@ -37,19 +38,19 @@ const isThisChannel = (channel: ChannelData, channelType: SpecialChannelTypes):
 /** whether this user can see the collaborative aspects of this channel */
 export const userCanSeeCollab = (_channel: ChannelData, _role: ForceRole): boolean => {
   const parts = _channel.participants
-  return !!parts.find((part: Participant) => {
+  return !!parts.find((part: CoreParticipant) => {
```
IanMayo: If you switch to `ParticipantCollab` instead of `CoreParticipant` then the below logic can still be done using the `viewUnreleasedVersions` property.  Aah, the `_channel` parameter won't be `ChannelData`, it will be `ChannelCollab`.  That will make the types easier.

packages/components/src/local/organisms/setting-channels/helpers/createChannel.ts
```diff
@@ -35,15 +36,16 @@ const createChannel = (
       format === SpecialChannelTypes.CHANNEL_COLLAB_RESPONSE
     ) {
       // create new participant
-      const participant: Participant = {
+      const participant: ParticipantCustom = {
```
IanMayo: This should be `ParticipantCollab`. The `if` structure should create the correct channel type - which in turn will create the correct participant type (well, I think it will).

packages/custom-types/channel-data.d.ts
```diff
@@ -22,7 +22,7 @@ export default interface ChannelData {
   /** name of this channel */
   name: string,
   /** list of participants for this channel */
```
IanMayo: I think the _correct_ route is for us to deprecate `ChannelData` - that will prevent us having to maintain it in this way.

packages/custom-types/channel-ui.d.ts
```diff
@@ -12,14 +12,14 @@ export default interface ChannelData {
   templates?: TemplateBody[],
   messages?: Array<MessageChannel>,
   name: string,
-  participants: Array<Participant>
```
IanMayo, line 15: Hmm, I don't like this class. I _think_ it should have a child member that stores an object of `ChannelTypes`.  Then we will still have access to the channel description (including participants), and collab-options if it's a `ChannelCollab`.

packages/custom-types/participant.d.ts
```diff
@@ -64,5 +64,5 @@ export interface ParticipantCollab extends CoreParticipant {
   // participant views un-released documents
   viewUnreleasedVersions: boolean
   // level of access for the participant
-  permission: CollaborativePermission
+  permission: number[]
```
IanMayo: Yes, I see this change.  But, our _model_ is that the Game Designer selects a permission, and the role has all permission levels up to that one.

packages/helpers/src/is-chat-channel.ts
```diff
@@ -1,14 +1,19 @@
-import { Participant, ChannelData } from '@serge/custom-types'
+import { PARTICIPANT_CUSTOM } from '@serge/config'
+import { ChannelData } from '@serge/custom-types'
+import { CoreParticipant, ParticipantCustom } from '@serge/custom-types/participant'
 
 const isChatChannel = (channel: ChannelData): boolean => {
```
IanMayo: Soon, this will get tidier, as we introduce "Add > Chat Channel" to the admin page, which will create an instance of `ChannelChat`.  A custom editor page will look like `Channel-Custom`, but it won't have the templates drop-down.

Have covered the above in https://github.com/serge-web/serge/issues/1295

packages/components/src/local/coa-status-board/index.tsx
IanMayo: Drop this comment - I'm deleting the component.

packages/components/src/local/molecules/channel-coa-message-detail-2/helpers/visibility.spec.ts
lebaphi: The data has been changed, we removed the `canCollaborate` property in the participant, so I think the test is not valid. I just updated the test to make it passed, let you delete it 😄

packages/custom-types/channel-data.d.ts
lebaphi: yep, I think so 👍

packages/custom-types/participant.d.ts
lebaphi: sure @IanMayo, update it back to `CollaborativePermission ` and we should verify the admin form again to make sure the Select permission works correctly :D

lebaphi:
Hi @IanMayo , I just play a bit and found 2 issues as this video below:
- Does not prevent accessing to the legacy channel
- After saving channel collab, it  show wrong  admin controls page :D

https://user-images.githubusercontent.com/26705621/148963932-c5f8e85d-33ab-40ea-8283-1371082e8644.mov

IanMayo:
> Hi @IanMayo , I just play a bit and found 2 issues as this video below:
> 
> * Does not prevent accessing to the legacy channel
> * After saving channel collab, it  show wrong  admin controls page :D
> 

Wow, thanks for that. On it!

IanMayo:
Ok. We now don't try to render the legacy channel.

I've had a look at what the channel _appears_ to lose it's "collaborative editing" controls on save.   It's saving the correct value, so that's not the cause.

But, it does seem to be rendering the wrong channel after the `Save` operation.

Tomorrow, would you mind trying to investigate it please @lebaphi ?

PR #1298, opened 2022-01-13
[1278] Update filter for collab channel

lebaphi:


packages/components/src/local/organisms/setting-channels/index.tsx
```diff
@@ -152,14 +147,7 @@ export const SettingChannels: React.FC<PropTypes> = ({
       setLocalChannelUpdates(localChannelUpdates)
     }
 
-    const handleChangeRow = (nextItems: Array<RowItem>, /* _itKey: number, */ participant: CoreParticipant): Array<RowItem> => {
-      // TODO: this next block was failing, because the [2] item was empty. Find out what the code
```
IanMayo, line 156: Hello @lebaphi  - do you recognise what this commented out block is doing?  It was failing at runtime. I couldn't see what it was doing, so commented it out.  That may have been a silly thing to do 😉

packages/components/src/local/organisms/setting-channels/index.tsx
```diff
@@ -194,9 +182,14 @@ export const SettingChannels: React.FC<PropTypes> = ({
       return dParts.map((participant, participantKey) => {
         const handleSaveRow = (row: Array<RowItem>): void => {
           const nextParticipants = [...data.participants]
-          nextParticipants[participantKey] = isCollab
-            ? rowToParticipantCollab(forces, row, participant as ParticipantCollab) : isChat ? rowToParticipantChat(forces, row, participant as ParticipantChat)
-              : rowToParticipantCustom(messageTemplatesOptions, forces, row, participant as ParticipantCustom)
+          if (isCollab) {
```
IanMayo, line 185: yes, this is more neat 🤣

IanMayo:
I'm not sure about the general idea here @lebaphi .

If we filter the channels so that only collab-channels are seen, we can't see the other channel types (such as chat or mapping).

That's not an option.  But, what are you proposing here?

lebaphi:
> I'm not sure about the general idea here @lebaphi .
> 
> If we filter the channels so that only collab-channels are seen, we can't see the other channel types (such as chat or mapping).
> 
> That's not an option. But, what are you proposing here?

Hi @IanMayo, the idea is do the filter to show all the valid channels in the admin controls page. In the past I see we show a un support message when clicking on the legacy channel data, so we can do a filter here then the legacy channel won’t show up in the admin controls page. We should update the filter condition, I think it’s not enough 😉

IanMayo:
> We should update the filter condition, I think it’s not enough 😉

Ahh, ok. I see.  The _actual_ answer is to delete the legacy channel. I should have done that :-)

But, I _think_ we will still have a problem where switching between `chat` and `collab` channels can still show the wrong columns.

packages/client/src/Views/GameSetupNew.jsx
```diff
@@ -202,8 +205,11 @@ const AdminGameSetup = () => {
       const template = forceTemplate
       template.name = id
       template.uniqid = id
-      template.roles.map(role => role.roleId = getUniquePasscode(forces.forces, "p"))
-      await dispatch(saveForce(currentWargame, id, template, id))
+      template.roles.map(role => {
```
IanMayo: hmm, doesn't `map` return a new array?  The returned results aren't being processed here. Shouldn't it be a `foreach` array function call?
(please teach me better JS :-D ).

lebaphi: I just fix the lint checking :D 
image

lebaphi:
> Ahh, ok. I see. The actual answer is to delete the legacy channel. I should have done that :-)

If we can delete the legacy data, it would be ok. I though to make it work with the legacy data in the old user machine :D

lebaphi:
> But, I think we will still have a problem where switching between chat and collab channels can still show the wrong columns.

I will check it out @IanMayo

IanMayo (dismissed):
Works perfectly, thanks.

packages/client/src/Views/GameSetupNew.jsx
```diff
@@ -55,6 +56,11 @@ const AdminGameSetup = () => {
     channels
   } = data
   const tabs = Object.keys(data)
+
+  // filter out all legacy channels
+  const allChannels = channels.channels
+  channels.channels = allChannels.filter(channel => !channel.format || !LEGACY_CHANNEL.includes(channel.format))
```
IanMayo: I'm going to remove the above block @lebaphi  - the correct fix is to remove (delete) channels in legacy formats. I'll (try) to modify the UI to show a "legacy channel" warning snippet when one is encountered. This will lead me/us to fix the data in the wargame definition.

Issue #1297, opened 2022-01-13
Refactor `version` component


IanMayo:
Ian doesn't like how the `version.jsx` component gets wargame and role from the URL.

@lilitkarapetyan 's proposed solution to the above is:
> we can use version with state if we will render it inside playerUi so we can render Version component inside 
> main PlayerUi provider and duplicate it in on other layout's
>
> 4. we can use version with PlayerUI state if we will render it inside playerUi state provider

PR #1302, opened 2022-01-16
[1296] Implement UI for table of active players

### LINKED ISSUE #1296: Implement UI for table of active players
This is to implement the component to view the table of active players:

image

The development of the above UI is discussed: #1218

1. The component will be displayed in a popup dialog
2. When open, the component will refresh every 5 seconds
3. The data for the columns comes from here:
    a. The list of messages (and ages) comes from `playerLog` in the `PlayerUIState`
    b. The time since last logged in comes from the `\playerlog` REST end-point on the server
    c. The list of players is the _union_ of all of the roles returned by the above two sources
4. The red/green bullet at the left of the `Role` items is green if there is a `playerlog` entry for that role within the last minute

lebaphi:
Fixes #1296 #1297


lebaphi:
Hi @IanMayo, I just had a draft PR for this, I have comments:
- I couldn't find the name of the player log icon, so I temporarily selected `faAddressBook` for it
- I'm not sure how to make a message like `Edit RFI` or `COA Release`?
- The first time you will see it shows the loading process is a bit long because `SERVER_PING_INTERVAL = 20000`, after that it won't take long.
- Do you think we should add something like `Refresh in <number> s` or adding a refresh icon then when fetching new data, we will have an animation then the UX is more friendly I think?
https://user-images.githubusercontent.com/26705621/149646130-4a5222f9-dbee-4e52-b96d-6a443409d0c1.mp4

lebaphi:
with refresh icon :D

https://user-images.githubusercontent.com/26705621/149665993-9bb06e58-0682-4b1f-a658-3ce7de53f200.mov

IanMayo:
Thanks @lebaphi 
> I couldn't find the name of the player log icon, so I temporarily selected faAddressBook for it

Sure, ok. 

> I'm not sure how to make a message like Edit RFI or COA Release?

Sure.  So, just take the current role id, and find the message `from` that role in the playerLog in PlayerUIState

> The first time you will see it shows the loading process is a bit long because SERVER_PING_INTERVAL = 20000, after that it won't take long.

Sure, ok.

> Do you think we should add something like Refresh in <number> s or adding a refresh icon then when fetching new data, we will have an animation then the UX is more friendly I think?

Yes, I think the refresh button should be ok.

packages/client/src/ActionsAndReducers/PlayerLog/PlayerLog_Reducer.ts
```diff
@@ -0,0 +1,18 @@
+import * as ActionConstant from '@serge/config'
+import { PLayerLogAction, PlayerLogPayload } from './PlayerLog_types'
+
+const initialState: PlayerLogPayload = {
+  wargame: '',
+  role: ''
+}
+
+export const addPlayerLogReducer = (state = initialState, action: PLayerLogAction) => {
+
```
IanMayo: can you remove the blank lines please @lebaphi ?

IanMayo:
Thanks @lebaphi 
Do we need the data to be organised in tabs, such as "Overview"?  Have any other tabs been discussed.

packages/client/src/ActionsAndReducers/PlayerLog/PlayerLog_Reducer.ts
lebaphi: updated

lebaphi:
> Thanks @lebaphi Do we need the data to be organised in tabs, such as "Overview"? Have any other tabs been discussed.

You mean we should sort the data by status or send at?

lebaphi:
> Sure. So, just take the current role id, and find the message from that role in the playerLog in PlayerUIState

I tried but it seems `undefined`, the current role id does not exists in the `playerLog`, so I set string `N/A` for those cases :D

IanMayo:
> > Sure. So, just take the current role id, and find the message from that role in the playerLog in PlayerUIState
> 
> I tried but it seems `undefined`, the current role id does not exists in the `playerLog` 🤔

Aah, ok.  Will check when I'm back at my desk tomorrow.  My problem ;-)

IanMayo:
> > Thanks @lebaphi Do we need the data to be organised in tabs, such as "Overview"? Have any other tabs been discussed.
> 
> You mean we should sort the data by status or send at?

No. In the React UI, the table is in a set of tabs, with the current tab being called "Overview".  We don't need a set of tabs.

image

lebaphi:
> > > Thanks @lebaphi Do we need the data to be organised in tabs, such as "Overview"? Have any other tabs been discussed.
> > 
> > 
> > You mean we should sort the data by status or send at?
> 
> No. In the React UI, the table is in a set of tabs, with the current tab being called "Overview". We don't need a set of tabs.
> 
> image

ah, I tried to reuse the `Tabs` component from `@serge/components`:
```
<Tabs activeTab={activeTab} onChange={onTabChanged} tabs={playerLogTabs} />
```
and this component need to input an array of tab to initialize, that is why I create an array with 1 item `overview`

ShendMikullovci (dismissed):
The log looks excellent. Thanks @phileba !

IanMayo:
Hello @lebaphi - I've progressed this a little more, and played with the layout a bit:
image

I have a remaining data issue.  When the panel loads, it gets the `playerLog` from `PlayerUI`.

This object contains the most recent message sent by each role.

But, the object doesn't get updated when new messages are processed.   I tried retrieving the value within the `refresh` method but that threw an `invalid use of hooks` error.

Would you mind trying to sort this for me please?

When it's fixed, sending a message from another user (in another window) should show up in the player log on the next update/.

lebaphi:
> Hello @lebaphi - I've progressed this a little more, and played with the layout a bit: image
> 
> I have a remaining data issue. When the panel loads, it gets the `playerLog` from `PlayerUI`.
> 
> This object contains the most recent message sent by each role.
> 
> But, the object doesn't get updated when new messages are processed. I tried retrieving the value within the `refresh` method but that threw an `invalid use of hooks` error.
> 
> Would you mind trying to sort this for me please?
> 
> When it's fixed, sending a message from another user (in another window) should show up in the player log on the next update/.

sure @IanMayo, I will check it now

PR #1322, opened 2022-01-26
1316 Player Log to react table

### LINKED ISSUE #1316: UI Improvement - Player Log to react table
The current player log isn't interactive.

When there are lots of players the Game Designer may wish to sort or filter some columns.

We should switch the table to the `react-table` component.

It should include these capabilities:
- sort on all colums
- click on column headers to filter values in that column.  This should be for `force`, `role ` and `last message` columns. When new data is received, the filter should still be in place.  Plus, there should be some visual indication that a column has a filter applied.

lebaphi:
Fixes #1316 

https://user-images.githubusercontent.com/26705621/151167684-ea587a1d-bb32-43a0-b7d5-19132b4a0951.mov


IanMayo (approved):
Working great, thanks :-)

PR #1331, opened 2022-01-29
1329 Add column filter for react table

### LINKED ISSUE #1329: React table column filter
In our custom-created table we had a column filter, our own version of this:
image

We've switched to the react data table component, but it doesn't have a column filter. There is a [feature request](https://github.com/jbetancur/react-data-table-component/issues/994) for a column filter.

This issue covers the effort to look at the react-data-table-component sources, and to decide if it is feasible to add a column filter.  

I _imagine_ the API looking like this: by setting a "filterValues" field to "strings", the table would provide a context menu in the column header that shows a list of unique values in that column. The user selects values of interest, and then Apply.  (Note: in the future a "numeric" option could also be added, which allows the user to provide a filter range for numeric data).

If it does look possible we would submit a PR to the `react-data-table-component` project.

lebaphi:
Fixes #1329


IanMayo:
@lebaphi - did you consider changing/extending the react component itself?

lebaphi:
> @lebaphi - did you consider changing/extending the react component itself?

yep @IanMayo , I try to make the function work first and then make it as a new component that extends the react-table component.

IanMayo:
Thanks @lebaphi - could we have a filter on "From" and "Owner" for the "Blue COA", and "From" for the "Blue RFI" please?

Hmm, maybe the collab-options needs to have a data element that lists the columns that should have a filter available.  Or, maybe it's possible to use introspection to decide if the column is suited to a column filter.

IanMayo:
Just had a play - nice UI :-)

1. Could the filter icon be modified in columns that have a filter applied?  I can't see a different icon that works, so I guess we can only play with size or color.
image

Maybe make the standard filter icon one pixel smaller than the current size, and the "filter applied" icon one pixel larger than the current size?  Could you do an experiment with that please?

Or, when applied, we use the light-clue color that's used in `Mark all as read` at the top of the channel?  Could we have an experiment with that too, please?

IanMayo:
Aah, I'll concentrate on playing with the implementation in the player-log.

The table breaks if we apply multiple filters.  So, if we have multiple filters, we can end up showing the same row multiple times.
image

I _think_ we have to apply multiple filters in this way:
* start with full set of rows
* work through columns
* if column has filter, apply filter to set of rows
* for other columns, start with the filtered set of rows, and apply new filter

I hope this helps :-)

IanMayo:
Have deleted my earlier comment.

I've implemented the successive filters change, and I realise we need to keep all the filter options if we're going to allow the player to extend the set of matches.

IanMayo:
Thanks for the update @lebaphi 

Could you have a go at the formatting to indicate which columns have a filter applied?

lebaphi:
> Thanks for the update @lebaphi
> 
> Could you have a go at the formatting to indicate which columns have a filter applied?

Can you have more detail @IanMayo , you mean the column have a filter applied will have the brighter color and the rest are more grey or something like that?

IanMayo:
> > Thanks for the update @lebaphi
> > Could you have a go at the formatting to indicate which columns have a filter applied?
> 
> Can you have more detail @IanMayo , you mean the column have a filter applied will have the brighter color and the rest are more grey or something like that?

Sure - I was imagining we would apply some formatting to the filter icon in columns that have a filter applied.  So - the formatting is just for the icon.

IanMayo:
Happy with new filter icon shading.

Think I just have this remaining issue.  We don't refresh the filter values enough. 

I'm in `Blue COA 2` channel, playing as `Blue-Senior`:
* for `State` I have `In progress` and `Pending review`.  
* if I expand a `Pending Review` and `Release` it, then it goes to `Released`.  

But, the items in the drop-down list for `State` don't change to include `Released`.

Aah, one other thing.  The filters are cleared when a new message (COA) is received.  We should re-run the filters on the new data.  This does mean that a new message may not be visible - but that is the intended behaviour.  Note: even though the messages do not get filtered, the `filter` icon is still shown in white.  I _guess_ we just need to re-run the `getFilteredRows()` method.

packages/components/src/local/react-table/index.tsx
```diff
@@ -135,17 +151,11 @@ export const ReactTable: React.FC<ReactTableProps> = (props) => {
...
-    let filtered = rows
-    appliedFilter.forEach((filter: { col: string, filter: CellFilter[] }) => {
-      // collate list of acceptable values
-      const acceptableValues: string[] = filter.filter.map(f => f.label)
-      // update set of filtered rows
-      filtered = filtered.filter((row: Row) => {
-        // see if this value is in the acceptable values
-        return acceptableValues.includes(row[filter.col])
-      })
+    const acceptableValues: string[] = flattenDeep(appliedFilter.map(filter => filter.filter.map(f => f.label)))
```
IanMayo: Ooh @lebaphi, this is good & clever "js-array" code.  But, I'm worried about the logic.  I may have this wrong, but if the user selects "Closed" in the "Status" column, it will check for "Closed" in other columns.  Aaah, is it doing a compare for every field in every row with the flattened list?

That's quite inefficient if the player has, for example, just asked to see "Status":"Pending" messages.

Happy to discuss :-)

IanMayo:
@lebaphi  - we have a heroku build failure. I've attached a copy of it.  Hopefully you can recognise the issue.
[build log.txt](https://github.com/serge-web/serge/files/7980944/build.log.txt)

packages/components/src/local/react-table/index.tsx
lebaphi: Yeah, I see what you mean, I will update a bit the logic, thank @IanMayo  👍

lebaphi:
> @lebaphi - we have a heroku build failure. I've attached a copy of it. Hopefully you can recognise the issue. [build log.txt](https://github.com/serge-web/serge/files/7980944/build.log.txt)

Hi @IanMayo , it seems the build has an issue with the sass-loader version, I found some one has the issue with the same log with us but different version of the `sass-loader`. I'm not sure why this issue happened while we don't change the `package.json` file 🤔

IanMayo:
> Hi @IanMayo , it seems the build has an issue with the sass-loader version, I found the some one has the issue with the same log with us but different version of the `sass-loader`. I'm not sure why this issue happened while we don't change the `package.json` file 🤔

Yes, it appears the bug is present on `develop` and being pulled into PRs.  I _guess_ it could be an update to an external dependency.

lebaphi:
Hi @IanMayo, I updated the filter logic, could you check it 😉

packages/components/src/local/react-table/index.tsx
```diff
@@ -150,11 +150,11 @@ export const ReactTable: React.FC<ReactTableProps> = (props) => {
...
    * @returns Row[]
    */
-  const getFilteredRows = (appliedFilters: { col: string, filter: CellFilter[] }[]): Row[] => {
-    const acceptableValues: string[] = flattenDeep(appliedFilters.map(filter => filter.filter.map(f => f.label)))
+  const getFilteredRows = (appliedFilter: { col: string, filter: CellFilter[] }[]): Row[] => {
+    const acceptableValues: { label: string, col: string }[] = flattenDeep(appliedFilter.map(filter => filter.filter.map(f => ({ label: f.label, col: filter.col }))))
     return rows.filter(row => {
       // see if this value is in the acceptable values
-      return Object.keys(row).some(key => acceptableValues.includes(row[key]))
+      return acceptableValues.some(({ label, col }) => label.includes(row[col]))
```
IanMayo: I _may_ be mistaken, but...   If the `acceptable` value is `Closed` , and the row status is `Not Closed`, then `includes` means it will be a match won't it?

Please correct me if I'm wrong, but using `===` seems better because:
* only exact matches pass the test
* it _may_ be quicker, since it's just comparing the hash, rather than a character by character comparison.

IanMayo:
> Hi @IanMayo, I updated the filter logic, could you check it 😉

Thanks. Have replied [here](https://github.com/serge-web/serge/pull/1331/commits/a1a2b6743597f8d180e214d2c0780fb9f701aea1#r798332556)

packages/components/src/local/react-table/index.tsx
lebaphi: > value is Closed , and the row status is Not Closed

ahh, okay @IanMayo, in this case we should use `===`, I will updated it 👍

lebaphi:
> I'm probably being silly here. But, won't the above do a comparison with the string `${row[col]}`?

```
return rows.filter(row => {
      // see if this value is in the acceptable values
      return acceptableValues.some(({ label, col }) => label.trim() === `${row[col]}`.trim())
    })
```
so the logic: 
- we have the array acceptableValues: `{ label: string, col: string }[]`, e.g: [ {label: 'Closed', col: 'status'}, {label: 'Not Closed', col: 'status'} ]
- we filter all rows, then loop `acceptableValues` array and check each row if has `row[col] === label` => return `Passed`. e.g: if row['status'] === 'Closed' => return `Passed` and continue with the next row.

So what is your concern @IanMayo ?

IanMayo:
Hello @lebaphi 

You have:
```
 label.trim() === `${row[col]}`.trim())
```
Shouldn't it be this (without the quotes around `${row[col]}`.
```
 label.trim() === ${row[col]}.trim())
```

My guess is that the test is against the text string `"${row[col]}"` rather than the contents of  `${row[col]}`

lebaphi:
> Hello @lebaphi
> 
> You have:
> 
> ```
>  label.trim() === `${row[col]}`.trim())
> ```
> 
> Shouldn't it be this (without the quotes around `${row[col]}`.
> 
> ```
>  label.trim() === ${row[col]}.trim())
> ```
> 
> My guess is that the test is against the text string `"${row[col]}"` rather than the contents of `${row[col]}`

ahh, I just cast it to string, it called template string in Js https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals?retiredLocale=vi, we can not use ${...} without the quotes ` ` 😄

IanMayo:
> ahh, I just cast it to string,

Perfect, thanks for the lesson ;-)

IanMayo (approved):
Approving changes. We still have the failing build issue, but that's on `develop` - not just here.

Issue #1317, opened 2022-01-21
Improved interaction tracking for Player Activity Log


IanMayo:
The Player Activity Log is intended to show when a player was last active.

The best test we have for this is when the `player log` message is sent, immediately after the `heart beat`.

This message carries no parameters, but the server stores the `current time` as the time of last interaction.

But, a player may be away from their Serge instance, and Serge will still think the player is interacting.

A better measure would be the time that the player last interacted with Serge.  We could to this by tracking some events happening. As they happen we would update a global `lastActivtyTime` to the current time.  It would be this time which is sent in the `playerLog` message.

We would track these actions:
* changing channel
* opening/collapsing a message
* sending a message

The `server.js` would have to be modified to store the new `lastActivityTime` rather than the current time.

IanMayo:
Fixed in #1334

PR #1334, opened 2022-02-01
add action for healtcheck api

lilitkarapetyan:
## 🧰 Issue

#1317 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Thanks @lilitkarapetyan - but this is more advanced than we need.

It _looks like_ this PR tells the server when an interaction happens.

I don't think we need that level of network activity.  Instead, I think this is enough:
- when these interactions happen:
    - store the time of the interaction (`mostRecentActivity`)
- when heartbeat request is sent:
    - pass the `mostRecentActivity` time in the request, which the server stores (instead of the time of heartbeat)

I hope this makes sense.

IanMayo:
Hello @lilitkarapetyan - it appears some dependency change in `develop` is crashing the Heroku build.

When I updated this PR, Heroku failed. Here is the log (from another PR - but the error is the same).

[build log.txt](https://github.com/serge-web/serge/files/7988322/build.log.txt)

Could you investigate it please?

IanMayo:
Nice work, thanks @lilitkarapetyan 

```
expiredStorage.setItem('activityTime', `${new Date().getTime()}`)
```

Could you swap instances of this bit of code with calls to a function please?

Then, if we want to modify the _activity-tracking_, we only have to do it in one place.

IanMayo:
@lilitkarapetyan - the logic is still a little confused here.

We have the `healthcheck` and `playerlog` messages.  Currently:
* `healthcheck` is passing the `activity-time` in the **get** which returns the status
* `playerlog` is **posting** `wargame` and `role` - and the server is injecting the current time
* `playerlog` is **getting** the log of activity for the specified wargame

We need to modify `healthcheck` to send `activity-time`, `activity-type`, `wargame` and `role` (all of the `PlayerActivity` elements.  The server should then store this data using the logic in `app.post("playerlog")` in `server.js`.

Then we don't need to **post** playerlog.  We will just include the details when we  **get**  the healthcheck 

Note: for the player-log I'd like us to start tracking the `activityType` too.  I've made the client-side changes - but we need to send it to the server & store it there, too.

packages/client/src/api/wargames_api/index.ts
```diff
@@ -159,9 +160,12 @@ export const listenForWargameChanges = (name: string, dispatch: PlayerUiDispatch
...
 }
 
-export const pingServer = (): Promise<any> => {
+export const pingServer = (activityDetails: {wargame: string, role: string}): Promise<any> => {
   const activityTime = expiredStorage.getItem(ACTIVITY_TIME) || 'The player has not shown any activity yet'
-  return fetch(serverPath + 'healthcheck/' + activityTime + '/healthcheck', {
+  const activityType = expiredStorage.getItem(ACTIVITY_TYPE) || 'The player has not shown any activity yet'
+  const activityLogs: ActivityLogsInterface = { activityTime, activityType, ...activityDetails }
+
+  return fetch(serverPath + 'healthcheck/' + JSON.stringify(activityLogs) + '/healthcheck', {
```
IanMayo: @lilitkarapetyan - would it be a problem if we passed the data as four fields , instead of using `stringify`?  The use of `stringify` is making the call harder to read:
image

There may be a strong reason to keep `stringify`.  If there is, please just say.

IanMayo:
I'm pretty sure there is an outstanding issue with this.

If I remember correctly, I don't think the `node` server is returning the correct `last-activity` value for the current user. I was logged in as `Game Manager`, but it said I hadn't interacted for 20 minutes:
image

Issue #1333, opened 2022-02-01
Two issues with Platform-Types in Game Admin page


IanMayo:
We have two issues with the Platform-Types editor page.

The first is that a platform-type can't be deleted until it has been saved/modified (as illustrated [here](https://github.com/serge-web/serge/pull/1323#issuecomment-1024248765) )

The second is that if we create several platform-types, deleting one deletes all of them - as illustrated here.

![screencast](https://camo.githubusercontent.com/478507df66651891a2f781904b8d69f1c00f4c6e0ed2089b5e4c07693a9dcbde/68747470733a2f2f692e6779617a6f2e636f6d2f36366461656633616437663138393938323866386261623434343565343434312e676966)

ShendMikullovci:
I encounter this issue frequently but with various manifestations. Today, I couldn't delete a platform before nor after saving. Then, after closing the tab and restarting a game, the platforms were immediately saved after creation, making their deletion possible.

IanMayo:
Could you look at this too, please @lilitkarapetyan ?

IanMayo:
Tested on `serge-review`, working.

PR #1342, opened 2022-02-09
delete platform type without saving

lilitkarapetyan:
## 🧰 Issue

#1333

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/components/src/local/organisms/setting-platform-types/index.tsx
```diff
@@ -88,8 +89,22 @@ export const SettingPlatformTypes: React.FC<PropTypes> = ({ platformType, onChan
   }
 
   const handleDelete = (item: Item): void => {
+    const db: PouchDB.Database = new PouchDB(databasePath + wargameName + dbSuffix)
     setSelectedItem(-1)
-    onDelete && onDelete(item as PlatformType)
+    db.get(localSettings).then((res: any) => {
```
IanMayo: @lilitkarapetyan - is this specific to PouchDb?  Will it work with RavenDB?

Hmm, the data is only in `_local/settings` until the game has been `Finalized`. Then the newer versions are in the normal collection of documents.

Hmm, I'm still confused. I don't think other tabs have to do their own database access - they just use the API.  The `Forces` tab doesn't do database access - but it doesn't have the issue we're trying to fix here.  Could you have a look and see how `setting-forces` handles it's data.

packages/components/src/local/organisms/setting-platform-types/index.tsx
```diff
@@ -88,8 +89,22 @@ export const SettingPlatformTypes: React.FC<PropTypes> = ({ platformType, onChan
   }
 
   const handleDelete = (item: Item): void => {
+    const db: PouchDB.Database = new PouchDB(databasePath + wargameName + dbSuffix)
```
IanMayo: ~~Hmm, @lilitkarapetyan - this feels like a bit of a bad code-smell.  It's only when `SAVE` is clicked that data goes to the database.  So, updating the client state should be enough.~~

~~Can you help me understand why we have the database round-trip?~~

Aah, sorry - I get it now. We have `Save` for platform-type, not for the full array of platform-types.

IanMayo:
Thanks @lilitkarapetyan , I ❤️  the new version.

IanMayo (approved):
Great solution.

PR #1349, opened 2022-02-16
1332: Improve editor component platform speed

### LINKED ISSUE #1332: Improved editor component for design of platform speeds
We're trying to see if we can re-design and update the tracking of asset speed. Our current design takes up too much space on the admin page and is not optimal for tracking. 

This was the initial idea:

Speed Box
 
Here, the user can change the interval values and length. Selecting one value auto-selects the previous values: The selected speed in this design is 18KTS, which means the asset can also go at 6 & 12 KTS if necessary.

@IanMayo brought up the following ideas: _It's possible for a warship to have a maximum speed for "flying operations", that's the maximum speed the can travel at and still use their helicopter. But, an autonomous underwater vehicle (AUV) could have a "most efficient speed" - that's the speed that maximises search range and battery endurance. There is a chance that this speed could be expressed in metres/second. So the AUV could have an "Optimal search speed" with units of "m/s". The Game Designer may provide different "optimal search speeds" for different instances of AUV._

The following questions arise: 
- Why would the most efficient speed be measured in m/s and not KTS? 
- Have we discussed this with the client? 
- Can they provide details on the most efficient speed per asset type based on real-life scenarios?

lebaphi:
Fixes #1332

### DIFF
--- a/packages/components/src/local/molecules/sortable-list/index.tsx
+++ b/packages/components/src/local/molecules/sortable-list/index.tsx
@@ -16,7 +16,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faCopy, faTrash, faGripVertical } from '@fortawesome/free-solid-svg-icons'
 
 /* Render component */
-export const SortableList: React.FC<PropTypes> = ({
+export const SortableList: React.FC<PropTypes> = React.forwardRef(({
   onClick,
   onChange,
   items,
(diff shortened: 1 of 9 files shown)


IanMayo:
Thanks @lebaphi 

The `Add speed` button on the admin page needs to be changed to `Edit`.  Hmm, or how about clicking on the read-only speeds to popup the dialog editor panel?

IanMayo (approved):
Works great, thanks.

PR #1359, opened 2022-02-24
1348 orientation marker data

### LINKED ISSUE #1348: Mapping changes for close-in wargames
The games that used the mapping feature in the past were performed on a map that covered 10s or hundreds of miles.

In the next wargame we wish to investigate supporting wargames running on a much smaller scale.

* [x] Create wargame covering a few miles, with small hex cells
* [x] Create `Asset` icon that is capable of showing current orientation (according to planned route)
* [x] Create `Asset` icon that is capable of showing orientation according to some asset attribute.

Calculate Asset orientation as one of these:
* [x] Value of direction to next planned step, if specified
* [x] Value of direction from most recent history, if specified
* [x] Absolute value of attribute
* [x] Value of attribute relative to heading

IanMayo:
Fixes #1348


IanMayo:
That's great.

The angle it's pointing at is (temporarily) hard-coded in `assets\index.tsx`. Could you change the `orientation` value in there and check they still align correctly please?

packages/components/src/local/asset-icon/styles.module.scss
```diff
@@ -91,12 +91,8 @@ $platforms: agi, auxiliary, boghammer, carrier, coastal-radar-site, datum,
...
 
 .orientation {
-  background-image: url("./counters/orientation-marker.svg");
-  background-size: 80px;
-  fill: #f00;
-  color: #0f0;
-  filter: hue-rotate(244deg);
-  background-position: 25px 25px;
+  mask: url("./counters/orientation-marker.svg");
+  background-color: rgb(190, 48, 48);
```
IanMayo: Aah, we're using RGB, but they are still hard-coded in the CSS.  Could you make it so we can specify the color in `index.tsx` please?
I'm pretty sure it will be specified in `#f48` notation.

lebaphi: okay @IanMayo

IanMayo:
Hmm, @lebaphi  - I'm seeing this. Am I doing something wrong?
image

[Here](https://serge-1348-orientation--wjsqfk.herokuapp.com/storybook/?path=/story/local-localmapping--natural-earth) is story I used to make the above.

lebaphi:
> Hmm, @lebaphi - I'm seeing this. Am I doing something wrong? image
> 
> [Here](https://serge-1348-orientation--wjsqfk.herokuapp.com/storybook/?path=/story/local-localmapping--natural-earth) is story I used to make the above.

You are not wrong @IanMayo, it seems to work with `yarn develop` but not with `yarn storybook`, I guess the problem comes from the css renderer when it is compiled. I pushed the fixed already.

IanMayo:
Thanks @lebaphi 

But, the icon still isn't "centred" on the centre of the asset icon.  In `assets\index.tsx` I changed the `orientation` from `320` to `220`, and got this:
image

I _think_ we may have to tell the browser (or maybe leaflet) where the "origin" of the SVG marker is.  Hmm, maybe as discussed in the [API](https://leafletjs.com/reference.html#icon) or [blog](https://leafletjs.com/examples/custom-icons/)

No problem if that's beyond your skills/experience - it is quite "niche".

lebaphi:
> Thanks @lebaphi
> 
> But, the icon still isn't "centred" on the centre of the asset icon. In `assets\index.tsx` I changed the `orientation` from `320` to `220`, and got this: image
> 
> I _think_ we may have to tell the browser (or maybe leaflet) where the "origin" of the SVG marker is. Hmm, maybe as discussed in the [API](https://leafletjs.com/reference.html#icon) or [blog](https://leafletjs.com/examples/custom-icons/)
> 
> No problem if that's beyond your skills/experience - it is quite "niche".

yes, I recognized this issue but we are rendering these 2 markers as 2 separate elements. A solution is needed to anchor one element to the center of another. I'm thinking about this.

IanMayo:
We give the map the `lat/long` for the each icon. We give the same lat/long to put them in the same place.

It's currently using the centre of the icon rectangle as the origin. That works fine for the circular icon. But, the pointer isn't square - so when we rotate it, it is rotating around the "geometric centre" - not the centre of the circular part.

I _think_ the `iconAnchor` code tells Leaflet/browser which xy offset to treat as the "centre of rotation".

PR #1367, opened 2022-03-03
1362: Update asset attribute when saving force

### LINKED ISSUE #1362: Two features missing in `attributes`
I've just realised we are missing two capabilities related to `attributes`. They are both in the admin pages.

image

1. When adding a new `asset` to a `force`, lookup the `attributes` for that `platform-type` and add the relevant `attributes` to the new `asset`. An `asset` is added to a force by dragging one of the platform-type icons into the `List of assets` section.
2. Before allowing a `SAVE FORCE`, do a check that the `attributes` are correct for each asset. Specifically, do this:
    a. Find the platform-type for the asset
    b. Check that each attribute is present on the asset
    c. Check that no extra attributes are present on the asset
    d. Popup a warning if anything is wrong.  The dialog should include the asset name(s) and what the problem is.

lebaphi:
Fixes #1362


lebaphi:
Hi @IanMayo , I refactored a bit our dialog, I create a new component called `CustomDialog` which allows us to easily display a modal and easily customize its style. I also updated the `Confirm` dialog to use `CustomDialog` internally. Now all dialogs will have consistent style.

- Show an error when validate an asset failed
Screen Shot 2022-03-04 at 5 16 03 PM

- New confirm dialog (wrap Custom Dialog inside)
Screen Shot 2022-03-04 at 5 15 01 PM

IanMayo:
Hello @lebaphi - I've made some changes, but couldn't quite get things working.

If the attributes of a force asset don't match the platform-type, the Game Designer is unable to fix that with the current UI.

But, if they don't match - it's quite simple to fix any problems: we can remove extra attributes, and insert missing ones.   I've implemented the code to that.  It also puts a message up if changes had to be made. 

But, the UI doesn't update, we still see the previous (partially wrong) set of attributes - would you mind investigating what is going wrong please?

Note: I've been testing with `NORT` on force `blue-1`.

IanMayo:
Thanks @lebaphi 

One change.  We have the `player can edit` flag.  We ignore that flag when we're in the Game Admin pages.  The Game Designer can change any attribute.  So - we can use the dialog, but we need to use a special flag when the editor is being used from the game admin pages.  Suggested the flag (prop) is called `userIsAdmin`.

packages/components/src/local/organisms/setting-forces/index.tsx
```diff
@@ -115,9 +115,21 @@ export const SettingForces: React.FC<PropTypes> = ({
...
+      const attrsbuteErrorList = attributeErrors.reduce((html, item) => {
+        html += `<li>${item}</li>`
+        return html
+      }, '')
+      attributeErrors.length > 0 && toggleModal(`The attributes for some assets did not match with type details. These fixes have been applied: <br/> ${attrsbuteErrorList}`)
 
       if (onSave) {
+        forcesData.some(force => {
+          if (force.uniqid === selectedForce.uniqid) {
+            // TODO: should loop each asset and assign only the `attributeValues` for each one instead of assign the whole asset to force?
```
IanMayo, line 127: Good thinking @lebaphi  - could you do this improved processing please?

IanMayo:
I'm pretty sure we're still missing one feature @lebaphi :
- when we add a new asset, it should get the attributes from the parent `platform-type`, including the default values (when present), otherwise zero.

IanMayo:
> I'm pretty sure we're still missing one feature @lebaphi :
> 
> * when we add a new asset, it should get the attributes from the parent `platform-type`, including the default values (when present), otherwise zero.

@lebaphi - I fixed this, in this commit: https://github.com/serge-web/serge/pull/1367/commits/8a08158f1db21e0c5308861cca3f7a10895f6670

packages/components/src/local/organisms/setting-forces/index.tsx
lebaphi: okay @IanMayo , I will fix it in another PR 👍🏻

PR #1380, opened 2022-03-05
MoreInfo component is overridden after dialog

### LINKED ISSUE #1368: MoreInfo component is overridden after dialog
We can see the MoreInfo component displays behind the dialog

image

lebaphi:
Fixes #1368 

image


IanMayo (approved):
Tested and approved

PR #1401, opened 2022-03-11
1385: Fix message display inconsistency

### LINKED ISSUE #1385: Message Display Inconsistency
Couple of things I wanted to mention: 

Chat Window Incongruity

There's inconsistency in how we display messages on channels vs general chat. In the message channels, older messages move higher up and newer messages occupy immediate lower spaces. This is a general consensus among messaging apps accross the board and I think we should apply it for the general chat as well. Currently, the general chat displays the opposite: old messages down, new messages up.

Also, in the general chat as you can see in the screenshot above, it's difficult to visually discern between teams simply based on the 2 colored small squares next to the message. Furthermore, unless the user is viewing their own messages where they see this message cloud type of icon below: ...

Own Message General Chat

... then the message is indistinguishable from the white background. 

I suggest we grab the same design from the channel messages and apply for the general chat. The grey would help distinguish the messages and the colored sidebar would stand out more than a small colored square. It would look something like this: 

New General Chat UX

Let me know your thoughts.

lebaphi:
Fixes #1385 

Screen Shot 2022-03-11 at 10 11 51 PM


IanMayo:
looks good. ok with you @ShendMikullovci ? Or do we need the force colour bar on the right of our own messages.

lebaphi:
> looks good. ok with you @ShendMikullovci ? Or do we need the force colour bar on the right of our own messages.

yeah, I see the legacy logic does not  render the color bar on the right of our own messages, if you want this,  I will update :D

ShendMikullovci:
The design itself looks great. The colors are nice and bold, clearly distinguishable. 
A couple of things we could fix: 

1. Own message doesn't need its color bar. However, it might be useful to make it darker:

Own Message Screenshot Edited

This way, we fully distinguish the messages and create the necessary me/them dichotomy present in most messaging applications. It doesn't have to be this dark grey either, just grey enough to bring it out from the white background and differentiate it from other force messages.

2. White is invisible: 

White Force Color Bar Invisible

As you can see, when viewing from a different force, the White Force color sidebar is virtually invisible. A simple way to fix this that I could think of is maybe adding an outline around the color? That would emphasize the white inside the outline.

IanMayo:
Thanks for that @ShendMikullovci 

1. Own force doesn't need color marker.  Sure, ok.
2. Make own messages slightly darker.  Sure, but as you say - not as dark as your example
3. Border around force color marker.  Yes, good idea.

I'll leave it with you @lebaphi :-)

ShendMikullovci:
> 2. Make own messages slightly darker.  Sure, but as you say - not as dark as your example

Here's a better color example: 
Example color

Color code: #d9d9d9ff

lebaphi:
Hi @IanMayo @ShendMikullovci , I've updated a bit of the UI, I think it's better

image

But I think we have a issue with message `hasBeenRead`, as you can see the message always bold although I click the button `Mark all as read` 🤔

ShendMikullovci:
> Hi @IanMayo @ShendMikullovci , I've updated a bit of the UI, I think it's better
> 
> image

Much better. Great work.

> 
> But I think we have a issue with message `hasBeenRead`, as you can see the message always bold although I click the button `Mark all as read` thinking

Apparently, we do. It seems to work for the Insights tab for GC, but not for the messages.

IanMayo:
@lebaphi - could you change the word-wrap so that it doesn't split up words, please?
image

lebaphi:
> @lebaphi - could you change the word-wrap so that it doesn't split up words, please? image

okie @IanMayo

IanMayo (approved):
Working fine.

ShendMikullovci (approved):
Just reviewed. It all looks and works fine. The issue with "MARK ALL READ" persists, as well as a small rectangle eating up the chat space. I'll add a new issue for these.

PR #1404, opened 2022-03-11
1402 Hotfix draggable color editor

### LINKED ISSUE #1402: Drag-able Color Pin
I was creating the wargame for issue #1400 and I noticed that when selecting the Force color, the color pin can't be dragged around in the color chart: 

https://user-images.githubusercontent.com/85491794/157916710-25e0d1d2-c428-458e-bcc8-e8ba1cf6c609.mp4

As you can see, I can only select a color by clicking on an area in the color chart. It would be better if the user could drag the color pin and view the preview on the icon color box. That would give them more and better control of their color choice (that, for a wargame with lots of different Force color nuances, is very important).

IanMayo:
Fixes #1402 

Recording: https://i.gyazo.com/89dc98d43a7188cfe213786d245f3f84.mp4


ShendMikullovci:
I was just going to add that clicking on the color interval should only change the chart, and not automatically select the clicked color, but I can see you fixed that too. Perfect!

ShendMikullovci (approved):

PR #1408, opened 2022-03-12
1391: Fix button color issue

### LINKED ISSUE #1391: Cancel & Save Button Color
Was just reviewing #1368 and realized that the Cancel and Save buttons are of the same color: 

Cancel   Save Button Same Color

I suggest we stick to the usual bold navy blue for the save button and keep the grey for the cancel button.

lebaphi:
Fixes #1391 

Screen Shot 2022-03-12 at 3 41 27 PM


IanMayo (approved):

packages/components/src/local/organisms/setting-forces/index.tsx
```diff
@@ -119,17 +119,25 @@ export const SettingForces: React.FC<PropTypes> = ({
...
+        // If not, just save the forcesData
+        if (attributeErrors.length) {
+          forcesData.some(force => {
+            if (force.uniqid === selectedForce.uniqid && force.assets) {
+              force.assets.forEach((asset, idx) => {
+                if (selectedForce.assets) {
+                  asset.attributeValues = selectedForce.assets[idx].attributeValues
+                }
+              })
+              return true
```
IanMayo, line 136: if we are using `some` won't this code drop out on the first `true` value?  My guess is that we should run through all of the forces.

ShendMikullovci (approved):
Great! Thanks guys!

PR #1411, opened 2022-03-13
1397 Update Admin Page Entry Display

### LINKED ISSUE #1397: Admin Page Entry Display
This is our current Admin Page Entry Display: 

Current Admin Page Entry Page

It displays no situational clues for the user to identify what page they are on. "Password" is shown twice on the page, which is a little redundant, and with the lack of clues, the user might not know what password we are referring to. 

I suggest the following additions: 

New Admin Page Entry

The SERGE logo with the title below, as well as the name of the page the user is on. The password title is removed and the password text box is aligned to the center of the screen, along with the logo and the two titles below it. I am also considering a small, introduction (most likely one to two sentences) similar to our Welcome Screen. 

Also, the ENTER button has been turned navy blue since it's an executive button.

What do you guys think?

lebaphi:
Fixes #1397 

image


packages/components/src/local/organisms/admin-login/index.spec.tsx
```diff
@@ -4,6 +4,11 @@ import renderer from 'react-test-renderer'
 
 import AdminLogin from './index'
 
+jest.mock('react-redux', () => ({
+  ...jest.requireActual('react-redux'),
+  useSelector: jest.fn(() => [])
+}))
+
```
IanMayo, line 11: Can you explain what this block of code does please @lebaphi ?

packages/components/src/local/organisms/admin-login/index.tsx
```diff
@@ -1,50 +1,65 @@
...
         Enter
-      </Button>
+        </Button>
+      </div>
     </>
   )
 }
 
-export default AdminLogin
+export default React.memo(AdminLogin)
```
IanMayo: @lebaphi - I don't know if we need to use `memo` here.  The admin page rarely opens, and when it does, there is only one instance of it.

Can you point me to a page/article that suggests we use `React.memo` everywhere?

packages/custom-types/root-state.ts
```diff
@@ -0,0 +1,58 @@
...
+  loadingWargames: boolean;
+}
+
+export type CurrentModal = {
+  open: boolean;
+  modal: string;
+  data: null;
+}
+
+export type RootState = {
```
IanMayo, line 46: Hello @lebaphi - can you explain what this new RootState is for? 

Was it an unstructured `any` in the past?

packages/components/src/local/organisms/admin-login/index.spec.tsx
lebaphi: in the `admin-login` component I use a hook `useSelector`, so when implement the unit test with jest, we should mock this hook to make the test passed because the jest will not know what is the `useSelector` is.

packages/client/src/Store/RootReducer.ts
```diff
@@ -0,0 +1,23 @@
+import { CurrentModal, DBLoadingState, ExportItems, GameInfo, Messages, UmpireMenu } from "@serge/custom-types"
+import { messagesReducer } from "../ActionsAndReducers/dbMessages/messages_Reducer"
+import { exportItems } from "../ActionsAndReducers/ExportItems/ExportItems_Reducer"
+import { loadingDbReducer } from "../ActionsAndReducers/loadingDb_Reducer"
+import { currentModal } from "../ActionsAndReducers/Modal/Modal_Reducer"
+import { notificationReducer } from "../ActionsAndReducers/Notification/Notification_Reducer"
+import { gameInfo } from "../ActionsAndReducers/sergeInfo/sergeInfo_Reducer"
+import { umpireMenuReducer } from "../ActionsAndReducers/UmpireMenu/umpireMenu_Reducer"
+
```
IanMayo, line 9: Hello @lebaphi - there is a big change here, but I can't quite see what it happening.  Could you have another go at explaining it please?

Aah, I think I've got it.  The reducers are written in JS rather than TS.   Right?

IanMayo:
@lebaphi the Serge icon on the main welcome page can be specified by the Game Designer in the `Welcome Screen` part of Game Admin.  Does the admin login page use the same logo?

image

packages/components/src/local/organisms/admin-login/index.tsx
lebaphi: the `admin-login` component has the callback function `onSubmit`, so we should apply the memoization. You can see here https://dmitripavlutin.com/use-react-memo-wisely/#:~:text=memo()%20wisely.,memoization%20for%20class%2Dbased%20components, and all callback functions should be wrapped in `useCallback` to reduce the component rendering.

packages/custom-types/root-state.ts
lebaphi: yep, in the past we use
```
export default createStore(combineReducers({
  ....
}), applyMiddleware(...middlewares))

```

we don't define the type the state of the store, so it uses the `DefaultRootState` as default, then all state which select from store will have the `any`.
Now I create an interface for the state, I called it `RootState` with all type  of all reducers. So we can query the state that we want with its property like the image below:

image

packages/client/src/Store/RootReducer.ts
lebaphi: yep, as you can see all our reducers were written in JS so they don't have a type for it. I created a new class to wrap each old reducer with the new one (in TS file) then we can have a type for it.

lebaphi:
Yes, it uses the same logo @IanMayo , you can see in the code, I used `${serverPath.length}/${gameInfo.imageUrl}` to load this image 😄

packages/components/src/local/organisms/admin-login/index.tsx
IanMayo: Thanks @lebaphi - but I think that article says we should avoid using `React.memo` unless we know there is a specific performance issue:
image

lebaphi: Another one @IanMayo 🤣  https://javascript.plainenglish.io/when-and-when-not-to-use-react-memo-e050d3cb1ffc
image

lebaphi:
```
When we have pure functional components. Given the same props, such components will remain the same — always rendering the same output.
```
In case if the state does not changed, using `React.memo()` will help use to avoid the component re-render often I think. So I think it's a good habit.

IanMayo: Sure.  But the `admin page` doesn't re-render a lot.  Hmm, I don't think it even takes any props.  If it does, they won't change.

So, this _feels like_ we're adding complexity without justification.

lebaphi: yeah, so we can remove it 😄

IanMayo (approved):
Working fine.

Issue #1420, opened 2022-03-15
Delete Role Bug


ShendMikullovci:
It appears roles can't be deleted once they are added to the list on a new wargame: 

![Delete Role Bug](https://user-images.githubusercontent.com/85491794/158413628-ee4f5117-c613-4e04-a7eb-48986afcf3dc.png)

This issue is also present in already existing wargames: 

![Delete Role Bug Small Wargame](https://user-images.githubusercontent.com/85491794/158413685-d0d13a3f-5386-4f9d-b8a1-f191a757c16a.png)

IanMayo:
Another for you, please @lilitkarapetyan . Maybe this is related to the other instances of the issue.

IanMayo:
Fixed in #1423

ShendMikullovci:
I think we agreed for the bin to be on the left, next to the name, before closing?

![Bin on the right](https://user-images.githubusercontent.com/85491794/159977764-c0bff25f-9afb-4e77-bdc7-a4b7d8db7ba4.png)

It's still on the right.

IanMayo:
Oops, well spotted.

@lebaphi - could you move the `trash` icon next to the `Roles` entry please?

ShendMikullovci:
Unfortunately, this bug has reappeared: 

https://user-images.githubusercontent.com/85491794/162999890-8e654e76-602f-4a9d-906a-6892b6fcdd0f.mp4

ShendMikullovci:
This has been fixed.

PR #1423, opened 2022-03-16
1420: Delete Role Bug

lilitkarapetyan:
## 🧰 Issue

#1420

## 🔨Work carried out:

- [x] Tests pass

## Confirmations

### DIFF
--- a/packages/components/src/local/molecules/sortable-list/index.tsx
+++ b/packages/components/src/local/molecules/sortable-list/index.tsx
@@ -66,18 +66,7 @@ export const SortableList: React.FC<PropTypes> = React.forwardRef(({
   }
 
   const handleRemove = (key: number): void => {
-    const newItems = [...items]
-    const role = newItems[key]
-    if (typeof role === 'object') {
-      if (role.isGameControl) {
-        if (typeof onDeleteGameControl === 'function') {
-          onDeleteGameControl(role)
-          return
-        }
-      }
-    }
-    newItems.splice(key, 1)
-    handleChange(newItems)
+    onDeleteGameControl && onDeleteGameControl(items, key, handleChange)
   }
 
   const sortableItems: Array<SortableItem> = items.map((item: Item, key: number) => {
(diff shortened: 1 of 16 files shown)


packages/components/src/local/organisms/setting-forces/settings-force-roles/index.tsx
```diff
@@ -134,7 +134,7 @@ export const RolesAccordion: FC<PropTypes> = ({ data, handleChangeForce, forces,
           <div className={cx(styles.col, styles.section)}>
             <FormGroup placeholder="Roles">
               <SortableList
-                remove={false}
+                remove={true}
```
IanMayo, line 137: @lilitkarapetyan - I've just been investigating, but I don't understand what this `remove` prop is doing.  Could you tell me please?  Then I'll add a javadoc to the props.

lilitkarapetyan: This boolean prop for the visibility of the delete button

IanMayo (dismissed):

packages/components/src/local/organisms/setting-forces/settings-force-roles/index.tsx
IanMayo: But there is some strange `removeLocal` code, too: 
image

IT _looks-like_ if `remove` is missing, then we still show the `trash` icon.

lilitkarapetyan: It seems to me that this part of the code is not needed because it should always be true, we see the icon when we hover over the role with the mouse

packages/components/src/local/molecules/sortable-list/index.tsx
```diff
@@ -66,18 +66,7 @@ export const SortableList: React.FC<PropTypes> = React.forwardRef(({
   }
 
   const handleRemove = (key: number): void => {
-    const newItems = [...items]
```
IanMayo, line 69: @lilitkarapetyan - I'm testing this branch. We need this removed code - for cases other than the roles listing (the platform-types admin page)

IanMayo (dismissed):

IanMayo (approved):

ShendMikullovci:
Hi everyone! I noticed you put the bin on the far right of the role tab: 

Delete Roles Bin on the far Right

I was wondering whether that was a random decision or a developing constraint? Usually, we present the bin next to the element name and it'd be nice if we could do the same here since it's quite less visible on the far right, where the screen is a little busier with all the toggles.

IanMayo:
@lebaphi - could you investigate moving the `trash` icon next to the role name please?

lebaphi:
sure @IanMayo

PR #1431, opened 2022-03-17
1409: Insights/Game Admin Tab Notification

### LINKED ISSUE #1409: Insights/Game Admin Tab Notification Bug
Insights/Game Admin tabs don't show any notifications when new messages are received: 

https://user-images.githubusercontent.com/85491794/158014534-85516316-3f3a-4e05-b5a9-717b3204d391.mp4

It'd be nice to have a simple notification counter on both tabs that disappear once the user hits "MARK ALL AS READ".

lebaphi:
Fixes #1409 

https://user-images.githubusercontent.com/26705621/158804213-4f75dc92-2846-4eea-9909-c808ee82cb5a.mov


IanMayo:
The tab title is clever, but I think it would be enough to have a `(2)` counter on the `Insights` and `Admin` channel tabs.

Let's save the tab title number for something else.

lebaphi:
> The tab title is clever, but I think it would be enough to have a `(2)` counter on the `Insights` and `Admin` channel tabs.
> 
> Let's save the tab title number for something else.

currently only `Admin` and `Insights` tabs have the counter @IanMayo 😄

IanMayo:
Do we display the unread message count in the `Admin` and `Insights` channel tabs?  I didn't see that.

lebaphi:
> Do we display the unread message count in the `Admin` and `Insights` channel tabs? I didn't see that.

hmm, I think you might need to clear the browser cache or open the browser in incognito mode 🤔  ahh can you try to open the dev tool and reload this page?

IanMayo:
Aah, I see the `Serge (1)` browser tab title. 

image

I don't want to put the message count in there.  I want put the unread message count in the `GAME ADMIN` title:
image

For anyone that can see the `INSIGHTS` channel, I'd like the unread message count to go in there too, please.

One day in the future we can decide what to put in the browser title.

lebaphi:
Hi @IanMayo, I just did some code cleaning and updates on how we manage the `read' state of the message. Currently, I see we are setting the `read` message state in `localStorage` and we have added a lot of logic to handle that state, I think it is not necessary or we have a business, do we need to do it?

So I think we should have a new property call `hasBeenRead` in the message body and it will be stored in the database. Then, when the page is loaded, we can simply check the message status by reading the `hasBeedRead` property value. It's simpler and there's no 'potential problems'  

https://user-images.githubusercontent.com/26705621/159019277-bb7b9a91-304b-4fdb-aad0-f8866464a917.mov

IanMayo:
> has been read in the database

We can't do that @lebaphi , because we need the setting per-player.  If player-one does `mark all as read` we can't use that for the other players.  Storing it in local storage "seems" the right way of doing it.

Refactoring always welcome :-D

Your video is valid because you are logged into the same role twice.  I real life - we don't want it go to "all read" for all players.

But, the `unread message count` display is just what I wanted :-)

lebaphi:
> > has been read in the database
> 
> We can't do that @lebaphi , because we need the setting per-player. If player-one does `mark all as read` we can't use that for the other players. Storing it in local storage "seems" the right way of doing it.
> 
> Refactoring always welcome :-D
> 
> Your video is valid because you are logged into the same role twice. I real life - we don't want it go to "all read" for all players.
> 
> But, the `unread message count` display is just what I wanted :-)

ahh I see, so it's easy I think, let me update 👍

lebaphi:
How do you think @IanMayo @ShendMikullovci ?

https://user-images.githubusercontent.com/26705621/159102933-94e1aab3-f997-4fb2-ac04-854d64128594.mov

IanMayo:
Thanks @lebaphi  - but the behaviour in the rest of Serge is that on browser refresh Serge remembers which messages have been read.  I _Think_ we remember what is read in `expiredStorage` (or something like that).

I can ask Lilit to explain if you need it.

lebaphi:
Hi @IanMayo, I don't think it's correct to remember a read message in `expireStorage`, because when you open Serge in another browser or clear the browser cache, all that "read" message will be lost and the message will be treated as unread.

In the latest video, you can see the "read" message state stored in the db for specific user, so it won't change although we clear cache or open in new browser 🤔 

So if it's the right way to do, I will refactor it.

IanMayo:
Hello @lebaphi 

>  when you open Serge in another browser or clear the browser cache

Serge users don't do either of those things :-)  If a Serge user _did_ do either of those things, no actual Serge data would be lost, the user would just have to mark the messages as read again.

I'm pretty sure we already store `message-read` for all of the other Serge channels in `expired-storage`.

lebaphi:
> no actual Serge data would be lost, the user would just have to mark the messages as read again

Ahh I don't mean data will be lost, I mean the "read" state of all messages will be lost and we need to mark the messages as read again 😄 okay, so I will keep the way we are doing now to mark the message read (in `expired-storage`)

IanMayo:
Hello @lebaphi - is this ready for review?

lebaphi:
> Hello @lebaphi - is this ready for review?

not yet @IanMayo, I refactor the way we store the `read` message state so there are many files was changed and it has an issue that need to be fixed. So if you want a more safer solution, I would create another PR and continue using `expired-storage` to manage the message `read` state.

IanMayo:
Hello @lebaphi - yes, let's do it in a fresh PR please.

lebaphi:
Moved to new PR https://github.com/serge-web/serge/pull/1439

Issue #1430, opened 2022-03-17
Clean channels on delete role or force


IanMayo:
We popup a dialog when deleting a role or force.

If the Game Designer "commits" to the change then we should clean that role/force out of the channel definitions as follows:
- if a force is being deleted, delete `participations` for that force in any channel.  If there are no remaining participations, delete the channel.
- if a role is being deleted, delete that role from any participations for that force in any channel.  If there are no remaining roles for the participation, delete the participation. If there are no remaining participations, delete the channel.

IanMayo:
Fixed in #1438

PR #1438, opened 2022-03-21
1430: Clean channels on delete role or force

lilitkarapetyan:
## 🧰 Issue

#1430

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/client/src/api/wargames_api/index.ts
```diff
@@ -534,6 +534,34 @@ export const deleteForce = (dbName: string, forceName: string): Promise<Wargame>
     forces.splice(forceIndex, 1)
     updatedData.forces.forces = forces
     updatedData.channels.complete = calcComplete(forces)
+    updatedData.channels.channels.filter((v: any) => v.participants = forces)
```
IanMayo: For the filter test above, shouldn't the logical test be `===` ?

lilitkarapetyan: Here we just change assign `forces` instead `v.participiants`

IanMayo: Ah, so instead of doing a `filter`, we're actually doing a `forEach`. Right?

packages/client/src/api/wargames_api/index.ts
```diff
@@ -509,6 +511,34 @@ export const deleteForce = (dbName: string, forceName: string): Promise<Wargame>
...
+      }
+    }
+    return updateWargame({ ...res, data: updatedData }, dbName)
+  })
+}
+
+export const deleteRolesParticipations = (dbName: string, roles: Role[]): Promise<Wargame> => {
+  return getLatestWargameRevision(dbName).then((res) => {
+    const newDoc: Wargame = deepCopy(res)
+    const updatedData = newDoc.data
```
IanMayo: @lilitkarapetyan - there is a mistake in the below code.  It is setting `roles` for every force - they all get given the new set of roles.  So, if I delete a `blue` role, the new list of roles gets stored for `white` and `red`, too.

IanMayo: Steps to reproduce:
- login as admin
- open `Small wargame`
- open `forces`
- open `White` then `roles`
- view 4 roles
- now open `Blue-1` roles
- delete `CT5-B5`
- click `Save Force`
- view `White` then `roles
- now see there is just the `CO` role for `White` (there should be 4 of them)
- also see it's the same single role for `Red`, too.

IMHO, `deleteRolesParticipations` should receive a `force-id` and `role-id` - so it can remove the correct `role` from the correct `force`.

lilitkarapetyan:
> Ooh, I'll need a helping hand here @lilitkarapetyan.
> 
> We return either an updated wargame, or an updated set of roles?
> 
> Why is that?

An updated wargame is returned if roles do not exist, updated roles are returned when updated roles are migrated for live updates.

packages/client/src/ActionsAndReducers/dbWargames/wargames_ActionCreators.js
```diff
@@ -337,10 +337,8 @@ export const deleteSelectedAsset = (data) => {
 
 export const deleteSelectedRole = (dbName, data) => {
   return async (dispatch) => {
-    await data.roles.splice(data.key, 1)
-    const wargame = await wargamesApi.deleteRolesParticipations(dbName, data.roles)
-    data.handleChange(data.roles)
-    dispatch(setCurrentWargame(wargame))
+    const wargame = await wargamesApi.deleteRolesParticipations(dbName, data.roles, data.key)
+    _.isArray(wargame) ? await data.handleChange(wargame) : dispatch(setCurrentWargame(wargame))
     dispatch(addNotification('Role deleted.', 'warning'))
```
IanMayo, line 342: @lilitkarapetyan - I'm trying to tidy the logic here.  The above `deleteRolesParticipations` returns two different types of result, which gets handled in two different ways.  (`handleChange` and `setCurrentWargame`).  Can you explain why there are to different paths followed?

lilitkarapetyan: HandleChange for real-time update roles, setCurrentWargame for live update wargame. They are needed to work properly.

IanMayo: i think I need to understand this a bit more.  We only call `deleteSelectedRole` from one place. What are the two separate paths? 

I've rewritten `deleteRolesParticipations`, but it is always returning the updated wargame.

IanMayo:
Hello @lilitkarapetyan - I could do with your support, please.

I've re-written the code to delete roles (and channel participations), and given it some test coverage.

But, we now just return the updated `wargame`.  I would appreciate you checking it, please.  I'll also need some support in handing the "two different returns", please.

packages/client/src/api/wargames_api/index.ts
```diff
@@ -503,11 +505,31 @@ export const deleteForce = (dbName: string, forceId: string): Promise<Wargame> =
     const updatedData = newDoc.data
     const forces = updatedData.forces.forces
     updatedData.forces.forces = forces.filter((force: ForceData) => force.uniqid !== forceId)
-    updatedData.channels.complete = calcComplete(forces)
+    updatedData.channels.channels.forEach((v) => (v.participants as CoreParticipant[] | ForceData[]) = forces)
```
IanMayo: Can you explain what the above line is doing?

`forces` is the full list of forces in the game.

We don't put the full list of forces into the `participants` of every channel.

lilitkarapetyan: The participant assigns a value to each channel each time, at the moment it can be seen in the console that it is also transferred to the mapping
mapping participant

PR #1439, opened 2022-03-21
1409: Insights/Game Admin Tab Notification

lebaphi:
Fixess #1409

Takes over from PR #1431 

https://user-images.githubusercontent.com/26705621/159275188-82a0fa6e-2357-46e3-939f-8153d47518ef.mov


lebaphi:
Moved to new PR https://github.com/serge-web/serge/pull/1439

IanMayo:
Hello @lebaphi 
one issue.

When  we first render the page, unread messages in Game Admin are shown in bold, read messages in normal.  But, when we "mark all messages as read" - I think the they should turn from bold to normal.

This is how they are shown on refresh.

So, I think we need to re-render after `Mark all as read`, IMHO.

screencast

lebaphi:
> Hello @lebaphi 
> 
> one issue.
> 
> 
> 
> When  we first render the page, unread messages in Game Admin are shown in bold, read messages in normal.  But, when we "mark all messages as read" - I think the they should turn from bold to normal.
> 
> 
> 
> This is how they are shown on refresh.
> 
> 
> 
> So, I think we need to re-render after `Mark all as read`, IMHO.
> 
> 
> 
> screencast
> 
> 

sure @IanMayo, I will update it 👍🏻

lebaphi:
I pushed the update @IanMayo

ShendMikullovci:
Great work so far guys! Hate to be annoying but I have two things to suggest:

1. The counter should be red, same as the counters for the message channels, to maintain uniformity and attract attention.
2. If I, as GC, send a message in the general chat, I shouldn't get an unread notification, since it's my own message. Instead, I should only get a notification when someone else sends it.

IanMayo:
> 1. The counter should be red, same as the counters for the message channels, to maintain uniformity and attract attention.

Yes, well spotted.

> 2. If I, as GC, send a message in the general chat, I shouldn't get an unread notification, since it's my own message. Instead, I should only get a notification when someone else sends it.

_Hopefully_ we can do that :-)

ShendMikullovci:
What do you guys think about refreshing our notification counter? With all the new features we've added, I think a bell icon would fit nicely on the screen. Here's what I came up with: 

Bell Icon Figma

Bell Icon Example

IanMayo:
Very stylish @ShendMikullovci, but...

Personally I find the font-size an issue.  We also have an issue where users have so many tabs that they fall onto the _overflow_ menu.  I think we should keep the tabs titles as narrow as possible.

IanMayo:
Ok @lebaphi - the only thing I think we have to do on here is to make the message-count red.

lebaphi:
> Ok @lebaphi - the only thing I think we have to do on here is to make the message-count red.

yes sir 😁

lebaphi:
Updated @IanMayo @ShendMikullovci 

image

packages/client/src/Views/MessagesListRenderProp.jsx
```diff
@@ -1,71 +1,12 @@
...
-    }
-  }
-
-  render () {
-    return (
-      <div className="message-list">
-        {this.props.render(this.state.messages)}
-      </div>
-    )
-  }
```
IanMayo, line 68: @lebaphi - this PR deletes this large bloc of code.   That's good news to me :-)

But, has this been replaced by new code?  If it has, I can't see it.

ShendMikullovci:
> Updated @IanMayo @ShendMikullovci
> 
> image

Just reviewed. Looks great!

packages/client/src/Views/MessagesListRenderProp.jsx
lebaphi: Hi @IanMayo, this block of code use to iterate over array of `messages` and recalculate `hasBeenRead` for each message. I think it's unnecessary because we already have a function `isMessageReaded` to do it. You can have a look at `MessagesListInsightsChannel.jsx`, code line 27.

IanMayo: Great, thanks @lebaphi :-)

IanMayo (approved):

ShendMikullovci:
I was looking at Cloud-One and noticed we've got a "NEW" counter next to "Game Admin". Intentional? If so, when does it appear?:

Cloud One Notification Counter

lebaphi:
> I was looking at Cloud-One and noticed we've got a "NEW" counter next to "Game Admin". Intentional? If so, when does it appear?:
> 
> Cloud One Notification Counter

can you help me to reproduce it? @ShendMikullovci

ShendMikullovci:
Of course @lebaphi! It seems the counter shows "NEW" once the message counter is higher than 8:

Test Message 8

At 8 messages, we get "(8)"

Test Message 9 NEW

Test Message 12 NEW

At 9 messages and higher, we get "NEW".

IanMayo:
Aah, it should be `9+` if we have more than `9`, please.

ShendMikullovci:
Just reviewed this. Works well. Thanks, guys!

9+ fixed

Issue #1433, opened 2022-03-17
"Blue Screen of Death" on Newly Initiated Wargame


ShendMikullovci:
I logged in to find out the initial 1400 wargame that I had created, saved and initiated had been deleted. So I created a new one, but encountered the blue screen after launching and trying to log in as a user:

![Blue Screen of Death](https://user-images.githubusercontent.com/85491794/158904737-81233686-c27b-4921-bd81-c3aaee0e7ffd.png)

IanMayo:
Hey, @lebaphi  - could you give Shend the steps to get the console trace please?

IanMayo:
It's really valuable to catch bugs like that @ShendMikullovci 

Based on the `serge-review` version, could you give us the exact steps to reproduce this please?

lebaphi:
> Hey, @lebaphi - could you give Shend the steps to get the console trace please?

I think you just need to open the dev tool or press F12 button, open the `Console` tab and the logs was there

ShendMikullovci:
> It's really valuable to catch bugs like that @ShendMikullovci
> 
> Based on the `serge-review` version, could you give us the exact steps to reproduce this please?

So, it turns out the bug doesn't happen when I log in through White Force. However, as soon as I add other forces, I can't access the game.

ShendMikullovci:
> I think you just need to open the dev tool or press F12 button, open the `Console` tab and the logs was there

This is what I got: 

![Dev Tool Errors](https://user-images.githubusercontent.com/85491794/159073947-16d8a709-b5b8-495a-9ff2-e3df43314b3b.png)

IanMayo:
Hello @ShendMikullovci  - I've implemented a workaround for this bug.

The underlying bug is that a problem with `Save channel` meant that we ended up with multiple channels with the same `id`.  Serge was falling over when trying to display multiple channels with the same id.  So now, when Serge loads, it removes channels with a duplicate id.

Clearly the correct fix is to stop saving duplicate channels, but my workaround (now in `develop` branch, which all other branches come from) - so you shouldn't see it again.

ShendMikullovci:
It appears that now I can't access the game at all: 

https://user-images.githubusercontent.com/85491794/159704403-99f92282-2531-4b99-af99-1b05f6401c41.mp4

Neither the admin page nor the player page allow me to log in.

IanMayo:
Sorry @ShendMikullovci - I'm on it.

ShendMikullovci:
[...]

PR #1445, opened 2022-03-23
Investigate Serge-Review crash

IanMayo:
Introduce some diagnostics to learn what's making live deployment crash.

Supports #1433


IanMayo:
Hmm, not failing in Heroku for this branch. Try merging it, so we can test in `serge-review`.

PR #1447, opened 2022-03-23
Allow players from multiple timezones

### LINKED ISSUE #1321: As a Game Controller I need to support players in different time-zones
My users are planning a distributed wargame in the Spring.  They are hoping to use this wargame to "prove" a number of capabilities, and test that Serge can support a distributed wargame.

So far, all wargames have been run in the same timezone - and the local time has been stored in the database.

For players to play a Serge wargame in different timezones, I believe we need these changes:
* dates & times stored in database have to be in UTC
* the UI must translate dates/times to local timezone before they are rendered (this includes the "One minute ago" labels).

Note: we may have to change the date-time values in the sample wargames.  If there aren't unit tests that break, then we should introduce unit tests to verify the above functionality.

IanMayo:
Fixes #1321 

When we create a date, we ensure it is always created from UTC.  When we display a date, we ensure it is always presented as UTC.

The approach taken in this PR has been to move the data-handling into two helper functions (`newDate()` and `formatDate()`). Within these functions we can centrally control the timezone related aspects.


IanMayo:
Hello @lilitkarapetyan  & @lebaphi - feedback on what I've done here is quite welcome :-)

IanMayo:
Aah, @lebaphi  and @lilitkarapetyan - I could do with you helping me with checking the timezones, please.

I'm keen to check that all timezones see the same values for time.  In that way, one player can refer to "the message sent at 16:30" - and someone in another timezone sees the correct time.

So, could you please go into the Heroku deployment for this PR, then go into the game titled `Cloudant-Test` as a `Blue` player, and see if you get the message timestamps:

The Blue-Chat message should show as being sent at `17:07` and the most recent Admin-Chat should be at `14:04` - as shown below.

image

Issue #1413, opened 2022-03-13
Channel Creation & Deletion Bug


ShendMikullovci:
When I create a new channel and save it, a new one is instantly created without me clicking on the button.
Furthermore, deleting the newly unsolicited channel actually ends up deleting the one I intentionally created. 

https://user-images.githubusercontent.com/85491794/158082311-8ee3c691-9346-4763-b993-5e18033cbbcb.mp4

ShendMikullovci:
It turns out the unsolicited channels are linked to the intentional ones. Check out how they're both selected simultaneously: 

![Linked Mapping Channel](https://user-images.githubusercontent.com/85491794/158082418-5729ba04-094e-451e-b5be-4c4e32b90e41.png)

![Linked Channel](https://user-images.githubusercontent.com/85491794/158082422-cc93e7ac-b99a-4833-a645-48cb08b5f3c4.png)

IanMayo:
One for you, please @lilitkarapetyan  - this looks like a recurring issue.

ShendMikullovci:
Another deletion bug I came accross is that the channel I hit delete on, doesn't get deleted. Rather, the 1st created one does, and the deletion proceeds in that order: 

https://user-images.githubusercontent.com/85491794/159074636-7884010b-5671-4ff9-a3d8-94a911981bc2.mp4

IanMayo:
This is a proper bug, where duplicate channels (with the same id) are created.  That also explains why the "wrong" channel is deleted.

ShendMikullovci:
This has been fixed.

PR #1448, opened 2022-03-24
Delete channels (even with duplicates)

IanMayo:
Supports #1413

The above issue involves multiple channels being created with the same id.  Unfortunately, it is unpredictable which channel will be deleted.

This doesn't solve duplicate channels getting created.  But, it does ensure all channels sharing the same id get deleted.


ShendMikullovci:
A duplicate is still created in this deployment as you said, and deleting one deletes both.

Issue #1446, opened 2022-03-23
Cloudant instance not pinging for updates


IanMayo:
I'm running `Serge` in `Serge-Review` heroku branch.

But, I think the `listen for changes` functionality isn't working.

The Network tab shows these failed calls:
![image](https://user-images.githubusercontent.com/1108513/159730450-a0e177fd-840a-4193-9820-0c9d1e6ea421.png)

Expanding a call gives this:
![image](https://user-images.githubusercontent.com/1108513/159730530-c771b6ed-e90f-418f-9f3c-fa8dc21420eb.png)

It _looks_ like we're trying to call `localhost:4000` - when we should actually be calling `serge-review` (I think).

IanMayo:
We've dropped cloudant.

PR #1455, opened 2022-03-25
1446: Cloudant instance not pinging for updates

lilitkarapetyan:
## 🧰 Issue

#1446

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/config/src/globals.ts
```diff
@@ -128,7 +130,7 @@ export const hiddenPrefix = '_#_'
 
 export const clearAll = 'clearAll'
 export const allDbs = 'allDbs'
-export const socketPath = 'http://localhost:4000'
+export const socketPath = hostname.toLowerCase().indexOf('herokuapp') !== -1 ? `https://${hostname}` : 'http://localhost:4000'
```
IanMayo: I think this socketPath may be wrong.

If we're not on `Heroku`, the browser may be connecting to server other than `localhost`.  So, on a network without an Internet connection one PC will be hosting Serge, and other network PCs will connect to it.

So, I think the hostname needs to come from the `window.location` 

`localhost` only works for us devs.

IanMayo:
Hmm, the `View deployment` for this PR is broken.  I don't know if it's:
- node at Heroku
- client-side Serge
- IBM Cloudant
Here is the server log:
image

IanMayo (approved):

PR #1474, opened 2022-03-31
1364 Support Turning Circle

### LINKED ISSUE #1364: Support turning circles in planned routes
For close-in wargames, the turning circles of ships becomes a significant factor.

* [x] Introduce data-type for turning circle in platform-type
* [x] If turning circle present for platform use "unwind-rope" algorithm to determine extent of achievable positions
* [x] If extent of achievable positions is present, restrict allowable hexes to the current one, plus any cells with a centre in the set of achievable positions
* [x] When player is planning route, exclude those not in the allowable hexes
* [x] When marker is dropped (and turn has been planned), see if it matches the direct route.  If it does, just store ends, otherwise store all cells
* [ ] Respect terrain data - don't allow ships on land, or land assets on sea
* [x] For partial leg plans, ensure correct amount of distance is removed (reflect complex path if present)
* [x] Don't show blue line if it passes through no-go area, and don't process turn if dropped in no-go area.
* [ ] Handle issue when immediately dropping pin into no-go area (see video below)

scrreencast

lebaphi:
Fixes #1364


IanMayo (approved):

ShendMikullovci:
Just reviewed this. Looks great! I noticed I forgot to put an info marker in the design. Do you guys think we should have it for both units of measurement icons?

Info Marker Turning Circles

IanMayo:
I think we're ok.  For the lower one, it may start with an empty value - so we need to indicate what it represents. For the upper one, the `m` is permanently there.

PR #1484, opened 2022-04-06
1472: Forces List Reacts to Collapsible Elements

### LINKED ISSUE #1472: Forces List Reacts to Collapsible Elements
As you can see below, the collapsible elements shrink the list width on the left. 

https://user-images.githubusercontent.com/85491794/160870278-1bb08dd5-0ed7-43dd-86b9-d049cbe3670c.mp4

I suppose this is because of the scrollbar on the right-hand side being introduced by the collapsible elements. 
Fairly low-priority, but it'd be smoother to have the screen stay still while the user is interacting with the elements.

lebaphi:
Fixes #1472


IanMayo:
@ShendMikullovci - you reported this issue originally.  Would you mind checking it's fixed in your browser?

(note: the `serge-review` instance was broken this morning. @lilitkarapetyan is working on a fix )

ShendMikullovci:
Hi guys! It seems that the screen is reacting now instead of the forces list:

https://user-images.githubusercontent.com/85491794/162227462-0f3876be-cd38-4ccc-861f-5f2b3c332ca5.mp4

Given that @phileba pointed out that this doesn't happen on MAC and that such a long name for a force would be unusual, I'd be ok with setting this aside for now.

IanMayo (approved):

Issue #1478, opened 2022-04-05
Move away from force-name in perception management


IanMayo:
In perception management, we use force-name a lot.

We need to move away from that, since the actual force-name may change.  Instead we have to use force-id.

Large parts of this are in:
- findPerceivedAsTypes
- visibility and perception form
- visible to in adjudication form
- perceptions in JSON wargame definitions

IanMayo:
Fixed, in #1485

PR #1485, opened 2022-04-06
1478 perception by force

IanMayo:
Supports #1478 

I've put quite a lot of work into this PR.  But, now I think we need to reconsider how we manage perceptions, including:
- `<AssetIcon>` should not use css styles for the icon color & image, the values of color and icon should come from the wargame document
- perception data should be passed via the color shade and icon URL

Once the above changes are made, the changes for perception should probably be started from fresh


IanMayo:
Aah, the SVG path works on localhost, but not on heroku.

Localhost:
image

Heoku:
image

PR #1495, opened 2022-04-08
1490 asset icon

### LINKED ISSUE #1490: Create icons using props not css-class
The icon and color for `AssetIcon` components are largely formatted using CSS-Styles.

This means they must be hard-coded within the game source code. But, the Game Designer is actually meant to create them in the Game Admin pages.

So, AssetIcon should format the icon using props, not css-styles.

The perception-management code should change, to allow this. Specifically, it should move away from providing CSS classes, instead provide actual color hexes and icon URLs.

IanMayo:
Fixes #1490


packages/components/src/local/form-elements/title-with-icon/index.tsx
```diff
@@ -8,8 +8,11 @@ import PropTypes from './types/props'
 import styles from './styles.module.scss'
 
 /* Render component */
-export const TitleWithIcon: React.FC<PropTypes> = ({ children, forceColor, platformType }) => {
+export const TitleWithIcon: React.FC<PropTypes> = ({ children, forceColor, platformType, icon }) => {
   const iconClassname = getIconClassname(forceColor.toLowerCase(), platformType)
+  console.log('title with icon', forceColor, platformType, icon, iconClassname)
+  // TODO: switch from using `iconClassname` to using the `icon` field in <GetIcon
```
IanMayo: Hello @lebaphi - this is where I have got stuck. This component currently uses CSS-classes to specify the icon svg path. But, we have to move to taking that from the data.  I've collated the data here, but can't get the icon to show.

It can be tested from the `storybook` for the component.

lebaphi: Hi @IanMayo, I will check it 😉

lebaphi:
Hi @IanMayo, when I tried 
```
return (
    <div className={styles.main}>
      <h2>{children}</h2>
      <div className={styles.icon}>
        {/* <div className={iconClassname} /> */}
        <GetIcon icType={platformType} color={forceColor} isSelected={false} imageSrc={icon} />
      </div>
    </div>
  )
 ```
The icon seems to load properly, is it right? 🤔 
image

**Comment by IanMayo** [comment]
The `<TitleWithIcon>` component is used with the `plan turn form`.

I'm playing with this game/role: `/?wargame=wargame-kzod28lc&access=rks5zfzd6`

Here is when I click on the right-hand ship: 
image

If I switch to `<GetIcon>`, I see no icon at all:
image

**Comment by IanMayo** [comment]
Ok @lebaphi - I've removed the instances of using CSS for the icon type.  

As you'll see from [Storybook](https://serge-1490-asset-icon-2qkdx6gs.herokuapp.com/storybook/?path=/story/local-mapping-smallscale--natural-earth) lots of icons are now missing.

image


I _guess_ the first fix is for `<GetIcon/>`, then `<TitleWithIcon>`.

File: .husky/pre-push [hunk 2]
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
```
IanMayo: @lebaphi - I don't think these next lines should have been commented out

packages/components/package.json
```diff
@@ -6,15 +6,16 @@
...
     "watch": "run-p watch:*",
     "watch:ts": "tsc --watch",
     "watch:style": "node-sass-chokidar ./src -o ./build --watch --ext scss",
     "copy": "copyfiles -u 1 -e src/**/*.ts* -e src/**/*.md src/**/*.* build",
+    "copy-assets": "copyfiles -u 4 src/local/asset-icon/counters/*.svg ../client/public/assets/counters",
     "cleanup": "rimraf build/vendor && rimraf build/local/**/mocks && rimraf build/local/**/*.stories.* && rimraf build/local/**/*.spec.*",
     "test": "run-p test:*",
     "test:ts": "tsc --pretty",
-    "test:unit": "jest --passWithNoTests",
+    "test:unit": "jest --passWithNoTests -u",
```
IanMayo: No, we shouldn't be appending `-u` here. Jest is there to show us when something has changed.  Having `-u` will _hide_ any changes.

lebaphi: it does not hide any changes @IanMayo, it just automatically updates the snapshot with new changes and we can view it in the git 😄

IanMayo:
I prefer the behaviour when push fails if a snapshot doesn't match.  We had instances where UI changes _slipped-through_ because the snapshot silently got updated.

lebaphi:
> I prefer the behaviour when push fails if a snapshot doesn't match. We had instances where UI changes _slipped-through_ because the snapshot silently got updated.

okay, I will remove it 👍

IanMayo:
You've done some great work here @lebaphi 

`storybook` is still broken though.  It's trying to get SVG icons from these locations:
image

`develop` gets them from these locations:
image

lebaphi:
> You've done some great work here @lebaphi
> 
> `storybook` is still broken though. It's trying to get SVG icons from these locations: image
> 
> `develop` gets them from these locations: image

strange, storybook used the old url, it should work as it should be 🤔

IanMayo:
> strange, storybook used the old url, it should work as it should be 🤔

will do fresh build and restart, just in case.

No, @lebaphi - I still get the same path.

lebaphi:
> /?wargame=wargame-kzod28lc&access=rks5zfzd6

I've checked it here, it we run storybook, it should use the original path
```
 if (process.env.STORYBOOK) {
      return `/static/media/src/local/asset-icon/counters/${url}`
    }
    return `/assets/counters/${url}`
```

IanMayo:
Thanks @lebaphi  - good fix. The copied svgs don't need to be included in the commit, do they?

lebaphi:
> Thanks @lebaphi - good fix. The copied svgs don't need to be included in the commit, do they?

yes it's up to you @IanMayo ,  I see we already have a `static` folder so committing these svgs into this is fine for me

IanMayo:
Can we just have the one version in `asset-icon/counters` please?

We can sort out the version that `info-marker` requires in the future.

lebaphi:
> Can we just have the one version in `asset-icon/counters` please?
> 
> We can sort out the version that `info-marker` requires in the future.

The svgs should be in the `public` (develop) or in the `static` (storybook) then we can use `fetch', otherwise we can only use css to import these svgs.

IanMayo:
> The svgs should be in the `public` (develop) or in the `static` (storybook) then we can use `fetch', otherwise we can only use css to import these svgs.

Yes, but `yarn` copies them into those two places.   So, we don't need them checked into the repo, they are `build artefacts`.

lebaphi:
> > The svgs should be in the `public` (develop) or in the `static` (storybook) then we can use `fetch', otherwise we can only use css to import these svgs.
> 
> Yes, but `yarn` copies them into those two places. So, we don't need them checked into the repo, they are `build artefacts`.

hmm, I also thought there should be a folder to store assets, but the problem is that assets can only be stored in public or static, if we save assets in `public` and start storybook (which loads assets in `static`), assets will not load properly.

IanMayo:
> hmm, I also thought there should be a folder to store assets, but the problem is that assets can only be stored in public or static, if we save assets in public and start storybook (which loads assets in static), assets will not load properly.

We can only run `yarn storybook` if we have already run `yarn build`. `yarn build` copies the files to the two public locations.   That means we don't need to commit need to commit the copies in the public locations.  Right?

lebaphi:
> > hmm, I also thought there should be a folder to store assets, but the problem is that assets can only be stored in public or static, if we save assets in public and start storybook (which loads assets in static), assets will not load properly.
> 
> We can only run `yarn storybook` if we have already run `yarn build`. `yarn build` copies the files to the two public locations. That means we don't need to commit need to commit the copies in the public locations. Right?

yep, we don't need to commit these svgs in 2 places, run `yarn build` will do it for us 😄 I will remove it

IanMayo:
Great, thanks @lebaphi :-)

IanMayo:
Oops, one last thought @lebaphi   We have `<GetIcon>` defined in `asset-icon/index.tsx`.  But, we don't use it in `asset-icon/index.tsx`.  Do you think we can modify `asset-icon/index.tsx` so it uses `<GetIcon>` ?

Sorry, @lebaphi - I'm growing the requirement here :-(

Could we do some refactoring to make `<GetIcon>` testable? 

Hold on, not now. Let's get it merged. I'll create a separate issue.

lebaphi:
great @IanMayo 👍🏻

PR #1513, opened 2022-04-15
1477: Display of icons broken in world map panel

### LINKED ISSUE #1477: Display of icons broken in `world map` panel.
image

When we added formatting for "view force assets" in Game Admin page (below), it broken the layout in the World Map panel, with all the text pushed to the right (above).

image

I think both panels use the `asset-icon` component - but we need two different ways of styling it - depending on which version it is.

lebaphi:
Fixes #1477 

Screen Shot 2022-04-15 at 6 19 28 PM


IanMayo (approved):
Fixed at last :-)

PR #1515, opened 2022-04-17
1503 Refactor Asset Icon

### LINKED ISSUE #1503: Refactor `<AssetIcon>`
IMHO, our icon implementation is a bit clunky.

Specifically, `<AssetIcon>` includes the `<GetIcon>` component, but it doesn't use it.

I'd like us to refactor `<GetIcon>` into it's own component, so we can give it an `storybook` test that allows `destroyed`, `selected`, `force` and `type` to be modified in the storybook control.

So, I _think_ we should:
1. Rename `AssetIcon` to `MapIcon` (since it is used for putting an icon on the map)
2. Rename `GetIcon` to `AssetIcon`
3. Modify _new_ `MapIcon` to use _new_ `AssetIcon`
4. Refactor `AssetIcon` to it's own component
5. Produce `story` and `spec` for `AssetIcon`

I hope all that renaming isn't too complicated 🤣

lebaphi:
Fixes #1503


IanMayo:
The file called `marker` appears to be a `.png` - can we give it the correct file-type please.

Can you explain the role of this file please @lebaphi ?   I can't see how it is used in `asset-icon`.

IanMayo (approved):
Great refactoring - code much neater :-)

PR #1521, opened 2022-04-21
1516: New asset styling, for info markers.

### LINKED ISSUE #1516: New asset styling, for info markers.
We are going to have one "pool" of SVG markers.  But, we need to give different styling to `map icons` and `info markers`.  Here the left-hand question mark is an `info marker`.

image

I propose we introduce a new styling for info markers:
- force-shade colors, as present
- white filled circle, with outer circle and icon SVG in force color

Here are mockups:
image

I only expect to see changes to the `<InfoMarker>` class.

lebaphi:
Supports #1516 

https://user-images.githubusercontent.com/26705621/164446170-7f5439b8-42b5-4fb7-ae86-277af3f29df4.mp4


IanMayo:
There is some trickery going on here @lebaphi .  I'd like to see if we can avoid it,

These info markers are different to the asset markers. We can make fresh requirements, if we want to.  What if our requirements are that they don't have a fill color?  Then I _guess_ the browser uses the `foreColor` when rendering.

IanMayo (approved):
I'm going to merge this, so I can use it in a demo. But, I remain interested in a finding a solution that doesn't require the DOM manipulation.

lebaphi:
yeah, I will find another better solution 👍🏻

IanMayo:
Thanks :-)

But, as I said - I'm happy to consider changes to the image strategy, if that helps.

PR #1525, opened 2022-04-26
more content checking

IanMayo:
Our new SVG code is breaking:
image

This workaround gets the code working, but I don't know what the underlying problem is.

@lebaphi  - do you see what is going wrong?


packages/components/src/local/info-marker/index.tsx
```diff
@@ -44,11 +44,14 @@ export const InfoMarker: React.FC<PropTypes> = ({
...
         const svgElm = document.getElementById(imageSrc)
         if (svgElm) {
-          const svgElms = Array.from((svgElm as any).contentDocument.getElementsByTagName('svg')) as HTMLElement[]
-          if (svgElms.length) {
-            const svgStyleElms = Array.from(svgElms[0].getElementsByTagName('style')) as HTMLElement[]
-            if (svgStyleElms.length) {
-              svgStyleElms[0].innerHTML = `.st0{fill:${color};}`
+          const asAny = svgElm as any
+          if (asAny.contentDocument) {
+            const svgElms = Array.from((svgElm as any).contentDocument.getElementsByTagName('svg')) as HTMLElement[]
```
lebaphi: Hi @IanMayo , as the error shows in the console log, I think the `svgElm.contentDocument` is null, so it should be `asAny.contentDocument.getElementsByTagName (...)` because you have checked `if (asAny.contentDocument){...}` 😄

IanMayo:
Close. The changes related to deleted code,

PR #1526, opened 2022-04-26
1259 Admin Page Step Cursor

### LINKED ISSUE #1259: Admin Page Step Cursor
There's a cursor on the top right-hand side of the screen on the admin page that reacts to the user switching the pages:
Step Cursor Admin Page

The "Platform Types" page is not included in the cursor. 

I also noticed that the "Forces" page is shown as a green icon on the cursor. Is there an explicit reason?

I suggest the green icon moves steps along with the pages as the user skips ahead.

lebaphi:
Fixes #1259


lebaphi:
removed @IanMayo @ShendMikullovci 👍

IanMayo (dismissed):
Have removed the component for the progress indicator. Have also removed a second progress indicator component that was unused.

IanMayo (approved):

Issue #1529, opened 2022-04-26
Refactor how we save channel & force


IanMayo:
Our save-channel/save-force logic is complex because it uses `name` to index the channels/forces, but the `name` could change.

Since that was written, we now index channels. & forces by `uniqid` - so the `saveChannel`\`saveForce` code in `wargames_api` can be simplified.

![image](https://user-images.githubusercontent.com/1108513/165298879-9b85dc46-cb8e-4d3b-a7e3-dde6a92b3852.png)

IanMayo:
Fixed in #1536

PR #1536, opened 2022-04-27
1529 Refactor how we save channel & force

lilitkarapetyan:
## 🧰 Issue

#1529

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/custom-types/force.d.ts
```diff
@@ -6,7 +6,7 @@ export default interface Force {
   // TODO: we need to constrain what we store below.
   // I _think_ we're storing a whole force. I'm sure it should be
   // a force-id
-  selectedForce: string,
+  selectedForce: string | { uniqid : string },
```
IanMayo: How about we specify `string | ForceData | undefined`

To be honest, I don't like the `|` union.  How about we remove usage of strings in there, and just use:
```
  selectedForce?: ForceData
```

Aah, but we **certainly don't** want to store the whole `ForceData` object when we persist to database.

packages/server/providers/couchdb.js
```diff
@@ -90,7 +90,15 @@ const couchDb = (app, io, pouchOptions) => {
         if (err.status === 409) {
           return retryUntilWritten(db, doc)
         } else { // new doc
-          return db.put(doc).then(() => res.send({ msg: 'Saved', data: doc }))
```
IanMayo, line 93: Is this fixing a bug @lilitkarapetyan ?

packages/server/providers/couchdb.js
```diff
@@ -161,7 +169,7 @@ const couchDb = (app, io, pouchOptions) => {
     remoteDb.allDocs({ include_docs: true, attachments: true })
       .then(result => {
         const messages = result.rows.reduce((messages, { doc }) => {
-          const isNotSystem = doc._id !== wargameSettings
+          const isNotSystem = doc._id !== wargameSettings || doc._id !== settings
```
IanMayo: Hmm, I think the above needs to be `&&`. Both need to be false to be sure it's not the system doc

packages/server/providers/couchdb.js
lilitkarapetyan: Yes, this part is for updating "initial_wargame" and I added "settings" if "initial_wargame" doesn't exist.

IanMayo (approved):

PR #1554, opened 2022-05-12
1553: Fix scss changed does not recompile src code

lebaphi:
Fixes #1153 

So my solution is:
- Clone `node-sass-chokidar` from https://github.com/michaelwayman/node-sass-chokidar/tree/release
- Edit `package.json`, increase version of `node-sass` to latest
- Using pkg to package this project to executable file then we can use as the way we want.

Tested on MacOs

https://user-images.githubusercontent.com/26705621/168073061-3c8fd844-8cfd-459e-9bc7-569d57679b74.mov


IanMayo:
Thanks for this solution @lebaphi 

If I can, I'd like to avoid a binary dependency.  Have you checked if `dart-scss` would meet our requirements?
https://sass-lang.com/documentation/cli/dart-sass

package.json
```diff
@@ -36,7 +37,7 @@
     "lint": "lerna run lint --stream",
     "storybook": "lerna run storybook --scope @serge/components --stream",
     "build-storybook": "lerna run build-storybook --scope @serge/components --stream",
-    "build": "lerna run build --scope @serge/components --scope @serge/config --scope @serge/helpers --scope @serge/store --scope @serge/client --stream",
+    "build": "yarn sass-watcher & lerna run build --scope @serge/components --scope @serge/config --scope @serge/helpers --scope @serge/store --scope @serge/client --stream",
```
IanMayo: @lebaphi - I think `build` just does a single build.  I thought that `develop` did the _watch_ on the codebase.

lebaphi: @IanMayo it runs the script `sass-watcher` to copy the binary to `node_modules/.bin/`, does not watch :D

lebaphi:
> Thanks for this solution @lebaphi
> 
> If I can, I'd like to avoid a binary dependency. Have you checked if `dart-scss` would meet our requirements? https://sass-lang.com/documentation/cli/dart-sass

yep I've checked but got no luck, I also thought current solution is not good enough because the binary size is too big. Need to reduce the file size or find another better

lebaphi:
@IanMayo I think the last commit is the best solution 😄

IanMayo (approved):
This works properly. I've checked by updating some CSS in `yarn develop`

PR #1575, opened 2022-05-27
1573: Display modal warning when data incompatible

### LINKED ISSUE #1573: Display modal warning when data incompatible.
This issue relates to this story:
http://localhost:6611/?path=/story/local-organisms-settingchannels--default

image

When editing a mapping channel, there are some data constraints.  If the "participant" (row) specifies that it is controlling one or more assets, then we need exactly one row in the "Role" column.  It can't be zero - since that wouldn't tell us who is controlling the asset. It can't be more than one, since we wouldn't know who was in control.

I've done the testing for this.  But, if the test fails, I need us to show a modal dialog, presenting the warning to the user.

It's the `todo:` item here:
image

lebaphi:
Fixes #1573 

https://user-images.githubusercontent.com/26705621/170651432-48c18214-688b-4148-adf4-358f2bebff12.mov


IanMayo:
Thanks @lebaphi - I'm away from my desk right now, but I think we've missed a path.

I think we need a `setProblems('')` if there are no problems.

IanMayo:
Ok, one issue.  The code is working fine, but could you make it so that it works in `storybook`?

IanMayo (approved):
great, thanks.

PR #1576, opened 2022-05-27
1574: Neater display of multiple items in EditableRow

### LINKED ISSUE #1574: Neater display of multiple items in EditableRow
When editing participants in a mapping channel the list presentation can be difficult to read:

image

I think we should add a `presentAsList` boolean prop for `EditableRow`. When it's set, the items should be displayed as an un-ordered list, but without bullets.

lebaphi:
Fixes #1574 

Screen Shot 2022-05-27 at 9 04 33 PM


IanMayo (dismissed):
Working fine.

PR #1590, opened 2022-06-07
1589 save asset

### LINKED ISSUE #1589: Can't always SAVE asset name changes
I'm playing around with the `Forces` page in `develop` branch.

On most occasions, changing an asset name does not get stored in the database.  It keeps the old name.

My _guess_ is that `setSelectedItem` isn't looking at the "real" object, but is looking at a `copy`.

screencast

IanMayo:
Fixes #1589 

The `Save` event was pushing a cloned copy of the `forces` object into the database, then using this cloned object in the client state.

But, the `selected asset` was an object from the `original` force.  So, when we store changes to that asset, it doesn't change the original.

Now, when a new `force` is received, we update the `selected item` to point to a valid item.


IanMayo:
@lilitkarapetyan  - could you check this fix is ok please?

Note: I'm pretty sure its still broken.  I'm pretty sure the problem is when we clone the asset:

image

Could you investigate, please @lilitkarapetyan ?

PR #1595, opened 2022-06-14
Windows & Linux executables failing

### LINKED ISSUE #1592: Windows & Linux executables failing
I've run `yarn pkg-build` on the `develop` branch.  I then copied the Windows executable to a `Windows 11` Virtual machine.

When I run the executable from `Windows Explorer` it silently fails.

When I run it from Command Prompt, it fails with this error:
image

The users are planning on running a wargame next week, and are keen on getting a test-able version on their network.

Ooops, it's also failing on Linuex:
image

Note: I also downloaded a CI build from this PR, and it failed with the same error:
https://github.com/serge-web/serge/actions/runs/2455919188

Ok. I ran the CI build on an old Windows 10 PC, and got this error. It gives more detail:
image

lilitkarapetyan:
## 🧰 Issue

Fixes #1592 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Thanks @lilitkarapetyan - I was expecting you to have to stick in the '.node' files for Marcos/Linux too.  I "think" this PR will only provide executables for windows 

Here is the lib file for macos:
[node_sqlite3.node_macos.zip](https://github.com/serge-web/serge/files/8897672/node_sqlite3.node_macos.zip)

Here is the lib file for ubuntu: 
[node_sqlite3.node_linux.zip](https://github.com/serge-web/serge/files/8898409/node_sqlite3.node_linux.zip)

Here is the executable running on linux:
image

packages/executable/package.json
```diff
@@ -13,6 +13,7 @@
   "bin": "../server/bin.js",
   "pkg": {
     "assets": [
+      "../../node_modules/sqlite3/lib/binding/**/*",
```
IanMayo: I don't think this is right @lilitkarapetyan 

When this runs, the `sqlite3` folder will only contain the lib for the current operating system.

image

But, we want `yarn pkg-build` to create executables for all 3 operating systems.  So, I think we have to manually put the 3 compiled `.node` files into the `executable\sqlite3.nodes` folder.

IanMayo:
I've checked out the branch on macos, run `yarn install`, then `yarn build`, then `yarn pkg-build`.   But, when I copy the windows executable to ms-windows, it fails like this:
image

lebaphi:
@IanMayo This one works for macos arm64 architecture 👍 
We should create a new folder under `packages/executable/sqlite3.nodes` for arm64, then update the `packages/executable/helpers/pkhCopyCrossPlatforms.js` too.

[node_sqlite3.node.macos_arm64zip.zip](https://github.com/serge-web/serge/files/8899684/node_sqlite3.node.macos_arm64zip.zip)

IanMayo:
> @IanMayo This one works for macos arm64 architecture 👍 We should create a new folder under `packages/executable/sqlite3.nodes` for arm64, then update the `packages/executable/helpers/pkhCopyCrossPlatforms.js` too.

Could you make that change, in this PR please @lebaphi ?  That's because @lilitkarapetyan wouldn't be able to test that `arm64` is working correctly.

lebaphi:
sure @IanMayo

packages/executable/helpers/pkhCopyCrossPlatforms.js
```diff
@@ -98,8 +105,10 @@ Promise.all(promises).then(() => {
     let outputPath = `${zipPath}/SERGE_${getDate()}`
     if (dir.indexOf('linux') !== -1) {
       outputPath += '_linux.zip'
-    } else if (dir.indexOf('macos') !== -1) {
+    } else if (dir.indexOf('macos') !== -1 && dir.indexOf('macos-arm64') === -1) {
```
IanMayo: @lebaphi  - the logic can be simplified if you change the order a little.

If you test for arm64 first, you just need to test for `macos` in the following test

IanMayo:
Hello @lebaphi , one more change, please.
image

We create two files called `executable-macos`. Could the first one be called `executable-macos-arm64` please?

IanMayo:
Hello @lilitkarapetyan - I get this error when I run the current built executable on MS-Windows:
image

Could you please check if it works for you?

lilitkarapetyan:
I have win64 installed and it works without errors.

IanMayo:
> I have win64 installed and it works without errors.

Ok. It ran fine on a Windows VM on Intel Macos.  It failed on Windows VM on ARM Macos.

lilitkarapetyan:
> 
Hi Ian,
Do you use win32 or win64 on  ARM Macos?

IanMayo:
> Do you use win32 or win64 on ARM Macos?

Aah, it _looks like_ the problem could be that it's 64-Bit ARM version of windows :-(

image

That's certainly not a target we formally need to support.

IanMayo:
I am probably worrying over nothing, but.

In this line I don't see a target for `macos-arm64`.  Can @lilitkarapetyan or @lebaphi explain why it isn't there?

image

One more thing, for @lebaphi:
Could you change the build description so it's clear which is `executable-macos` and which is `executable-macos-arm64`?  I did try to do this myself, but broke the build.  The reason for this, is that I want it to be quite clear where errors occur.
image

lebaphi:
> I am probably worrying over nothing, but.
> 
> In this line I don't see a target for `macos-arm64`. Can @lilitkarapetyan or @lebaphi explain why it isn't there?
> 
> image
> 
> One more thing, for @lebaphi: Could you change the build description so it's clear which is `executable-macos` and which is `executable-macos-arm64`? I did try to do this myself, but broke the build. The reason for this, is that I want it to be quite clear where errors occur. image

Hi @IanMayo, executable for mac os can run on intel chip and arm64 chip, so we just need to build 1 version and use for both. Only sqlite3.node file needs to be built with its correct architecture.

PR #1632, opened 2022-06-22
Delete annotation marker

### LINKED ISSUE #1628: Delete Marker Feature
This is our current marker editor on the bottom left: 

Map SC

Currently, we do not support a delete feature, where GC can remove the marker from the game entirely. 

I suggest we do so by adding a red delete button at the bottom of the editor. 3 options

Option 1: 

Marker Editor Option 1

Delete button on the far left since the action-ability of the editor is heavily located on the right hand side. It is also consensus for MacOS, iOS and Android to have the delete button immediately after the cancel button.

Option 2: 

Marker Editor Option 2

Bigger attentional impact given that 2 bright colors sit beside one another. 

Out of 1 and 2, I have a preference for option 1. 

Also, notice that I've rendered the cancel button transparent. That is to reduce visual pollution since we're adding a red color down there.

Option 3: There's also the option of having the bin placed instead of the delete button on the far left: 

Marker Editor Option 3

Less invasive for the eyes and somewhat of a smoother look, but not very conventional. 

Let me know your thoughts.

IanMayo:
Fixes #1628


IanMayo:
@lebaphi  - I could do with some support on this tomorrow, please.

You can access the story for the `annotation marker` via `Mapping \ Small Scale \ Natural Earth`.

There are two things:
* [x] The trash icon needs to be laid out like in the [UX](https://github.com/serge-web/serge/issues/1628#issuecomment-1163212942)
* [x] I need a confirm dialog after delete is pressed. I'll try to insert a comment/note for where this would be.

packages/components/src/local/marker-form/index.tsx
```diff
@@ -71,6 +71,10 @@ export const MarkerForm: React.FC<PropTypes> = ({ formData, updateMarker, closeF
     updateMarker(UPDATE_MARKER, payload)
   }
 
+  const deleteMarker = (): void => {
+    console.log('handler to delete marker')
+  }
```
IanMayo, line 74: I need a delete confirmation dialog to popup when `deleteMarker` is called.  If the user cancels, then no change. If the user confirms, then the code should call a `deleteMarker` callback.  Yes - I have some naming trouble here ;-). Let's rename this handler method as `deleteMarkerHandler`.

lebaphi: I will take care of it @IanMayo 👍

lebaphi:
@IanMayo I updated the UI, the logic delete marker is not implemented yet, do you want me to take care of it?

https://user-images.githubusercontent.com/26705621/175288436-a0f561a2-6cc2-4f0f-93de-7acaf51b29a9.mov

IanMayo:
it's ok thanks @lebaphi  - I'll handle the delete :-)

ShendMikullovci:
Just checked deployment. Works well. Thanks guys!

IanMayo:
screencast

PR #1644, opened 2022-06-28
Fix issue style run time error

### LINKED ISSUE #1642: Runtime style error
In the debug console we're displaying a warning for `rcb` elements:
image

It doesn't really bother me, but it's cluttering the console.

Steps to reproduce:
* run wargame
* log into `small-wargame` as `White` - `Game Control`.
* click on an information marker
* warning appears on console

lebaphi:
Fixes #1642

### DIFF
--- a/packages/components/src/local/form-elements/rcb/index.tsx
+++ b/packages/components/src/local/form-elements/rcb/index.tsx
@@ -16,10 +16,7 @@ const buildStyles = (options: any[]): any => {
   const opts = {}
   options.forEach((option, idx) => {
     opts[`root-${idx}`] = {}
-    opts[`checked-${idx}`] = {}
-    opts[`checked-${idx}`]['&$checked'] = {}
     opts[`root-${idx}`].color = `${option.colour} !important`
-    opts[`checked-${idx}`]['&$checked'].color = option.colour
   })
   return makeStyles(opts)
 }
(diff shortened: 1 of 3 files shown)


packages/components/src/local/form-elements/rcb/index.tsx
```diff
@@ -16,10 +16,7 @@ const buildStyles = (options: any[]): any => {
   const opts = {}
   options.forEach((option, idx) => {
     opts[`root-${idx}`] = {}
```
IanMayo, line 18: well, I wasn't expecting it to be this simple 🤣 

Does the UI for this form still look ok @lebaphi  ?

lebaphi: Yep, it works fine @IanMayo but I see other error when loading the page on first time.

https://user-images.githubusercontent.com/26705621/176198455-3d4f1511-f731-4a8e-a916-ab086235de38.mov

IanMayo: Ok, I see a few errors:

* `Warning: Cannot update a component` - yes, it would be good to fix this.  I'll create an issue for it.
* `[Violation] 'message' handler` - I don't know much about this
* `HEAD http://localhost:3000/assets/couners/unk` - Yes, this may be an issue.  I'll investigate.

IanMayo (approved):
Issue resolved.

PR #1645, opened 2022-06-28
1587 laydown logic

### LINKED ISSUE #1587: Force Laydown for both Umpire & Players
When designing a game, if the Game Designer adds any assets, their initial locations need to be set.  This _could_ be implemented by the Game Designer entering the (horrible) hex coord.  But, it's easier to create the assets, and then arrange them via drag/drop within the game itself (before game initiated).

image

Also, for some wargames, the Game Designer may let player(s) specify the initial positions for some assets.

image

### Location Pending pen
We need to think through the UX for both - including where we locate the "pending" assets. After all, they don't have an initial area.  One solution is to have a box at the top-left, with one un-placed asset per cell.  The GD or player drags the asset from there to the desired location.

Maybe think of a better name than `location pending pen`, too.

### Laydown phases:
1. Game Design (game not initialised)
2. Umpire laydown (game initialised, one or more assets set to `Force laydown`)
3. Player laydown (game initialised, no assets set to `Force laydown`, one or more set to `Player laydown`)
4. Game-play (game initialised, no assets set to `Force laydown` or `Player laydown`)

### Tasks: 
* [x] Allow game designer to set laydown/initial location in admin pages (Almost complete)
* [x] Logic to decide which laydown phase we're in. See `laydownPhaseFor()`
* [x] _Somehow_ make Game Designer (or players) aware of which laydown phase we're in.
* [x] Allow routes to be generated for assets that don't have a position or location
* [x] Logic to let the `asset-icons` component decide where to put the _Pen_, render the pen, and also create positions/locations for the `asset-icon` component to use
* [x] Logic to allow player to drag/drop pending assets
* [x] Logic to submit laydown positions

IanMayo:
## 🧰 Issue

Fixes #1587 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/components/src/local/map-icon/index.tsx
```diff
@@ -72,14 +72,17 @@ export const MapIcon: React.FC<PropTypes> = ({
   // get top orient marker in the list
   const lastOrientation = orientationData?.length ? (orientationData[orientationData.length - 1] as OrientationData).orientation : 0
```
IanMayo, line 82: Hello @lilitkarapetyan - I could do with a hand in this PR please.

For assets that are available for laydown (player can move them before game starts), we'd like them to blink, as described here:
https://github.com/serge-web/serge/issues/1587#issuecomment-1161443960

Could you make assets with `locationPending` flash/blink please?

It _looks like_ it may be possible to do this at the level of `assetIconComponentAsString`, `divIcon`, or `Marker` - I don't know which is the best to use.

PR #1652, opened 2022-07-05
1635: Update tab button style

### LINKED ISSUE #1635: General Chat Redesign
I thought of redesigning the general chat, since we've encountered multiple issues in implementing new features (main menu button #1609), bugs (#1619), or improving UI for current features (#1580). 

Came up with this: 

New General Chat Opened

The title of the wargame at the top on the right hand side, accompanied by user's role and force. On the left of the label, the force icon. 
Objectives would appear by clicking on the Force icon, similarly to our current layout. The difference here would be:

1. Bottom of the screen is reserved for messaging (both on the left hand side for channels, and for the right hand side for chatting), as we are used to in most messaging apps.
2. It reinforces the Z pattern that our layouts try to encourage: 

Z Pattern Example

Here, the user would naturally encounter information following the usual Left-Right, Top-Bottom route.

Home button, would be added next to the Insights, Player Log and Intro buttons. 
The Turn and Phase information also gain importance and centrality here, considering they are crucial info for GC. The action button in light blue against a white background for clarity and presence. 

Furthermore, the chat would become collapsible to give GC the ability to reduce information load: In a busy game, it would be harmful for GC to be distracted by messages from the right hand side, especially if they are redundant. 
The tabs remain similar, with the exception of their appearance: they look like Google Chrome tabs. My girlfriend accidentally stumbled upon an open SERGE Screen and asked me what she was looking at. She then let me know that Insights/Game Admin didn't look like tabs at all. I realized she might be right...
The "Mark All Read" button gets replaced by our usual envelope to maintain uniformity and consistency. All of this gives the chat a much more "chat"-y feel. 

When collapsed, it would look like this: 

Reducing information load dramatically and giving a clean look. GC would still have a marker on the chat to remind them of accumulating messages on the chat bar at the bottom right.

Expanded, it would look like this: 

General Chat Expanded

Here, GC can view older messages without having to scroll upwards. If messaging is important or task-specific, the user can expand the chat to increase focus. Bold text and message chip for new, unseen messages, that remains bold for a specific time.

I think this would improve game flow, all the while solving some issues we've encountered concerning our current UI.
Let me know what you think.

lebaphi:
Fixes #1635 

image

### DIFF
--- a/packages/themes/dependencies/flexlayout-react.scss
+++ b/packages/themes/dependencies/flexlayout-react.scss
@@ -128,6 +128,12 @@
   @include typo-secondary;
 }
 
+.flexlayout__tab_button--unselected {
+  border-top-left-radius: 10px;
+  border-top-right-radius: 10px;
+  background-color: 	#1c3a55;
+}
+
 .flexlayout__tab_button--selected {
   font-weight: bold;
   color: #ffffff;
--- a/packages/themes/dependencies/flexlayout-react.scss
+++ b/packages/themes/dependencies/flexlayout-react.scss
@@ -132,7 +132,8 @@
   font-weight: bold;
   color: #ffffff;
   box-shadow: 0 -2px 4px 0 rgba(0, 0, 0, 0.4);
-  border-style: solid;
+  border-top-left-radius: 10px;
+  border-top-right-radius: 10px;
   border-width: 1px;
   border-image-source: linear-gradient(to bottom, #3fb7ac, #9dfff1 36%, rgba(2, 37, 96, 0));
   border-image-slice: 1;
--- a/packages/themes/_print.scss
+++ b/packages/themes/_print.scss
@@ -1,5 +1,5 @@
 .print-canvas {
-  background: white;
+  background: #fff;
   padding-top: 1px;
 
   section {


IanMayo (dismissed):
Very good. You know your CSS :-)

IanMayo:
@lebaphi - now that we've gone for the genuine tab "metaphor", I think the un-selected tabs should be a little lighter, so we can see their curved outline, too.  Let me have a play.

packages/themes/dependencies/flexlayout-react.scss
```diff
@@ -132,7 +132,8 @@
   font-weight: bold;
   color: #ffffff;
   box-shadow: 0 -2px 4px 0 rgba(0, 0, 0, 0.4);
-  border-style: solid;
+  border-top-left-radius: 10px;
+  border-top-right-radius: 10px;
   border-width: 1px;
   border-image-source: linear-gradient(to bottom, #3fb7ac, #9dfff1 36%, rgba(2, 37, 96, 0));
```
IanMayo, line 144: @lebaphi  - I _don't think_ we need the gradient background image any more, do we?

IanMayo (approved):
Have made unselected tabs darker/rounded.

PR #1657, opened 2022-07-07
1579 UX for Mapping Constraints

### LINKED ISSUE #1579: UX for mapping constraints
While it has been hard-coded in the games so far, we need to allow the Game Designer to specify the constraints on the mapping.

These are the fields we require for the mapping:
* **Bounds**: These are the top-left and bottom-right corners of the map to show. It should be two pairs of values, taking - say a number of degrees, with up to two decimal places (e.g. `12.25`)
* **H3 Resolution**: this is the hex cell grid size to use for the [H3 grid](https://h3geo.org/docs/core-library/restable/) (integer, from 1 to 10)
* **Tile Layer**: URL of path to where backdrop tile images may be found
* **minZoom**: the minimum zoom level, preventing the player from zooming out any further. (integer, from 1 to 12)
* **maxZoom**: the most a player can zoom in
* **maxNativeZoom**: the limit of zoom on the backdrop tiles. After this limit is reached, the smaller one will get zoomed appropriately
* **gridCellURL**:  path to file where grid cell styles can be retrieved.  So, for each cell there will be a style such as "1 - Land", "4 - Storm Areas", "12 - Fishing Area"
* **polygonAreasURL**:  for areas covering huge areas, plotting a simple set of polygons for each terrain type (land, deep-water, etc) is quicker than plotting all of the discrete hexes.  A scientist was able to do the clustering for his hex cells, and provide it here - which made the task of rendering such a large area much easier.

We have some existing thoughts on this, and the overall layout of "Overview" in #991 & #1425. We should consolidate the existing designs controls with new controls for the above set of fields.

lebaphi:
Fixes #1579 

image


IanMayo:
Thanks @lebaphi 

Couple of things:
- could you make the new panel collapsible. It should start off as collapsed.  Once collapsed a "text overview" label should be provided.  I can fill in the details if you give me a placeholder.  There should be a visible "collapse" control that closes/expands the new panel.  We use a `+` icon for this elsewhere.
- could you reduce the overall space required?
- the text fields are currently 3x the font height. Removing the `margin` from the formControl makes it all a lot smaller
- for the TileLayer, could you add a "memo" font-awesome icon for the "attribution"?  Then the triangle is just used for the tile-layer control

lebaphi:
sure @IanMayo

lebaphi:
Hi @IanMayo , I've updated 😄 

https://user-images.githubusercontent.com/26705621/178022460-6daa37bc-d601-4607-a494-6a08450f9fd7.mp4

IanMayo:
Nice work. Could you put a text label in the accordion, under the title? 
I'll put some summary text in it (such as the corners of the bounding box).

lebaphi:
I've updated

image

IanMayo:
> I've updated

Looks perfect, thanks :-)

IanMayo:
Ok. For the 4 text fields, it would be good if they were wider. So, could they please expand to consume the remaining horizontal space?
That won't work for the h3 resolution, so could that be moved to the bottom of the first column please?


image

lebaphi:
Hi @IanMayo , I've updated, for the H3, I think we should keep its wide as the same as the others in the first column 🤔 

image

IanMayo:
The new layout works well, thanks. I need to give some more detail on what some of the fields mean.  Could you give each field icon a "more info" icon, like we do in other panels?

image

lebaphi:
okay @IanMayo

IanMayo:
> The new layout works well, thanks. I need to give some more detail on what some of the fields mean. Could you give each field icon a "more info" icon, like we do in other panels?
> 
> image

Oops, forget that please @lebaphi .  I've done it myself - I'm just writing the guidance now.
image

IanMayo (approved):

PR #1664, opened 2022-07-11
let game designer edit all attributes

### LINKED ISSUE #1616: Editable vs Non-Editable Attributes
Attributes in the Admin Page can be editable by players or not editable thanks to the toggle feature we've added. 
However, I noticed that in the Forces --> Assets tab, if an attribute has been made non-editable by players, the GC can't edit it either: 

https://user-images.githubusercontent.com/85491794/174677398-00555ad7-f69a-471c-98bc-4b3e3e436160.mp4

GC should be able to edit the attributes in both of the tabs (Force --> Assets and Platform Types), since they are not a "player" and have special access. This would also simplify GC's work process in the case where they might change attributes for a force. Much simpler to just go to Forces --> Assets and modify things there. This way, we give the GC versatility in doing this process.

### LINKED ISSUE #1617: Attribute Editing to be Updated Across Tabs
Editing attributes in the Forces --> Assets tab, should automatically update the attributes in the Platform Types page. Currently, this isn't the case: 

https://user-images.githubusercontent.com/85491794/174677820-2dfa9d5c-0670-4807-995c-f2364d74a8c2.mp4

If we are to keep this versatility on editing the attributes (which I'm pro of), the attribute edited on one tab, should be updated on the other.

IanMayo:
Fixes #1616


ShendMikullovci:
Seems to be working correctly. Thanks!

ShendMikullovci:
Update: It's only working one way it seems like.

https://user-images.githubusercontent.com/85491794/178754328-c4aae1c4-c840-431b-856d-53362f3246e2.mp4

IanMayo:
> Update: It's only working one way it seems like.

Ok. I can't say it was intentionally implemented that way, but what you saw was correct.  If we create several frigates for a force, we may wish to give them different initial values of, for example, fuel.  It would be wrong for editing an instance of an attribute to change the original definition of that value.

So, on the `platform types` page the attribute is the `master`. On the `force - assets` page, the attribute is an `instance` of the `master`.

ShendMikullovci:
> If we create several frigates for a force, we may wish to give them different initial values of, for example, fuel. It would be wrong for editing an instance of an attribute to change the original definition of that value.

Sure. However, as far as I understand, you can't change instances for ALL frigates within a Force. Only for one of them. 
If _one_ instance is changed in the assets page, it should technically only alter _one_ frigate in the platform types page, correct? Or am I missing something?

IanMayo:
> If one instance is changed in the assets page, it should technically only alter one frigate in the platform types page, correct? Or am I missing something?

The _thing_ on the platform types page is a `template`.  When a new frigate is created in a force, a _copy_ of the `template` for the attributes is added.  When the frigate it edited, it's just the _copy_ that is being edited.  Each frigate has their own copy of the original attribute template.

ShendMikullovci:
Hmmm. So what happens to the frigate I edited on the asset page? Which attribute does SERGE consider as the _correct_ one to use? Platform types one (since it's the master page)?

IanMayo:
When you edit the frigate on the asset page, the attribute value should be stored to the database, and used in the game.

So, if the attribute hasn't been edited, it will still have the default value (from the platform types page). But, if it has been edited, it will have the new value.

ShendMikullovci:
So, once edited on the Asset page, it will be used In-Game, but will not be shown in Master page?

IanMayo:
The "thing" on the platform-types is unchanged. It's still the master, to be used as a template for new frigates.

I'm stuggling for a metaphor here , but I'll give it a go :-)
1. I get a ferrari brochure, and like the look of one of them
2. I buy a ferrari, and it looks just like the one in the brochure
3. I fit monster truck wheels on the ferrari, and take it off-road
4. The vehicle in the brochure still has standard tires.

ShendMikullovci:
Hilarious yet very clear and comprehensive analogy.

ShendMikullovci:
> The "thing" on the platform-types is unchanged. It's still the master, to be used as a template for new frigates.

The only thing I'm worried about is any possibility of confusion for GD when they see one value on Platform Types and another on Assets page. That's why I suggested this change.

IanMayo:
Will prepare animated infographic of my metaphor, and insert it into the tooltip 🤣

ShendMikullovci (commented):
Resolved with Ian in Conversation.

Issue #1400, opened 2022-03-11
Large numbers of forces


IanMayo:
For the next wargame, the Game Designer wants lots of people driving ships.

Since we have one "driver" per force, that means we have lots of forces.  So, lots of Blue Forces with one player each, and lots of Red Forces with one player each.

This could introduce problems in these areas:
* In the mapping window Umpire forces can choose which force to view the map as.  This will become a really tall list once we have 20 forces. So, if there are a large number of forces it should turn into a drop-down list
* In the "Perception" dialog the Game Controller specifies which forces can see an asset.  Once we have more than, say, 6 forces we should switch to a drop-down list of toggles, one per force.
* In the Message Channel for `custom` messages we show an icon for each member force at the top. This may need to scroll/wrap once there are a lot of them
* Since all blue forces will have a similar shade, it may become difficult to spot which are a players own asset(s).  So, we could do with introducing some other styling for "own" assets on the map.
* We currently use color to to indicate which force a message comes from.  But. if there are lots of blue forces we will need the force name in the message somewhere.

We probably need to create a wargame with, say, 10 blue forces and 5 red forces, and see where the bottlenecks/problems arise.

IanMayo:
@ShendMikullovci - I could do with your support in this, please - both in spotting the issues, and in coming up with solutions to them.

ShendMikullovci:
Sounds very exciting! I'll do some research on the consensus among strategy wargames and get back to you ASAP.

IanMayo:
Aah, a related fix, please @lebaphi   For the admin pages that have a list of entries, the height of the left-hand column is smaller than it could be.  This will be a pain if/when we have lots of forces.  It already can't show all of the platform types.

So, the admin pages with a list of entries should use the full height of the left-hand column.

![image](https://user-images.githubusercontent.com/1108513/157916566-36f86883-9bf4-46db-9584-6f11e464af85.png)

ShendMikullovci:
> ... the height of the left-hand column is smaller than it could be. 

This is also relevant for the Game Selection Page : 

![Game Selection Page](https://user-images.githubusercontent.com/85491794/157917551-5b9b11d5-86a6-47f7-9b2b-818c37732d25.png)

If we could show the entire list of the games we host, we'd eliminate the scrollbar entirely and the screen would look smoother and more fluid.

ShendMikullovci:
[...]

Issue #991, opened 2021-08-11
As a Game Designer I need to specify turn interval in months/years


IanMayo:
Serge currently allows game turn lengths to be specified using days, hours, minutes, seconds:
![image](https://user-images.githubusercontent.com/1108513/129016115-c59b0f9b-feb3-4c25-abb3-630395456901.png)

For larger scale wargames, the interval could be in months or years, such as each turn spanning a two-month period (Jan21-Feb21, then Mar21-Apr21, then May21-Jun-12, etc). Given the changing length of both of these items (in terms of days), we need an alternate way of specifying game turn length.

So, I _think_ we need a range of turn length strategies:
* days/hours/minutes
* months
* years

Months and years would just take an integer value, using as appropriate in the turn time incrementing.

I _think_ we need a two-step UI.  First select the turn mode, then enter the value.

Here is what the data structure will look like, I guess:
```
export interface MillisecondTurns {
  millis: number
}

export interface MonthTurns {
  months: bigInt
}

export interface YearTurns {
 years: bigInt
}

export type GameTurnLength = MillisecondTurns | MonthTurns | YearTurns

```

Well, to be backward compatible with our current data, it would be:
```
export type GameTurnLength = number | MillisecondTurns | MonthTurns | YearTurns
```

IanMayo:
Hello @ShendMikullovci 
this is a feature that overlaps with your other design task (#954)

So far, the turn-period has been expressed in days, hours, or minutes.  All of these can be represented by a single number (milliseconds) in the software, or via a `DD HH:MM:SS` masked text control.

But, in this game they want the turn to move forward 2 months. So, they expect it to step from 1st January to 1st March - **not by 60 days**.  The time period represented by two months changes - so we need a different way of representing that.  While we add `months`, we should probably also support `years`

So that's 3 different ways of expressing the turn size:
* Days/Hours/Minutes
* Months
* Years

I _guess_ we could support this through a drop-down to select the type, then a suitable text field to provide the detail.  Or, a control with three tabs - one for each style. Each tab would contain a suitable edit control - and we'd store the value of the selected tab.

Aah, once the design for this is complete, we can feed it back into #954 

(I've separated this out because this feature would be useful for the current wargame)

ShendMikullovci:
After our talk, and our joint affinity over the Tabs idea, I decided to give the design a try. This is how the tabs look: 
[...]

PR #1674, opened 2022-07-15
1635: General chat redesign

### LINKED ISSUE #1635: General Chat Redesign

lebaphi:
Fixes #1635 

image


IanMayo:
Nicely done @lebaphi 

1. The icon next to the `wargame title` is an `exit` button - which should take the player back to the `select game` / `select role` view, please.
2. I think we can get rid of the horizontal line beneath the wargame title.  The block below is in a different colour, so we already know they're different elements.

lebaphi:
I've updated @IanMayo

IanMayo:
Thanks @lebaphi 

One more item. There's an extra "Planning" in the turn panel:
image

Aaah, hold on, that's not the issue.  I realise what @ShendMikullovci intended.  The box should contain the two words "Planning" and "Adjudication".   The "current" one will be shown in highlight, and the other darker.  So, if you play as "White - Game Control", you can click on "Start next phase" and the panel will cycle between the two states.

ShendMikullovci:
Hi guys! Looks much smoother already. Couple of things: 

1. The screen seems to be outlined now: 

Outline

I'm not sure if this effect was intended. Kind of gave me the feeling that I was sharing the screen on Zoom or like my screen was being recorded. 

2. The features on the right hand side seem to be very tightly joined together: 

Space

The area in red is quite squeezed in while the area in orange occupies too much unnecessary space. We could gain quite a bit of space by shrinking the chat length and redistributing the features, spacing them out and giving them some more leg room.

3. Start New Phase button could do in light blue, to attract attention as key button for this screen.
4. The tabs could do a better job depicting their function if they were squared onto the chat itself. Nothing too fancy, simply making them the same size as the chat box. It could look something like this: 

Chat Tabs

5. The Read/Unread button seems to only be marking messages as read, but not unread.

IanMayo:
1. Well spotted @ShendMikullovci - I hadn't seen that.   @lebaphi - there is a thin grey line around the edge of the app (left side in following screenshot):
new border
2. You are on a tall screen.  In the past 2 wargames, players have sat in a meeting room, all looking at a projector version on a screen.  The projectors available are quite low resolution - so Serge really fills the screen. That is partly why the "out-of-game" component is so narrow.  The only way for Serge to work on a low resolution is to squash the controls together, but let other controls expand to fill the remaining space.
3. `Start new phase`.  Only the Game Control sees this button.  If it was lighter, it would lose contrast, since the letters are white, and the background is white.  Is there a precedent for a light-blue button anywhere else in the app?  Another option is to use a larger text.
4. Yes, I see your point. I guess they don't have to exactly fill the space, but the left-hand of "Insights" should certainly align with the left of the panel.  So, IMHO the right-hand of the tabs doesn't have to align - since we don't do that in the "in-game" messages:
image
5. Yes, well spotted. Could you record this as a separate issue please, it's something that was already broken, and doesn't relate to these UI updates.

lebaphi:
@IanMayo @ShendMikullovci , for the issue 5, I only see the button `Mark All As Read`, that's why we only have one icon to mark all message as read.

image

IanMayo:
Hello @lebaphi - one more item:
https://github.com/serge-web/serge/pull/1674#issuecomment-1185625824

ShendMikullovci:
> 3. `Start new phase`.  Only the Game Control sees this button.  If it was lighter, it would lose contrast, since the letters are white, and the background is white.  Is there a precedent for a light-blue button anywhere else in the app?  Another option is to use a larger text.

Oh, my bad. The blue I'm referring to is the one I used in the latest designs of #991 ("Save Overview" button), #1400 ("Start New Phase" button), etc. It's HEX is 1A65D7. It doesn't lose contrast in white (actually ends up popping out more) and it doesn't blend in with our dark blue backgrounds. 

> 4. Yes, I see your point. I guess they don't have to exactly fill the space, but the left-hand of "Insights" should certainly align with the left of the panel.  So, IMHO the right-hand of the tabs doesn't have to align - since we don't do that in the "in-game" messages:
>    image

I understand. However, the general chat tabs and in-game channel tabs are two different features that serve very different purposes. They've also been clearly presented as different (through the intro steps), which is why a different display logic would apply to each. That's why I think it's important to treat the chat as such.

ShendMikullovci:
> @IanMayo @ShendMikullovci , for the issue 5, I only see the button `Mark All As Read`, that's why we only have one icon to mark all message as read.
> 
> image

I'll include animations for buttons that have double-use from now on.

ShendMikullovci:
> > @IanMayo @ShendMikullovci , for the issue 5, I only see the button `Mark All As Read`, that's why we only have one icon to mark all message as read.
> > image
> 
> I'll include animations for buttons that have double-use from now on.

Here's the animation for that @lebaphi :
https://user-images.githubusercontent.com/85491794/179865778-2c033009-4ba4-4692-889e-5ee54eeb02cb.mp4

As you can see, hitting on the "Mark Read" button removes the notification counter and the bold effect on the messages. Hitting on "Mark Unread" brings back the counter and the bold effect.

IanMayo:
Nice work, thanks @lebaphi 

I _think_ we should de-highlight the "not current phase" label.  The current one is larger, but I the non-current one is the same size & color as the "Phase" label.  IMHO it should be 50% transparent, or something like that...

What do you think @ShendMikullovci ?
screencase

lebaphi:
I've updated @IanMayo 
Screen Shot 2022-07-20 at 8 19 53 PM
Screen Shot 2022-07-20 at 8 22 26 PM

IanMayo:
Hi @lebaphi - I'm just testing new phase presentation.

It looks good. But, I have a minor vertical alignment issue with this wargame:
image


I _guess_ the `h3` and the `fontawesomeicon` need to be vertically aligned in the middle of the parent box.

IanMayo:
Oh dear @ShendMikullovci - I think I've just come up with a scary issue.

`START NEW PHASE` is below 3 buttons and above 2 tab buttons.  A bad mouse click for any of them could accidentally trigger a step to a new phase - throwing away routes currently being planned.

But, if we move the button to above the "turn marker" box, it's only near to the "return to welcome page" button - which isn't expected to be used very often at all.  Aah, or we could put it back inside the "turn marker" box. If it was above the "Current time" label, there is a "buffer" between the 3 action buttons and the next step button.

image

Note: if we do put "next phase" button inside "turn marker" box, we'll need a different shade.  @ShendMikullovci has already suggested shade to use:
image

IanMayo:
After endless linter fixes, I've pushed a version with "Start Next Phase" back inside the Turn Progression box:

image

So, @lebaphi - I think the only remaining issue is the "Welcome page" icon alignment: https://github.com/serge-web/serge/pull/1674#issuecomment-1190347468

ShendMikullovci:
Thanks guys! 

Love the dimming of the inactive phase. The button inside the box works as well, helps us avoid that accident hopefully. 

Couple of things I wanted to address: 

1. Could we have rounder corners for the action buttons please? It smoothes out the entire look and makes them look more like actual buttons. 
2. I also noticed the button reacts by changing colors upon hover. I don't think that's necessary. A simple dropshadow could do the trick as well:

https://user-images.githubusercontent.com/85491794/180098662-5cdb98ef-d224-4273-bbb6-2d04db25007c.mp4

3. The turn box could do with some rounder corners as well and the same color as the "Game Admin" chat tab. Now that I think of it, why did we settle on the turn box not covering the width of the chat? @lebaphi could be right by putting an outline between the chat and the screen and we might have rushed our hand telling him to remove it...

Essentially, the right hand side has had two main issues regarding visibility: It blends in with the screen (unless we use bright colors) and it blends with itself (chat blends in with upper features/buttons). This latter issue is even more important because the chat and the features that happen to be above it are actually completely unrelated. By placing an outline, we make that dichotomy: 

New Screen

Here, the time box (+ features) and the chat are visibly separate from the channels (which maintains our conscious intent on separating them). But more importantly, the time box (+ features) and the chat are _visibly separate from each other_. Our current display doesn't really emphasize that. In this design, every element takes up its adequate spot and is equally visible (I feel like?). The reason why I left the 3 buttons outside of the box (namely, Feedback, Player Log and Intro) is because we want to avoid the user accidentally clicking Start New Phase. This visual separation reduces that risk. What do you guys think?

IanMayo:
> Rounder corners

Yes please @lebaphi - let's have rounder corners for the whole "turn progression" box, and the "Start new phase" button.

> I also noticed the button reacts by changing colors upon hover. I don't think that's necessary. A simple dropshadow could do the trick as well:

I've just done some investigating.  I didn't think we had hover action on any other buttons, but it seems most of the `Submit` buttons do change color on hover (I think it's a standard shading that comes with our Material UI toolkit).  The hover shadow is probably too subtle for many Serge users.  So, I think it's right to stick with the hover - we've just got to find the right color.

@ShendMikullovci - there is a gap between the bottom of `Game Admin` and the `New admin message` box.  I don't think this is right, since the `new admin message` box is a part of `Game Admin` - and it's not present for the `Insights` tab.

IanMayo (approved):
Couple of outstanding items, but will handle these as separate issues.

ShendMikullovci:
> @ShendMikullovci - there is a gap between the bottom of `Game Admin` and the `New admin message` box. I don't think this is right, since the `new admin message` box is a part of `Game Admin` - and it's not present for the `Insights` tab.

Hi Ian! You're right. That's because the tabs lack the curvature to make them visually connected to the chat. My bad, lost focus on that. Here we go: 

Rounded + Connected tabs

ShendMikullovci:
This got me thinking about the design for the player end. They don't have access to an insights tab, so how would the chat look for them? 

Couple of ideas: 

Chat Looks for Players

Tab look: The tab is expanded to fit accross the chat window, with the side curvature included. One problem with this is that you can argue that essentially, it's not a tab, since there's no other tabs next to it. Instead, it's more of a header. 

Header 1: Same size without curvature on the bottom placed in the same spot on the X/Y axis. 

Header 2: Larger size (to touch the top of the chat) placed in the same spot on the X/Y axis. For Header 1 and 2 you can argue that the color is inappropriate, since for GC we use the navy blue to indicate the opened tab (which has been clicked on). Players' side doesn't present the option to click (since there's no Insights or other tab to click on).

Header 3: Same as Header 2, except white. Problem here is that it would be laying on top of a white background. Tried a grey one and it's just plain ugly.

We could also go with a no-tab look: 

No Tab Chat

1st one only has a long underline, similar to the title of a wargame in the admin pages. The 2nd one only underlines the title of the chat. 

Let me know what you guys think!

IanMayo:
Hello @ShendMikullovci - thanks for considering this.

Right now, the single-tab issue isn't a great worry.  There _may_ be games where the player is only a member of one channel - so there will only be one tab on the left-hand side.  That's not an issue.

My issue with your designs above remains.  The "new message" box has to be a _child_ of the `Game Admin` component. That's because it isn't present when viewing `Game Insights`.  So, there should not be a space between the list of messages Game Admin messages and the new message box.  The box will not be present for "Insights".

The current layout suggests the new message box isn't directly related to "Game Admin" messages.

ShendMikullovci:
Oh! Something like this you mean?: 

Chat Message Without Box

IanMayo:
Yes - that solves the issue.

PR #1678, opened 2022-07-19
1677: Add Default Value Column

### LINKED ISSUE #1677: Need field for "Default" value in "Attributes"
In the Admin Pages, we can edit the attributes for each platform type.
image

The `Number Attribute Type` element includes an optional value for "default value":
image

I see that the original design (in #1291) did include the default value:
image

So, we should add a column for "Default value".

lebaphi:
Fixes #1677 

image


IanMayo:
Working perfectly :-)

IanMayo (approved):

PR #1681, opened 2022-07-20
1676 enumerated attributes

### LINKED ISSUE #1676: Enumerated attribute values
As a ***Game Control*** I wish to ***enumerated attribute values*** in order to ***improve attribute usability***

For a future wargame (WATU) the depth of submarines will be significant - and should be an attribute.

While we could allow players to enter a depth in metres, it would be better allow them to choose one of `Deep`, `PD` (short for _Periscope Depth_ and `Surface`.

We'll have to make this change both in how attributes are specified (in platform types) and how they are edited (in the attribute editor).

For attribute specification we'll need a new drop-down for the attribute. It will contain "Number" or "Strings".  If "Strings" is selected, then we'll have to allow the Game Designer to enter multiple values, plus the default value.  

For the editor we'll provide a drop-down list of values.

Maybe @ShendMikullovci 's existing design documents are available, and we can extend them to support the new functionality.

IanMayo:
Fixes #1676


IanMayo:
[request deleted, I've coded it myself]

You can use the `ssn` platform-type in the `Storybook` for `Setting platform types`.

There is now just this to handle, for `Enumerated Attributes`:
- the `default value` should be a combo box that contains the elements from the `values` array

I did think we needed a popup editor for the possible values, but it's actually possible to just parse the text value as a comma-separated set of values. I can implement that.

Note: there is some `fancy-processing` of the list of values on a change. This is because while there is one list of values in the AttributeType, we treat it as two lists in the editor table. So, on a change we have to produce the correct list before we can use the row `key` value.

PR #1694, opened 2022-07-27
1579: Update UI Mapping Constrains

### LINKED ISSUE #1579: UX for mapping constraints

lebaphi:
Fixes #1579 

https://user-images.githubusercontent.com/26705621/181159168-e1e55d68-a5c9-4d0a-9c86-320787fe06b3.mp4

### DIFF
--- a/packages/components/src/local/organisms/setting-channels/styles.module.scss
+++ b/packages/components/src/local/organisms/setting-channels/styles.module.scss
@@ -128,8 +128,11 @@
 }
 
 .accordion-title-group {
+  width: 100%;
   display: flex;
-  flex-direction: column;
+  flex-direction: row;
+  justify-content: space-between;
+  padding-right: 20px;
 }
 
 .accordion-title {


IanMayo (approved):
Great, thanks.

Good skills, @lebaphi

PR #1695, opened 2022-07-27
1683: Update Blue Shade

### LINKED ISSUE #1683: Consistent use of blue shade
We appear to be using three shades of dark blue (`a`, `b`, `c` below):
Screenshot 2022-07-21 at 12 02 34

There could be benefit in switching to just one shade.

Or we could have one shade of dark-blue for _in-game_ (left) and another shade for _out-of-game_ (right) components.  If we do want different shades, to highlight different parts of the game, then I suspect we should:
1. increase the difference between the two shades
2. use the same shades for `b` and `c`.

What do you think, @ShendMikullovci ?

lebaphi:
Fixes #1683 

Screen Shot 2022-07-27 at 11 22 40 AM


IanMayo:
Hey, @lebaphi  - that fade in the Turn Progression panel is stylish.  But, part of me thinks it's too subtle, and another part of me thinks it's unnecessary.

I'm pretty sure that as we get further down we lose contrast on the white text.  Damn, I've probably wasted too much time on this already 🤣   What do you think @lebaphi  & @ShendMikullovci ?

Zero fade:
image

10%:
image

20%:
image


30%:
image

40%:
image

50%:
image

60%:
image

lebaphi:
> 

lol, I tried to make the panel color as similar to the tab button color as possible. So imo, 0 or 10% fade is fine @IanMayo :v

IanMayo (approved):
Have left in top-fade, to give a "chrome" effect.

ShendMikullovci:
Love the final look. Great work guys!

PR #1707, opened 2022-07-29
Updating couchDb for live wargame

lilitkarapetyan:
## 🔨Work carried out:

- [x] Tests pass

## Confirmations

### DIFF
--- a/packages/server/providers/couchdb.js
+++ b/packages/server/providers/couchdb.js
@@ -72,12 +72,13 @@ const couchDb = (app, io, pouchOptions) => {
 
   app.put('/:wargame', (req, res) => {
     const databaseName = checkSqliteExists(req.params.wargame)
+    const cleanName = databaseName.replace(dbSuffix, '')
     const db = new CouchDB(couchDbURL(databaseName))
     const putData = req.body
     wargameName = req.params.wargame
 
-    if (!listeners[databaseName]) {
-      addListenersQueue.push(databaseName)
+    if (!listeners[cleanName]) {
+      addListenersQueue.push(cleanName)
     }
 
     const retryUntilWritten = (db, doc) => {
--- a/packages/server/providers/couchdb.js
+++ b/packages/server/providers/couchdb.js
@@ -185,16 +185,16 @@ const couchDb = (app, io, pouchOptions) => {
       res.status(404).send({ msg: 'Wrong Wargame Name', data: null })
     }
 
-    const db = new PouchDB(databaseName, pouchOptions)
+    const db = new CouchDB(couchDbURL(databaseName))
 
     db.find({
       selector: {
         messageType: INFO_MESSAGE,
-        _id: { $ne: wargameSettings }
+        _id: { $ne: wargameSettings, $gte: null }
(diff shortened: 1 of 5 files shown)


packages/server/providers/couchdb.js
```diff
@@ -185,16 +185,16 @@ const couchDb = (app, io, pouchOptions) => {
...
     }
 
-    const db = new PouchDB(databaseName, pouchOptions)
+    const db = new CouchDB(couchDbURL(databaseName))
 
     db.find({
       selector: {
         messageType: INFO_MESSAGE,
-        _id: { $ne: wargameSettings }
+        _id: { $ne: wargameSettings, $gte: null }
```
IanMayo, line 194: But @lilit - at the start of the game, there is just the `initial_wargame` document.  That one _is_ the `latest`.

IanMayo: Hmm. The game is saving and updating correctly, but I don't know how it gets the `initial_wargame`.

I ran this query, and nothing was returned.
image


Aaah, hold on. Maybe we _only_ get the `latest` wargame after it has been initiated - in which case we don't want the initial wargame :-)

Let me try that out on Monday :-)

IanMayo: Ok. I've just played.  It doesn't quite work.

In CouchDb we just have `initial_wargame`.  It is at Turn Zero, and `wargame_initiated: false`.

I then enter player view as White - Game Control.

My task is to give the ships their initial positions.  So, I zoom in a bit, then one-by-one select each ship from the "Laydown Pen" and position it.

Once they are all positioned, I click on "Submit Routes", and say "Yes" to the dialog.

Since the wargame is not yet initiated, we should still be overwriting `initial_Wargame`.

The _reducer_ for `force laydown` gets the newest wargame, sets the ship positions, and saves it as a new wargame.

image

It should have overwritten `initial wargame`.  It's only once `wargame_initiated: true` that we create new version(s).

IanMayo: Issue resolved! 

We only get the `latest` if the wargame has been initiated.  It's working fine :-)

Hold on, no.  I clicked on "Initiate wargame" in the admin pages, then came back to the player page and clicked "Start new phase".  The phase didn't move forward, but the player UI doesn't change. _Maybe_ we're correctly creating a new document ,but we're still displaying "initial_wargame"?  Will investigate more on Monday

IanMayo:
Hello @lilitkarapetyan - I think I've [fixed](https://github.com/serge-web/serge/pull/1707/commits/a1ac8f76b2a98b86919080ec06794c14ca9c4ee2) the remaining issue.

Would you mind checking it looks a valid fix?

lilitkarapetyan:
> Hello @lilitkarapetyan - I think I've [fixed](https://github.com/serge-web/serge/pull/1707/commits/a1ac8f76b2a98b86919080ec06794c14ca9c4ee2) the remaining issue.
> 
> Would you mind checking it looks a valid fix?

Hi Ian, it works fine, I think you fixed the rest of the issue.

PR #1708, opened 2022-07-31
1703: Refactor RCB into separate classes

### LINKED ISSUE #1703: Refactor RCB into separate classes
The RCB component is quite fragile, and has experienced a number of bugs.

Part of the fragility may be that it supports both Radio boxes and Check-boxes, which represent their data in different ways.  The props (especially update state handler) are poorly defined.   Oh, and there is "special handling" for lists of forces where colors are provided for the values.

Split into three components (one for RadioButtons, and one for CheckBoxes, one for Forces), and cover props (especially input data and change handler) with Typescript definitions.

lebaphi:
Fixes #1703

I have 2 small questions:
- Do we need to remove rcb component?
- Do we support Radio button in new Forces component?


IanMayo:
1. Yes, let's throw RCB in the garbage bin.
2. We only need check boxes for Forces, not radio buttons (it is always multi select).

packages/components/src/local/form-elements/checkboxes/types/props.d.ts
```diff
@@ -15,26 +11,19 @@ export default interface PropTypes {
...
-  force?: boolean
   /**
    * Sets the onChange state on a parent
    */
-  updateState?: any
-  /**
-   * Whether to use compact format (esp for list of color shades)
-   */
-  compact?: boolean
+  onChange?: (data: any) => void
```
IanMayo: can we provide type information for the `data` parameter please?  I guess it's going to be a string or a number.

packages/components/src/local/form-elements/forces/types/props.d.ts
```diff
@@ -0,0 +1,31 @@
...
+   */
+  options: SelectOption[]
+  /**
+   * The value to set the 'checked' parameter to
+   */
+  value: string | string[]
+  /**
+   * Sets the onChange state on a parent
+   */
+  onChange?: (data: any) => void
```
IanMayo: could you check what gets passed in data please?  I don't know if it's a `SelectOption` or the index of the selected items.

packages/components/src/local/form-elements/forces/types/props.d.ts
```diff
@@ -0,0 +1,31 @@
...
+   */
+  name?: string
+  /**
+   * The list of options available to the form section
+   */
+  options: SelectOption[]
+  /**
+   * The value to set the 'checked' parameter to
+   */
+  value: string | string[]
```
IanMayo: Hmm, I'm pretty sure this API will work better if this is the index numbers of the `SelectOption`?  Or, we should (somehow) indicate which `SelectOption` field it is, like `SelectOption['name']`.

packages/components/src/local/form-elements/checkboxes/types/props.d.ts
```diff
@@ -15,26 +11,19 @@ export default interface PropTypes {
   /**
    * The list of options available to the form section
    */
-  options: Array<string | number | FormOption>
+  options: (string | number)[]
   /**
    * The value to set the 'checked' parameter to
    */
   value: string | number | Array<string | number>
```
IanMayo: The `value` parameter seems to loosely defined for me - it could cause complexity in rendering the component.  Can we force it to be an array, even if it's just one item?

IanMayo:
One last thing, please @lebaphi 

I'd like to trim down the props a bit more, please.  We allow both `number` or `string` values for `radio-buttons` and `check-boxes`.  Could you have a look at where these components are used, and check if both number and string are used?   If either isn't used for one component - could we remove it from the props?

packages/components/src/local/form-elements/forces/types/props.d.ts
lebaphi: it should be an array of name (string[]) of selected items @IanMayo. I think it should be better if it's an array of an id of selected items.

lebaphi:
> One last thing, please @lebaphi
> 
> I'd like to trim down the props a bit more, please. We allow both `number` or `string` values for `radio-buttons` and `check-boxes`. Could you have a look at where these components are used, and check if both number and string are used? If either isn't used for one component - could we remove it from the props?

@IanMayo currently only the Forces component is used and the options data are all strings, so I removed the number props in this components already.

IanMayo:
Hey @lebaphi - it looks like the `CheckBoxes` component isn't used anywhere now. We can delete it :-)

Oh dear - I may be going mad here. It looks like `RadioButtons` isn't used anywhere either 😨 

Could you double-check that, and delete both classes if they aren't used anywhere please.

lebaphi:
> Hey @lebaphi - it looks like the `CheckBoxes` component isn't used anywhere now. We can delete it :-)
> 
> 
> 
> Oh dear - I may be going mad here. It looks like `RadioButtons` isn't used anywhere either 😨 
> 
> 
> 
> Could you double-check that, and delete both classes if they aren't used anywhere please.

yep, that is what Im thinking, they are not used anywhere so we should delete it.

IanMayo:
> yep, that is what Im thinking, they are not used anywhere so we should delete it.

Great. If you can delete them next time you're working - I can test the new Forces List control in the Serge panels.

IanMayo (approved):
A larger piece of work than either of us expected @lebaphi :-)

PR #1715, opened 2022-08-04
Not cache-ing ASSET icons

lilitkarapetyan:
## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Thanks @lilitkarapetyan - could you explain how the changes work in the two code blocks please?

IanMayo:
There are still more icon downloads than I'd like, but I guess we're using the icons in different locations - so they get downloaded from different locations.

In the future we should probably create a hidden "icon cache" component. JS would request an icon from the cache, and the cache would only retrieve it if it hadn't been retrieved already (cached).

IanMayo (approved):

lilitkarapetyan:
Screenshot (116)
Hello Ian, in both code blocks we use <AssetIcon_ /> component to show svg (pictures). In case of information marker icons fetch works 2 times, as because we have 2 svg there.

PR #1738, opened 2022-08-07
1696: Delete assets when Platform-Type removed

### LINKED ISSUE #1696: Delete assets when Platform-Type removed
In `Small Wargame` we have platform-types, and assets.  Assets have a (non-editable) platform-type.

If we delete a platform-type that is used by one or more assets the "Forces - Assets" table falls over - since the platform isn't found.

So, when the user requests to delete a platform-type we should:
1. Count how many assets are of that type
2. Warn the player that "4 Assets of this type will be also be deleted"
3. If the player confirms, delete the assets as well as the platform type

lebaphi:
Fixes #1696 

https://user-images.githubusercontent.com/26705621/183272118-9c503c38-e8f9-4ee4-bcc6-1cf3b79a3657.mp4


IanMayo:
One DeepScan issue please @lebaphi :
image

packages/client/src/Components/Modals/DeleteConfirmation.tsx
```diff
@@ -1,34 +1,61 @@
...
+        }
+      })
+      return result
+    }, [] as PlatformTypeCount[])
+  }
+
+  useEffect(() => {
+    if (type === 'platformType') {
+      const platformTypeUsed = countPlatformTypeUsed()
+      if (platformTypeUsed) {
```
IanMayo, line 56: Shouldn't this be:
`if (platformTypeUsed.length === 0)` ?

IanMayo: Aah no, I misinterpreted the logic.

IanMayo (approved):
Re-approve

PR #1781, opened 2022-08-13
1721: Refactor Admin Channel

### LINKED ISSUE #1721: Refactor Channels Admin page (modularize)
The Channels Admin Page code is quite complex, since it contains custom editors for different types of channel.

We should _modularise_ it - by pulling custom implementations into their own file.

The benefit of this, is that only files with code for the particular channel type would be edited in one PR, and it makes the file(s) easier to understand.

_Ideally_ we would define an interface that channel-type editors have to implement, then let the channel editors provide suitable instances.

I _guess_ the interface would include:
* custom config panels
* convert data to row
* convert row to data
* row headings

lebaphi:
Fixes #1721


packages/components/src/local/organisms/setting-channels/types/props.d.ts
```diff
@@ -24,4 +24,32 @@ export type MessageGroupProps = {
   value: string[]
 }
```
IanMayo, line 26: This is a minor suggestion @lebaphi - but how about declaring each props in the type for that channel?  So, `CollabChannelProps` would be in `collab.tsx`?

IanMayo:
@lebaphi - we have `createParticipant()` which contains code to determine the channel type, then create the correct type of participant.  But, it is called from our "type-specific" classes - so we don't actually need to use "createParticipant" - we can let the type-specific classes (`chat.tsx`) create their own participants.

Here's what I'm trying to do with the refactoring.  I'm trying to reduce the number of modules that deal with all 4 types of channel.   So, in this instance I'd like to avoid `createParticipant` having to handle each channel type.  I'd like to move the functionality into `chat.tsx`, `mapping.tsx`, etc, etc.

packages/components/src/local/organisms/setting-channels/index.tsx
```diff
@@ -66,809 +33,101 @@ export const SettingChannels: React.FC<PropTypes> = ({
...
-
-  /** init data for collab panel controls */
-  const messagesValues = getMessagesValues(isCollab, selectedChannelState)
-  const [messageLocal, setMessageLocal] = useState<MessagesValues>(messagesValues)
-
-  useEffect(() => {
-    /** on changes channel, update the message data local */
-    setMessageLocal(messagesValues)
-  }, [selectedChannelState])
+  const isChat = isChatChannel(selectedChannelState)
```
IanMayo: hello @lebaphi - we have these 4 `isChat` style booleans.  But, the booleans are only used in the `renderChannelContent` method.  Instead, we could just use a `switch` inside `renderChannelContent` - and check the `channelType` value.

IanMayo:
I was about to try to do a bit of the refactoring for you, but StoryBook failing for the component:
image

packages/components/src/local/organisms/setting-channels/types/props.d.ts
IanMayo: Hmm, maybe that's not possible.  I'll look at the code if/when I get chance.

Aah, hold on. The props files would be in the `channels` folder.

lebaphi: I'm not clear the purpose of this @IanMayo ?

IanMayo: Sure - happy to explain :-)

This is the idea..  We have 4 types of channel,  one is mapping.  If we make changes to the mapping channel configs, I hope that only `mapping` specific files would get edited.   Then we _know_ that we can't accidentally break another type of channel. Currently, the `mapping` props are in the same file as the other channel types.  So, adding a new mapping prop, and "tidying up the file" could mean that we accidentally change the props for another channel type.

lebaphi:
> @lebaphi - we have `createParticipant()` which contains code to determine the channel type, then create the correct type of participant. But, it is called from our "type-specific" classes - so we don't actually need to use "createParticipant" - we can let the type-specific classes (`chat.tsx`) create their own participants.
> 
> Here's what I'm trying to do with the refactoring. I'm trying to reduce the number of modules that deal with all 4 types of channel. So, in this instance I'd like to avoid `createParticipant` having to handle each channel type. I'd like to move the functionality into `chat.tsx`, `mapping.tsx`, etc, etc.

updated

lebaphi:
I've updated @IanMayo

IanMayo:
> I was about to try to do a bit of the refactoring for you, but StoryBook failing for the component: image

Note: I only get the above issue in `Storybook`. The editor works fine in `develop`.  But, we do need it fixed in `storybook`.

lebaphi:
> > I was about to try to do a bit of the refactoring for you, but StoryBook failing for the component: image
> 
> 
> 
> Note: I only get the above issue in `Storybook`. The editor works fine in `develop`.  But, we do need it fixed in `storybook`.

I fixed it already

packages/components/src/local/organisms/setting-channels/types/props.d.ts
lebaphi: ah I see what you mean. Yep, the mapping type is in the same file with other channel types but it's a separate type object, we can change it without effects to others. But we can move it into the channel file. I will update

IanMayo:
Thanks for the fix @lebaphi. The story doesn't crash  - but we can't select a new story - the value of `selectedChannelState.channelType` is `undefined`

packages/components/src/local/organisms/setting-channels/types/props.d.ts
IanMayo: You'll see that because I am crazy, I have tried to use generics to declare the props.  To do this I had to switch from `type` to `interfaces`.  This doesn't seem to introduce any compiler warnings, and things seem to work ok.

It would still be _neat_ to declare the types in channel-specific files, if we can.  Aah, the `props` for `CustomChannelProps` are only accessed from `custom.tsx` - so I guess we could just declare it in `custom.tsx`.

lebaphi:
> but we can't select a new story

Do you mean this issue @IanMayo ? I don't think it's an issue of the channel setting, I also got the same error with another story 🤔 

image

IanMayo:
No, @lebaphi , not that one. On each new build of storybook I get that issue (technical debt).

This is the issue: In current storybook I can't select a channel. The UI reports 'legacy channel', but actually it's looking at an 'undefined' value.

lebaphi:
> No, @lebaphi , not that one. On each new build of storybook I get that issue (technical debt).
> 
> This is the issue: In current storybook I can't select a channel. The UI reports 'legacy channel', but actually it's looking at an 'undefined' value.

okay, let me check this

IanMayo:
Here is what is rendered in the story:
image

It's related to how we handle state in the panel.

lebaphi:
Hi @IanMayo , I fixed the storybook issue

IanMayo:
Hmm, @lebaphi - shouldn't we see the "Constraints" panel for the mapping channel?

If VHF16 is `chat` then we should have the panel with `hide message authors` toggle?  Hmm, it's ok in the app, but the controls are missing from Storybook@

image

lebaphi:
yep, we should handle some callbacks function in the story class, it's not an issue of the setting channel.

IanMayo (approved):
Great, thanks.

PR #1782, opened 2022-08-14
1747: Add UI for Support Panel

### LINKED ISSUE #1747: A. Create Support Panel
Create child component that extends the mock `ChannelPlanning` developed in #1745 with following properties:
- Has tabs (similar to WorldState panel)
- Just have single "mock" tab to start with
- Occupies full height of page
- Drag bar to change width
- Clicking on tab collapses panel
- Has these props:
    - player role
    - player force
    - messages for this channel
    - current `forces` from wargame

Note: I've created a `PlanningChannel`, which itself contains a `PlanningMessagesList`.  This `PlanningChannel` will get extended with the above requirements, with `PlanningMessagesList` being the initial `My Orders` tab.

Hmm, I wonder if it's worth using `flexlayout-react` for the vertical tabs?  We're already using that component, after all.  The tabs won't need to be editable, but I think the implementation will be simpler than the way we did it with WorldState.

Here is a mockup:
image

lebaphi:
Fixes #1747


IanMayo:
Thanks @lebaphi  - looking good :-)

lebaphi:
https://user-images.githubusercontent.com/26705621/184802367-95e05c68-7325-4f39-8c53-2826ba5d1d41.mp4

IanMayo:
We're nearly there @lebaphi 

Just one thing: could we have a visual clue/indicator that the panel can be resized.  Maybe a pair of "grab-bars" (or handle-bars) to the left of the 3 tabs?

lebaphi:
How about this @IanMayo 😄 

image

IanMayo:
I think we need something a bit more visible, please.

lebaphi:
How about this one @IanMayo ?

image

IanMayo:
yes, great.

IanMayo:
This is weird @lebaphi - I get some "artefacts" on the new tabs.  On initial renders there are a couple of light blue marks. When I expand the panel, the marks grow:
image

I guess it's not your code.  Is it the new component?

I'm on Chrome 104

lebaphi:
It's weird, I don't see this issue. I use this lib to support drag and resize https://www.npmjs.com/package/react-rnd, can you try this on firefox or safari?

IanMayo:
Sorry, only have ChromeOS, on my Chromebook. Let's skip it.

IanMayo (approved):
Some graphic artifacts on ChromeOS, but will solve those if they appear on other operating systems as separate issue.

PR #1785, opened 2022-08-28
1749: Add Planning Assets component

### LINKED ISSUE #1749: D. Create PlanningAssets component
Create a component that shows a table of assets.

A `renderer` method (passed in as prop) will take an `Asset` object and return a `string[]` of table cells.

Props (all from state):
* ForceData[] (source data)
* PlayerForce - display data for that force
* isUmpire - if true, included drop-down for umpire to choose `view-as` force

The component should receive a `forces` object, `player-force` value and `renderer` method in `props`.

If `player-force` is provided, only show assets of that force.  If no `player-force` provided, show all assets.  The renderer method will either return the real values (when looking at own force) or `perceived` values (when looking at other forces).

Use this table component:
https://github.com/mbrn/material-table

Where an asset `comprises` other assets, the `children` assets should be grouped beneath the parent.

Test data is present in `p9.mock.ts`, developed in support of #1745 

Have attempted to mockup of how a player may see "own" and "other" assets:
image

lebaphi:
Fixes #1749 

Screen Shot 2022-08-28 at 11 21 24 AM


IanMayo:
Thanks for this first version @lebaphi 

Could you also implement the "comprising" of assets using the Material Table "tree" functionality please?
https://material-table.com/#/docs/features/tree-data

An example of this is how `CTF 511` _comprises_ two other assets in `p9.mock.ts`.

lebaphi:
> Thanks for this first version @lebaphi
> 
> Could you also implement the "comprising" of assets using the Material Table "tree" functionality please? https://material-table.com/#/docs/features/tree-data
> 
> An example of this is how `CTF 511` _comprises_ two other assets in `p9.mock.ts`.

Hi @IanMayo , I'm working on it 💪

lebaphi:
@IanMayo 👇 

https://user-images.githubusercontent.com/26705621/187644982-c19a5c2a-b21b-4968-aee3-7a2c01e1949a.mov

IanMayo:
[update]. while I did the changes on the wrong branch, I've sucessfully moved them over.

Tomorrow I'm going to work on these bits:
* [x] Only show the current player's _perception_ of the assets in the `OpFor` table
* [x] Collate the data (class names) necessary to show an icon of the asset

Once I've done the second one, I'll be pushing it back to you, please - so we can display an icon.  I presume I've just got to provide a field with the correct values, and you'll provide a custom renderer that displays the icon.

IanMayo (approved):
It's in a good enough condition for us to move on

PR #1786, opened 2022-08-29
1716 - track player activity

### LINKED ISSUE #1716: Extend player-activity tracking with messages-read (including markers)
The clients want to start tracking when players have read messages, for analysis once a game is complete.

We currently have "player activity" - where Serge remembers the most recent interaction, and sends that to the server every 20 seconds.

We store the most recent event for each player on the server, in the `playerLog`.

I'd like to keep a persistent record of player interactions, but without affecting the Serge wargame.

So, I'd like us to make these changes:

1. modify `setActivityTime` to append the event to a list of pending activities (instead of just storing it)
2.  introduce new event tracking for `marking a message as read` (including per-message tracking in `mark all as read`). Also track the player clicking on an `information marker` (since that's when they effectively read the contents of the message).
3. modify `pingServer` to push all of the events to the server as ,`{event: millis-since-epoch}` (rather than just the most recent one), then flush the list
4. modify the server so that on `healthcheck` receipt, push the json to a text file in a `logs` sub-folder (creating it if necessary).  The file should be named `wargame`-`role`-`date-time`.`json`.
5. Aah, we should be able to switch this off, so we would have a new `Log player interactions` control in the `Overview` admin page, and in the `WargameOverview`structure.
6. introduce new server API call 'GET /$wargame$/logs` to get logs for a wargame.  It will collect all log files for that wargame, zip them up, and return them.

### Worries
1. I don't quite remember why we're storing the activity in `expiredItem` as a cookie, and not just keep the item(s) in memory
2. the current `healthcheck` call sends a tiny amount of data in the URL.  We could be sending a lot more data in the call - so I guess it may have to be a `put` instead of a `get`.  **Lot more data**: a player may be looking at a channel with a hundred messages, and presses `Mark all as read`. This would be hundreds of data entities.
3. Can we `intercept` the browser tab closing, or being reset?  If we can, it would be good to append a `quitting` event to the activity log, and pushing a `healtcheck` call.

lilitkarapetyan:
## 🧰 Issue

Fixes #1716 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


packages/client/src/api/wargames_api/index.ts
```diff
@@ -152,42 +150,50 @@ export const listenForWargameChanges = (name: string, dispatch: PlayerUiDispatch
...
-    .then((response: Response): Promise<any> => response.json())
-    .then((data: any) => {
-      return data
-    })
+  const activityTime = activityTimes || activityMissing
+  const activityType = activityTypes || activityMissing
+  
+  if (!wargame) return null
+  
+  const { db } = getWargameDbByName(wargame)
```
IanMayo, line 160: @lilitkarapetyan  - I think you had a mistake in the above line.  I think you were trying to get `player-logs`, but that's not a wargame database

packages/client/src/api/wargames_api/index.ts
```diff
@@ -152,42 +150,50 @@ export const listenForWargameChanges = (name: string, dispatch: PlayerUiDispatch
...
+      } else {
+        const newPlayerlog: ActivityLogsInterface = {
+          ...dbDefaultPlaylogSettings,
+          wargame: wargame,
+          role: role,
+          activityType: activityType, 
+          activityTime: activityTime
+        }
+        
+        db.put(newPlayerlog)
```
IanMayo: It _looks like_ we're pushing the new log to the wargame here.  In the backend, we should be pushing the entries to the `player-logs` database.  But, I don't know if the server is _invisibly_ handling this.

packages/custom-types/player-ui.d.ts
```diff
@@ -73,6 +73,7 @@ export default interface PlayerUi {
   /** whether access codes are displayed for current wargame */
   showAccessCodes: boolean,
   /** whether logged in user can view insights & feedback */
```
IanMayo, line 77: @lilitkarapetyan - we need to fix the jsdoc for the new property.  This toggle is used elsewhere in Serge to indicate if this player is allowed to see the button that opens the player-log-history.   If we have a new property, we'll need to give it a different name/

packages/custom-types/player-log.d.ts
```diff
@@ -20,6 +24,43 @@ export interface PlayerLogUI {
...
+  aType: typeof MAP_ANNOTATION_READ_INTERACTION
+  _id: string
+}
+
+/** union of all types of activity */
+export type PlayerInteraction = PlainInteraction | MessageReadInteraction | MessageUnReadInteraction | MapAnnotationReadInteraction
+
+export interface PlayerLogEntry {
+  activityTime: string
+  activityType: PlayerInteraction
```
IanMayo, line 68: @lilit - instead of `activityType` as a string, we now have it as an object that can take its own metadata (such as the id of the message that has been read).

IanMayo:
Hello @lilitkarapetyan  - I've made progress with this in the past couple of hours.  It was failing when It tried to retrieve the `player-logs` from the `wargameDB` array.  I've fixed that, and introduced some stronger typing for different types of interaction.

But, Serge is now failing when it tries to send the periodic `server-ping`. Well, I think it's sending the ping, but I think the data is going into the wargame, instead of into the `player-logs` structure.  This means there's something other than a message or an infoType in the wargame documents, and SErge is falling over.

Good luck with making further progress :-)

**Note:** the jest snapshots are failing in 7 places. But, they don't fail locally.  Hmmm. I think it's a `yarn install` dependency. Yes, I think that worked ok.

* [x] Change frontend so that `pingServer` pushes (possibly empty [*]) array of documents to backend
* [x] Change PouchDb backend to rename SQLite `playerlogs` to `playerlogs.sqlite` - so it opens more easily
* [x] Change backend(s) so they push array of player logs (if present) to the `playerlogs` database, rather than into the wargame
* [x] Fix the heartbeat in the `<Version/>` component.

The `pingServer` checks the server is alive, in addition to sending activity.  So, if no activity is present, we should still ping the server, with an empty array, as a way of confirming the server is alive.

packages/client/src/api/db/types.d.ts
```diff
@@ -1,13 +1,14 @@
...
+import { Message, MessageInfoType, Wargame, ActivityLogsInterface } from '@serge/custom-types'
 import DbProvider from '.'
 
 export interface DbProviderInterface {
   changes: (listener: (doc: Message | Wargame) => void) => void
   destroy: () => void
   get: (query: string) => Promise<Wargame | Message | { status: number }>
-  put: (doc: Wargame | Message) => Promise<Wargame | Message>
+  put: (doc: Wargame | Message | ActivityLogsInterface) => Promise<Wargame | Message | ActivityLogsInterface>
   allDocs: () => Promise<Message[]>
```
IanMayo, line 9: @lilitkarapetyan - I think we need a different `put` event signature here, that takes an array of ActivityLogsInterface messages.  Or, I guess we could declare a new type that represents an array of ActivityLogsInterface items.

packages/client/src/Views/PlayerLog/index.tsx
```diff
@@ -41,11 +44,18 @@ const PlayerLogComponent: React.FC<PlayerLogProps> = ({ isOpen, onClose }): Reac
     }
   }, [isOpen, playerMessageLog])
 
+  const selectedForceId = selectedForce ? selectedForce.uniqid : ''
+  
   const collatePlayerLogData = (messageLog: PlayerMessageLog): void => {
-    getPlayerActivityLogs().then((activityLog: ActivityLogsInterface[]) => {
+    getPlayerActivityLogs(currentWargame, currentDbname).then((activityLog) => {
```
IanMayo, line 50: @lilitkarapetyan - this method is "putting" the player activity logs, rather than getting them. Could you rename it please?

packages/client/src/api/wargames_api/index.ts
```diff
@@ -152,46 +151,64 @@ export const listenForWargameChanges = (name: string, dispatch: PlayerUiDispatch
...
-  })
-    .then((response: Response): Promise<any> => response.json())
-    .then((data: any) => {
-      return data
-    })
-    .catch((err) => {
-      console.log(err)
+  const { wargame, role, activityTime, activityType } = first
+
+  return await getPlayerActivityLogs(wargame, log.currentDbname)
```
IanMayo: @lilitkarapetyan - this code appends the logs to the existing set for this wargame.

We don't need to do that.  We can just push all the log items to the `player-log` database.  The server API currently just allows one document to be pushed - so we just need to either switch it to taking an array (and insert them as individual docs), or to use a "bulk update" PouchDb API call to push/store all the docs at once.

lilitkarapetyan: Ok, I will work on it

packages/client/src/api/db/types.d.ts
IanMayo: Aah, @lilitkarapetyan - there is a _knock-on_ change from the above.  

We will no longer have a tidy list of activity for the current wargame.  Instead the `player-logs` database will have one lone list of activities.  So, we _may_ need to do some server-side processing to filter the activity for the specified wargame, to only include the newest entry for each role.

packages/client/src/api/wargames_api/index.ts
IanMayo: (seem to have lost this comment)

@lilitkarapetyan - we're making these changes because a scientist wants to download the full history of activities for a wargame - not just the most recent one.  So, on the client we are building up an array of activities, and pushing them all to the server on `pingServer`.   That means `player-log` will grow to have all activites for all roles for all wargames.

That's a lot more data.  It also means that when we retrieve the player logs for a wargame, she server needs to only provide the most recent activity (newest `id`) for each role.

packages/client/src/api/wargames_api/index.ts
```diff
@@ -169,7 +169,7 @@ export const pingServer2 = async (log: ActivityLogsInterface, logAllActivity: bo
   // In addition to pushing data to the server, we're also checking the server is still alive
   // So, even if the log is empty, we should push an empty list, since still we want to get a 
   // 'success' back from the server
-  return db.putPlayerLogs(items).then(res => res)
+  return db.putPlayerLogs(items).then(res => res.msg)
 }
```
IanMayo, line 173: I've almost certainly misunderstood this @lilitkarapetyan 

In the above line, we return `res.msg`.  But, `putPlayerLogs` only returns a `Promise<string>`.  Should it return a structure?

Aah, I do think the `<Version>` component was failing to complete `pingServer` properly, since it always stayed grey. I see it's working now. Is that related to this fix?

IanMayo:
@lilitkarapetyan  - I've put the read/unread marker into the message cell, like this:
image

That means we can simplify the structure used to represent the table.  Is that ok?

lilitkarapetyan:
> @lilitkarapetyan - I've put the read/unread marker into the message cell, like this: image
> 
> That means we can simplify the structure used to represent the table. Is that ok?

Yes it's good

IanMayo (approved):
Also checked on CouchDb

PR #1812, opened 2022-09-02
1809: View icons in assets panel

### LINKED ISSUE #1809: View icons in assets panel
I've merged the `assets` PR.

The code in this PR collates the data ready for the icon to be displayed:
image

I believe we can now change the `renderer` for the `icon` column so that we display an `asset-icon`:

image

lebaphi:
Fixes #1809 

image


IanMayo:
Thanks for this @lebaphi, I have one issue.  If we put a debug line in `getRows()` we see it gets called repeatedly as the Support Panel is resized.

We have already wrapped `getColumns` and `getRows` in `useState`.  I tried switching to `useMemo`, but it didn't help.  I also tried creating `playerForceState` and `forcesState` objects - and only running an update if the new value was different to them.  But, it didn't help.

Could you think about how to reduce the rendering please?

Note: one fallback could be to generate the data in `support panel`.  Hmm, if I move the `useState` to `SupportPanel` - it doesn't get re-generated on each resize activity.

Aaah, I _guess_ we could cache the generated cols/rows in `Planning-Assets`, and only call `setColumns`, `setRows` if the cached values have changed.  But, it _looks-like_ `Planning-Assets` state is all lost on each call.

IanMayo (approved):
I'm going to merge this. We can solve improvements as new issues.

PR #1819, opened 2022-09-03
1810: Support detail for orders panel

### LINKED ISSUE #1810: Support detail for orders panel
The `Orders` documents can be quite complex.  While we display a summary for the table view, players need to be able to expand the item to view all of the detail.

So, we need to use the `detailPanel` capability of `material table`.  Once we've got the detail to display some plain old message fields, we'll move onto using `JSON-Editor` to render the detail.

Hmm, there's a chance the JSON-configuration will look different for different templates.  If that's the case, I guess we may need to return the contents of a method, rather than plain HTML.

lebaphi:
Fixes #1810 

Hi @IanMayo , the sample JsonEditor was added into the detailPanel

https://user-images.githubusercontent.com/26705621/188258639-713d1721-49c4-461c-82c4-9f5fed3e2fed.mp4


IanMayo:
Brilliant, thanks @lebaphi  

It did seem a "gamble" on whether we could integrate the JSON-Editor.

I've re-started the build, since it failed on a mock wargame.  If it's still broken later, I'll try to fix it.

IanMayo:
Hmm, I think it may because I've provided JSON instead of JS object.

lebaphi:
> Hmm, I think it may because I've provided JSON instead of JS object.

I think these test files are the same as the develop branch, but the build passed on develop but not on remote branch 🤔

IanMayo:
Ah, got the build issue @lebaphi .

It's in this file:
https://github.com/serge-web/serge/pull/1819/files#diff-9189b5a1ffe9eea03af4dd6db285b44b497defd58ed97e48736c0b5a20d083b5

That `index.tsx` is being built for production, but it pulls in `mocks` data. We only refer to `mocks` data in `.spec` files.  I'll sort out the `plumbing` to get the data from the parent.

IanMayo (approved):

PR #1820, opened 2022-09-04
1811: Map placeholder in background of support panel

### LINKED ISSUE #1811: Map placeholder in background of support panel
Eventually the `PlanningChannel` will have a `LeafletJS` map in the background.

But, that's not until the next phase.

So, for now we should just put a static image on the background, so we can start learning about it:
map image

Aaah, could we have a copy of that image on our repo?  It will be useful to demonstrate what things look like on a network without an internet connection.

The SupportPanel should _float over_ the map image.

lebaphi:
Fixes #1811 

https://user-images.githubusercontent.com/26705621/188295414-b1ed950c-e846-4064-80f6-e039fa30d095.mp4

Hi @IanMayo , the image size does not fit the screen but I tried to make it fit, that's why we see the background image looks a bit skewed


IanMayo (approved):
works fine

PR #1821, opened 2022-09-04
1816: Improve rendering of Planning Assets component

### LINKED ISSUE #1816: Improve rendering of `Planning Assets` component.
I have one issue with the `PlanningAssets` component. If we put a debug line in getRows() we see it gets called repeatedly as the Support Panel is resized.

We have already wrapped getColumns and getRows in useState. I tried switching to useMemo, but it didn't help. I also tried creating playerForceState and forcesState objects - and only running an update if the new value was different to them. But, it didn't help.

Could you think about how to reduce the rendering please?

Note: one fallback could be to generate the data in support panel. Hmm, if I move the useState to SupportPanel - it doesn't get re-generated on each resize activity.

Aaah, I guess we could cache the generated cols/rows in Planning-Assets, and only call setColumns, setRows if the cached values have changed. But, it looks-like Planning-Assets state is all lost on each call.

Note 2: the current re-rendering causes the icon images to be repeatedly downloaded, too.  If the rendering-related fixes in this issue don't avoid repeatedly downloading the icons, then we should consider some icon cacheing (note: we're using SVG icons, so they're actually just text snippets). This was also covered in #1717

lebaphi:
Fixes #1816

### DIFF
--- a/packages/components/src/local/organisms/support-panel/index.tsx
+++ b/packages/components/src/local/organisms/support-panel/index.tsx
@@ -1,9 +1,9 @@
 import Slide from '@material-ui/core/Slide'
 import MoreVert from '@material-ui/icons/MoreVert'
-import { PlatformStyle, forceColors, platformIcons, ForceStyle } from '@serge/helpers'
+import { forceColors, ForceStyle, platformIcons, PlatformStyle } from '@serge/helpers'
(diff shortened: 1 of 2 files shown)


IanMayo (approved):
Much simpler than I was expecting :-)

PR #1822, opened 2022-09-06
1709: Time Remaining Indicator is missing

### LINKED ISSUE #1709: Time Remaining Indicator is missing
I've just realised that we aren't currently displaying the time remaining for a turn.

image

The wargame that ran at the end of last year didn't use SERGE timing/steps, so I hid it. I clearly forgot to reinstate it.

I can switch it back on, but it distorts the panel. At least it's working 🤣 

gif

The time remaining helps players schedule their planning process - ensuring adjudication can happen on track.

The value counts down to zero in planning (which has fixed length), but increases up from zero in adjudication (since adjudication can take a greatly varying amount of time depending on the complexity of the game at that time).

The current "Turn Progression" panel looks great, but we also need to find a location for the time-remaining value. _Sorry_ 😉

lebaphi:
Fixes #1709 

Screen Shot 2022-09-06 at 6 53 03 PM


IanMayo:
Thanks @lebaphi - one issue. The "time remaining" should be a countdown counter.  I guess you may have to go back to previous versions of `TurnProgression` to find the code that supported that.

ShendMikullovci:
Looks good to me!

IanMayo (approved):
There was a bunch of dodgy legacy that got fixed in here, too.

packages/components/src/local/molecules/turn-progression/index.tsx
```diff
@@ -87,8 +87,9 @@ export const TurnProgression: React.FC<Props> = (props: Props) => {
   }
 
   const countup = (): any => {
+    const startTime = Math.round(new Date(adjudicationStartTime).getTime() / 1000)
     const now = Math.floor(new Date().getTime() / 1000)
-    const seconds = now - progressionState.startTime
```
IanMayo, line 88: This was a strange issue, @lebaphi  & @lilitkarapetyan 

`progressionState` is the state of this form.  But, here inside this timer callback, it wasn't receiving the updated state. I _think_ that's because the timer callback gets defined when the timer is initiated, and that version of the function gets `cached`.  So, I had to generate the data I needed using form props, rather than form-state.

Things worked in the end.

PR #1828, opened 2022-09-08
1814 turn markers

### LINKED ISSUE #1598: Notification Counter Responsive to Own Messages in General Chat
I think it'd be best if the notification counter only responded to others messages to reduce cognitive load on users.
It feels redundant and counter-intuitive to have notifications for your own messages. 

https://user-images.githubusercontent.com/85491794/173696283-cb8f56b3-ce66-45b0-9abd-2c1c692a5df3.mp4

Side-note: I noticed that hitting enter sends you to a new line instead of sending the message, forcing the user to click on send by mouse. Is this necessary? What is the nature of the messages to be sent here? If the messages won't be long and complex, I'd suggest having Enter send the message.

### LINKED ISSUE #1663: New messages shown at top in all channels
I invited a power user to give feedback on chat vs non-chat channels, and got this:

> I think it's probably worth harmonising at least in part - having one be most recent at top, and one most recent at bottom, could be counterintuitive - I suspect most would assume the newest message would appear at the bottom, like a text message chain.

Here is example of text message chain growing from the bottom:
image

> Otherwise, I don't think it's problematic to have the chat channels showing the new message box constantly, it makes sense to have it available for free text but as a selectable option for template channels. I could go either way on that one, it would probably be down to how users felt about it!

Currently, chat channels display new messages at the top, and non-chat channels display new messages at the bottom.

This is inconsistent. New messages should appear at the bottom in all channels.

Hmm.  If a player doesn't have a message expanded, _maybe_ the channel should scroll down to the new message.

### LINKED ISSUE #1799: Notification Counter On Own Messages
The notification counter shouldn't be activated by one's own messages, only other's:

Notification Counter

### LINKED ISSUE #1814: Turn Markers missing from Chat Channel
This game is on Turn 6, and I believe the message were sent across multiple turns.

So, there should be turn markers in the channel.
image

IanMayo:
Fixes #1814
Fixes #1663 
Fixes #1799
Fixes #1598

### DIFF
--- a/packages/client/src/ActionsAndReducers/playerUi/helpers/handleWargameMessagesChange.ts
+++ b/packages/client/src/ActionsAndReducers/playerUi/helpers/handleWargameMessagesChange.ts
@@ -39,7 +39,7 @@ export const handleNewMessage = (payload: MessageChannel, newState: PlayerUi): S
   // TODO: only one of `payload` or `newState` will have been received. We should have 
   // two different handlers, one for each change.
   const res: SetWargameMessage = handleNewMessageData(payload, newState.channels, newState.chatChannel,
-    newState.playerMessageLog)
+    newState.playerMessageLog, newState.selectedRole)
   return res
 }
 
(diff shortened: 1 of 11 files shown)


ShendMikullovci (dismissed):
Tested on a couple of wargames. Seems to be good.

PR #1836, opened 2022-09-10
1795: Configuration Option Interferes with Channel Label

### LINKED ISSUE #1795: Configuration Option Interferes with Channel Label
https://user-images.githubusercontent.com/85491794/187749960-4e10e7f0-c4bc-4a42-b3f4-c957d63a3e97.mp4

Clicking on Hide Message Authors seems to reset the label of the chat channel.

lebaphi:
Fixes #1795

### DIFF
--- a/packages/components/src/local/organisms/setting-channels/index.tsx
+++ b/packages/components/src/local/organisms/setting-channels/index.tsx
@@ -64,42 +64,48 @@ export const SettingChannels: React.FC<PropTypes> = ({
   }
 
   const renderChannelContent = useMemo(() => {
-    const channelType = (selectedChannelState && selectedChannelState.channelType)
-    switch (channelType) {
-      case CHANNEL_COLLAB:
-        return <CollabChannel
-          channel={selectedChannelState as ChannelCollab}
-          forces={forces}
-          messageTemplates={messageTemplates}
-          onChange={onChannelDataChange}
-        />
-
-      case CHANNEL_CUSTOM:
-        return <CustomChannel
-          channel={selectedChannelState as ChannelCustom}
-          forces={forces}
-          messageTemplates={messageTemplates}
-          onChange={onChannelDataChange}
-        />
-
-      case CHANNEL_MAPPING:
-        return <MappingChannel
-          channel={selectedChannelState as ChannelMapping}
-          forces={forces}
-          onChange={onChannelDataChange}
-        />
-
-      case CHANNEL_CHAT:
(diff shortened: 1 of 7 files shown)


packages/components/src/local/organisms/setting-channels/index.tsx
```diff
@@ -166,14 +178,21 @@ export const SettingChannels: React.FC<PropTypes> = ({
   return (
     <AdminContent>
       <LeftSide>
+        <Dialog
+          isOpen={deleteSelectedChannel}
+          header='Error'
+          content="You can't delete selecting channel"
```
IanMayo: `selected`

But, does this prevent users deleting the one last channel?

@lebaphi - What's the reason for not being able to delete the selected channel?

lebaphi: Hi @IanMayo ,
> But, does this prevent users deleting the one last channel?

yep it will prevent users deleting the last channel, I will update the fix

> What's the reason for not being able to delete the selected channel?

I found the problem is that when we delete the selected channel, the search list is updated with the new channels but the data in the right panel is not, it still shows the deleted channel. We can fix this but it requires more work to debug/refactor the code. I saw some apps with similar behavior (preventing user from deleting selected item) so I decided to use this solution 🤣

lebaphi: Hi @IanMayo, so what approach do you want to take?

lebaphi:
Hi @IanMayo, I removed the new solution we discussed above, now it's back to the old approach and we can delete the selected channel as well, and of course with my fix 😄

IanMayo (approved):
Minor improvements made in review

IanMayo:
Thanks @lebaphi 

Does this commit look ok?  I've realised I'm checking if a channel is present in two different ways. Do you think one is better than the other?

https://github.com/serge-web/serge/pull/1836/commits/eb5e8c2ef872cf2cbcde8a87bfa4e13c2638c73d

lebaphi:
I think it is not necessary to check the selected channel value before the switch case, let the switch case handle it in the default block.

IanMayo:
> I think it is not necessary to check the selected channel value before the switch case, let the switch case handle it in the default block.

Great, thanks @lebaphi .  I did want to provide a different message if there isn't a channel present - but I've done it as another `case` in the switch:
https://github.com/serge-web/serge/pull/1837/files

PR #1837, opened 2022-09-12
Don't handle missing channel in separate block

IanMayo:
Minor code improvement, following @lebaphi  review.

### DIFF
--- a/packages/components/src/local/organisms/setting-channels/index.tsx
+++ b/packages/components/src/local/organisms/setting-channels/index.tsx
@@ -64,46 +64,40 @@ export const SettingChannels: React.FC<PropTypes> = ({
   }
 
   const renderChannelContent = useMemo(() => {
-    if (selectedChannelState === undefined) {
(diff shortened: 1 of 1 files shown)


packages/components/src/local/organisms/setting-channels/index.tsx
```diff
@@ -64,46 +64,40 @@ export const SettingChannels: React.FC<PropTypes> = ({
...
+        />
+      case CHANNEL_CHAT:
+        return <ChatChannel
+          channel={selectedChannelState as ChannelChat}
+          forces={forces}
+          onChange={onChannelDataChange}
+        />
+      case CHANNEL_PLANNING:
+        return <div>Editor not yet provided for planning channel. Waiting for data model to mature. Channel:<br />{JSON.stringify(selectedChannelState)}</div>
+      case undefined:
```
lebaphi, line 97: I think we should create a constant for it, something like CHANNELS_EMPTY

PR #1851, opened 2022-09-15
1841 Table Improvement

### LINKED ISSUE #1841: Table improvements
We still don't have business requirements for the `SupportPanel`.

Instead, I'd like us to carry on with technical features.
* [ ] remove pagination
* [ ] The `map` will show the assets currently shown in the `my assets` table.  So, we need a `listener` for the table that gets called with a list of `uniqid`s for the  currently visible assets in the table. (in the future we'll do the same for `orders`).  We currently only have the free-text-search in the table, but when entering a phrase like `frigate` the handler should be called with the ids of the matching rows.  Note: it looks like we need to listen for `onSearchChange`.
* [ ] Also allow the assets table to take a `selected` callback. This will pass the `uniqid` of a asset row when selected (and '' when no row is selected).
* [ ] Map mistakenly shown in table footer:  image

lebaphi:
Fixes #1841


IanMayo:
Thanks for the @lebaphi  Today I've been playing with trying to get hold of the currently visible set of rows, but hadn't managed what you did.

In the next phase of the project we'll introduce a map, and show counters on the map for all assets currently shown in the `ownForces` and `opForces` tables.  That's why `SupportPanel` will have a list of `opForces` and `ownForces` in state.

I've had a play. You'll see in my commit [here](https://github.com/serge-web/serge/pull/1851/commits/07182878ee3859233bf3b9bb4c1dbc2220a8dc19#diff-c6aa963ec22d0ec4980b9e74e5753fe92558b1909a041147337c2f4e988e5963R37) that I'd like to keep track of which rows are visible in each table (one for own/friendly forces, and one for opposition/unfriendly) forces.   But, if I try to store the state (`setOwnForces`) in the callback, I get into an infinite loop and break the call stack.   I _think_ the callback needs to test if the new `data` is different to the existing data, and only update when different (I did this via `_.isEqual()`) - but setting state goes into another infinite loop.

Are you able to help out please?

The end results will be that we have `opForces` and `ownForces` state arrays that only get updated when the table filter changes.

lebaphi:
sure @IanMayo , I'm looking into it

packages/components/src/local/organisms/support-panel/index.tsx
```diff
@@ -71,71 +71,102 @@ export const SupportPanel: React.FC<PropTypes> = ({
...
-        </div>
-      </Slide>
+          </div>
+        </Rnd>
+        <TabPanelActions onChange={onTabChange} />
+      </div>
+    </Slide>
+  ),
+    [
+      isShowPanel,
```
IanMayo, line 150: thanks @lebaphi - can I reduce the items in this list, to reduce the data change that trigger a re-render?

lebaphi: Hi @IanMayo , yes you can.

IanMayo (approved):

PR #1856, opened 2022-09-19
1855: add collapsible component

### LINKED ISSUE #1855: Support Panel - create new orders
The Support Panel has an orders tab:
image

The tab currently shows a list of templates for that set of orders.  We should allow players to create new orders.  The "New orders" will be shown as a drop-down list, that the user clicks to indicate what type of document to create. ( Hopefully we can re-use the implementation in "New Message", now that it has been refactored):
image

See the `Blue & SME` tab here: https://serge-review.herokuapp.com/?wargame=wargame-kskl1i3t&access=pkskl17ti

lebaphi:
Fixes #1855


lebaphi:
Added Collapsible component @IanMayo 

https://user-images.githubusercontent.com/26705621/190978184-72fb1250-7e6c-46fe-9e30-c84e0246fc59.mov

So I'm wondering about the next step, we'll add a new Input component for the user to enter message => save when clicking the new order button, and the `message.details.messageType` should match one of the `templates` entries. ?

IanMayo:
@lebaphi  - at the end of the issue description is a link to a working example from a "Custom Channel". I'd like us to try to copy that implementation, please.

lebaphi:
https://user-images.githubusercontent.com/26705621/191046479-00975fa6-546d-441f-a0ee-9588e9a8f895.mp4

lebaphi:
Hi @IanMayo , how does it looks?

https://user-images.githubusercontent.com/26705621/191049674-5a0bfc62-bea8-4c6d-bed6-dd40845138cf.mov

IanMayo:
brilliant! Not working today, but dying to have a play with it!

packages/components/src/local/organisms/support-panel/types/props.d.ts
```diff
@@ -1,30 +1,28 @@
...
-  gameDate: string
   /** new orders templates for this player */
   templates: TemplateBody[]
   /** descriptions of platform types (used to generate icons) */
   platformTypes: PlatformTypeData[]
+  activityTimeChanel: (role: string, message: string) => void
+  saveMessage: (currentWargame: string, details: MessageDetails, message: any) => Function
+  saveNewActivityTimeMessage: (role: string, activity: PlainInteraction, dbName: string) => Function
+  dispatch: Dispatch<any>
+  state: PlayerUi
```
IanMayo: @lebaphi - I'm having a problem with the `story` loading the state data.  But, I think I'd rather drop the `state` from props, and pass in the specific parameters (for and role).
Do you see any problem with this?

lebaphi: it's okay @IanMayo, passing specific parameters will make the props list too long instead of putting them all in state

IanMayo (approved):
Seems to be working fine

packages/components/src/local/organisms/support-panel/types/props.d.ts
IanMayo: I had a look, and `PlayerUi` wasn't referred to from anywhere in the `package/components`. It _seemed_ right to continue that practice :-)

PR #1870, opened 2022-09-22
1859: SupportMapping component

### LINKED ISSUE #1859: Introduce LeafletJS map in background of Support Panel
Now that we've got assets in the Support Panel, and we're letting users filter them, we should start displaying them on a map.

The map will replace the static image.  It will use the `leaflet` map that we already for the `mapping` channel.  See the `mapping` component.  We won't be re-using the `mapping` component, because it contains lots of complexity - and we don't need all that functionality for `SupportPanel`.  But, `mapping` may be useful for seeing how it is done.

Once we have the `SupportMapping` component working, I'll create other issues, to let us put arrays of Assets onto it. `SupportPanel` will pass these asset arrays in response to `ownForces` and `opForces` state objects.

But, for this issue, we will just be creating the `SupportMapping` component, which will have a toolbar with zoom in & zoom out.

lebaphi:
Fixes #1859 

https://user-images.githubusercontent.com/26705621/191785626-2b6ad7b2-01b2-4fe1-b5b0-3f6a9c93ae48.mov


IanMayo:
Brilliant :-)

IanMayo:
Thanks for this @lebaphi 

When I'm playing with data, it's really convenient to be able to test things from a story - since it's so quick/easy to modify the mock data, or to use StoryBook controls.

It would be good if `SupportPanel` and `SupportMapping` were in the same story. Is it possible to switch `PlanningChannel` to be a `Component` like in our `packages\components` package?

packages/components/src/local/organisms/support-mapping/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+import React, { useState } from 'react'
+import { Map, ScaleControl, TileLayer } from 'react-leaflet'
+import { MapConstants } from './helper/MapConstants'
+import styles from './styles.module.scss'
+import PropTypes from './types/props'
+
+export const SupportMapping: React.FC<PropTypes> = ({ }) => {
+  const TileLayerProps = MapConstants.TileLayer
+
+  const [position] = useState<LatLngExpression>([51.505, -0.09])
```
IanMayo: hello @lebaphi - for now, could we move `position` and `zoom` to be props for `support-mapping`?

They can then be specified in the story/unit test.  It's (temporarily)  ok to hard-code them in `PlanningChannel`.  Eventually we'll calculate the bounds/zoom from the visible `assets` - then use those values.

IanMayo:
Aah, one more issue.  With any of the `SupportPanel` expanded, the map doesn't respond to drag/zoom.  With the SupportPanel collapsed, it is possible to drag/zoom

lebaphi:
I will update @IanMayo 👍

IanMayo:
very excited about this @lebaphi :-)

IanMayo:
I'm going to do some refactoring @lebaphi, to remove `state`.  This is the only component in `packages/components` that uses `PlayerUI`.   I know it will mean a _lot_ more props, but it's will be easier to generate the test data when I just need to think about the data that's required.

IanMayo (approved):
Working great, thanks!

PR #1875, opened 2022-09-24
1861: Show assets on map

### LINKED ISSUE #1861: Show assets on map
Our SupportPanel includes `ownForces` and `opForces` listings.   These contain a list of our own assets, and assets from other forces that we are aware of (or all for umpires).

We should create a `<PlanningAssets>` component that can render an array of assets.  It will put them onto a `LayerGroup` component (like we do with `<Assets>` ).

Note: I suggest that in `Phase 1` we display the raw icons, but once we are rendering them, we should do a `Phase 2` that clusters the markers, using this library: https://www.npmjs.com/package/react-leaflet-cluster

So, here is `Phase 1`:
* [x] introduce state in `PlanningChannel` for `ownForces` and `opForces`.  These will be set from `PlanningAssets`.  I _guess_ we need to pass a `setter` down to `PlanningAssets`
* [x] extend `SupportMapping` props to take two arrays of `AssetRow` items (`ownForces` and `opForces`)
* [x] extend `SupportMapping` to introduce new `PlanningForce` component - which will display an icon for each asset, using name, type, force data.  This will be (much) simpler version of the `Assets` component - just a leaflet `LayerGroup`  containing `Marker` items.  `SupportPanel` will have two instances of `PlanningForce` - one for `ownForce` and one for `opForces`.   We're putting the two sets of forces into different `LayerGroups` so we can switch them on and off, if we want to in the future.

I tried to do the first two steps here this afternoon, but got into an infinite loop of stack updates - with stack overflow. I'm sure you'll do better than me :-)

Note: in #1872 I made some early progress on this.  But note:  in that branch I'm plotting all of the assets.  In your _real_ code you will be plotting the `ownForces` and `opForces` - which will be fewer assets, and the `opForces` will have `perceived` values - as in their `AssetRow` objects.

lebaphi:
Fixes #1861 

https://user-images.githubusercontent.com/26705621/192095368-88637f41-0f4f-43eb-874c-13c414875660.mov


IanMayo:
Hello @lebaphi - is the icon missing from the Storybook build?  I see this for `PlanningChannel`:
image

Hold on, the marker icon is also missing from `yarn develop`: 
image

lebaphi:
sorry I forgot storybooks, let me update,the marker icon works on my machine for yarn develop, can you try to run serge in incognito? 😂

IanMayo:
(sorry - out for next couple of hours)

IanMayo:
Hmm - how about trying Phase-2 - the real asset icons?

IanMayo:
Hmm, that's strange.  In incognito I still don't see the markers.  From the network tab I can see that Chrome is downloading the icon:
image

I don't see them in Safari, either:
image

Hmm, the URK for the image source looks like it's missing a slash:
image

This image path returns a valid icon:
image

Aah, I think I've found it. For the `mapping` component, we include a `leaflet.css` in the component folder. I'm trying that for `SupportMapping`.

IanMayo (approved):
The fix works perfectly

PR #1884, opened 2022-09-26
1861: display assets icon (phase 2)

### LINKED ISSUE #1861: Show assets on map

lebaphi:
Fixes #1861 

https://user-images.githubusercontent.com/26705621/192245270-5e939bd5-28c6-42c7-afa4-2fe6c5d3273b.mov

about adding new lib https://www.npmjs.com/package/react-leaflet-cluster, I got an issue that it does not work with old version of react-leaflet, (we are using v2)

Screen Shot 2022-09-26 at 4 20 35 PM

Increase the leaflet version may break something


IanMayo:
> about adding new lib https://www.npmjs.com/package/react-leaflet-cluster, I got an issue that it does not work with old version of react-leaflet, (we are using v2)

oh dear. We may see a lot of that.  `leaflet.draw` looks like the no-brainer solution for our drawing component/. But, it hasn't had updates for over 4 years :-(

IanMayo (approved):
Hooray - looks brlliant

lebaphi:
> > about adding new lib https://www.npmjs.com/package/react-leaflet-cluster, I got an issue that it does not work with old version of react-leaflet, (we are using v2)
> 
> oh dear. We may see a lot of that. `leaflet.draw` looks like the no-brainer solution for our drawing component/. But, it hasn't had updates for over 4 years :-(

so should we upgrade the leaflet version to the latest? @IanMayo

IanMayo:
Hello @lebaphi  - sorry, there isn't an obvious answer.   Yes, let's update to the newest version.  We'll step back if we need to support some legacy component :-(

Me again :-)

It looks like the newest version of `react-leaflet` is using `leaflet 1.9`. It would probably make sense to update to that version.  The scary bits may happen when leaflet (eventually) moves to `2.x`.

PR #1899, opened 2022-09-27
1886: Select assets from the map

### LINKED ISSUE #1886: Select assets from the map
We can currently only select assets from the table.

Players should be able to select assets from the map, too.

Note: currently the `Own Assets` and `OpFor Assets` tables are multi-select.

We should support this on the map, via the `ctrl` key.  So, if the player clicks on an asset - it should become the new single selected item.  If they hold down the `ctrl` key, the asset should be added to the current selection.

Note: we will also have to modify the `Own Forces` and `Opp Forces` tables - so that their items get marked as selected with triggered from the map.

I _suspect_ that `SupportMapping` will have to be modified to introduce `setSelectedItems()` and `selectedItems` props.  These will be passed down to `PlanningForces` in order that the onClick` handler for the markers can either replace the selected items with the new one, or to add the new item to the list (`ctrl` held down).

The `selectedItems` and `setSelectedItems` will sit in `PlanningChannel` - replacing `selectedItem` and `setSelectedItem`.

lebaphi:
Fixes #1886 

https://user-images.githubusercontent.com/26705621/192509598-69cfb548-7ae8-47b9-882f-067e79e23d46.mov

### DIFF
--- a/packages/components/src/local/form-elements/private-chat-input-toggle/index.tsx
+++ b/packages/components/src/local/form-elements/private-chat-input-toggle/index.tsx
@@ -18,10 +18,10 @@ export const PrivateChatInputToggle: React.FC<Props> = React.forwardRef(({ place
   }, [])
 
   useImperativeHandle(ref, () => ({
-    setFormState(text: string): void {
+    setFormState (text: string): void {
       setFormState(text)
     },
-    clear(): void {
+    clear (): void {
       setFormState('')
     }
   }))
--- a/packages/components/src/local/organisms/adjudication-messages-list/index.spec.tsx
+++ b/packages/components/src/local/organisms/adjudication-messages-list/index.spec.tsx
@@ -46,7 +46,7 @@ describe('AdjudicationMessagesList component: ', () => {
     }]
 
     const tree = renderer
-      .create(<AdjudicationMessagesList forces={forces} setSelectedItems={(): any => noop} template={planningMessageTemplatesMock[0]} gameDate={P9Mock.data.overview.gameDate} channel={planningChannel} hideForcesInChannel={false}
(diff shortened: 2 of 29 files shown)


IanMayo:
@lebaphi - some compiler issues for you to fix. please:
image

IanMayo:
Hello @lebaphi 

Thanks for this. IMHO, we only need to put the marker at the centre of the map if we have selected it from the table.  If we've selected it on the map - we don't expect it to jump away when we click it.

So, I think we need a (slightly) separate handler for the MaterialTables. When they modified `selectedItems` we and there's only one item we change the map position.  But, if the handler has come from `SupportMapping` then we don't change the position.

I'm dying to get my hands dirty with your code - since it has now become quite familiar to me.  But, I'm onsite again tomorrow morning - so I won't be able to until tomorrow afternoon, at the earliest.

IanMayo:
@lebaphi  - I've seen that `selectedItems`  is being passed to `adjudication messages list`.

We have two concepts:  `Assets` (ships, aircraft, etc) and `Orders` (the Orders document is a set of instructions).  So, on `Own Forces` and `Op Forces` we have `Assets`, and on `My Orders` and `Adjudication` we have `Orders`.

I now realise that `selectedItems`  is too vague. We actually need two separate _State_ objects: `selectedAssets` and `selectedOrders`.  

So, could you switch this PR to `selectedAssets` please? 

Hmm, we should also introduce `selectedOrders` - which is the state object that is shown/modified in the `My Orders` and `Adjudication` tabs, please.

Aah, one other thing.  When I select an asset on the map,  I see the icons refresh.  I've just checked the `Network` tab - and I see we're re-downloading the SVGs.  We really need to cache them in the browser.  Could you fix that too, please?

PR #1903, opened 2022-09-29
1757 Save JSON-editor message

### LINKED ISSUE #1757: Contents of Chat message being written get lost on changing tab
I'm writing a chat message. I switched tab to the mapping tab, then back to chat - and my text had disappeared!

Hmm, I guess that the event that triggered the "wipe" shouldn't cause a refresh.

I don't think there's an easy solution.   12 months ago we worked hard on performance with channels that have lots of messages - so we don't render channels that aren't visible.

Before we did that, I guess the draft contents of a new message would still be there when the player tabs back.  But, I don't think we can return to rendering all tabs (visible and non-visible).

Hmm, so what are the possible solutions?
* for a hidden tab, don't render the message list, but do render the edit form?
* monitor form changes, and put copies of the form contents into browser-memory?
* monitor tab going invisible, and persist any open editors to memory. Restore them on tab coming visible?

***Update:*** - we need a similar fix to this for when JSON-editor is editing a document in a "Custom Channel".  An example of a custom channel is the `Blue & SME` channel in `PAPA FOXTROT` wargame, when logged in as `Blue - Senior Blue`.   For long templates, is it worth trying to store all the JSON in one token?

***Update 2:***. I _like_ the idea of putting the full message into one token.  I was worried about the length of data we an fit into expired storage.  But, it has just stored and restored a 85,000 character message. That's 24 pages in a word document.  I think we'll be ok :-). The _great_ news is that it's persistent between sessions, so if I refresh the browser, it still remembers the text :-). Yes, let's store the whole message (as JSON) in expired storage.

lilitkarapetyan:
Fixes #1757


IanMayo:
Thanks @lilitkarapetyan 

Two initial problems:
- I have a draft message open. I change tab. When I come back, the `new message` panel has collapsed, and when I open `new message` it is showing a different message type.  When I select the correct type of message, I do see the saved data
- The `private` text box isn't stored/restored.

packages/components/src/local/form-elements/message-creator/types/props.d.ts
```diff
@@ -10,13 +10,18 @@ export default interface PropTypes {
   selectedForce: ForceData
   selectedRole: string
   selectedRoleName: string
```
IanMayo, line 12: @lilitkarapetyan - could you include jsdoc comments for these new props, to explain what they do, please?

lilitkarapetyan:
> Thanks @lilitkarapetyan
> 
> Two initial problems:
> 
>     * I have a draft message open. I change tab. When I come back, the `new message` panel has collapsed, and when I open `new message` it is showing a different message type.  When I select the correct type of message, I do see the saved data
> 
>     * The `private` text box isn't stored/restored.
I will work on it now

packages/components/src/local/form-elements/message-creator/types/props.d.ts
lilitkarapetyan: Yes, I will include comments in that part of the code.

IanMayo:
Getting closer :-)

Two things still remain:
- when I change tabs and come back, the message editor is collapsed
- the `Private` field isn't captured

Note: I'm testing with this wargame/role: http://localhost:3000/?wargame=wargame-kskl1i3t&access=pkskl17ti

IanMayo:
Getting closer, thanks @lilitkarapetyan 

Two remaining issues:
- with a JSON-editor template open, when I switch to another tab, then come back, the message editor is collapsed.  When I open it the data is there, but the message-editor should still be open when I return
- I can open a new message in JSONEditor, and start filling in a message. Then I can `cancel` it.  That message should now be cleared from the cache. But, if I start to create a new message of that type, then the partially entered message is restored.  It shouldn't be.

lilitkarapetyan:
> 

Hello Ian, I will work on it now

> Getting closer, thanks @lilitkarapetyan
> 
> Two remaining issues:
> 
>     * with a JSON-editor template open, when I switch to another tab, then come back, the message editor is collapsed.  When I open it the data is there, but the message-editor should still be open when I return
> 
>     * I can open a new message in JSONEditor, and start filling in a message. Then I can `cancel` it.  That message should now be cleared from the cache. But, if I start to create a new message of that type, then the partially entered message is restored.  It shouldn't be.

IanMayo (approved):
tested and working

PR #1908, opened 2022-09-30
1888: Support Panel has to share horizontal space with leaflet map

### LINKED ISSUE #1888: Support Panel has to share horizontal space with leaflet map
Currently the `SupportPanel` sits above the map.

When we select an asset, leaflet `flies` us to the asset, putting in the centre of the screen.

But, when the support panel is extended past 1/2 way over the map, this puts the selected asset _underneath_ the support panel - and it can't be seen.

There may be a number of solutions to this, but I'd like us to start off by experimenting with sitting the Support Panel and that map in the same space. So, as the user expands the support panel - the map collapses.  In this way - all of the map is visible, and when a player selects and asset - the selected asset will still be positioned in the centre of the visible map.

lebaphi:
Fixes #1888 

https://user-images.githubusercontent.com/26705621/193188166-6620d81c-385a-46f7-b5ef-4dd1a7976013.mov


IanMayo:
Hello @philb89 - it's not quite right.   It _looks like_ the map control has been pushed to the right, and now sits underneath `Game Admin`.  When we select two assets - the map should zoom to show both.  But, the right-hand asset is under GAme Admin.  Hmm, no. That's not the case - since the toolbar and attributions are visible.

Do you have any thoughts @lebaphi ?

https://i.gyazo.com/104c23b7e078ac00ae2b60a42773e1f9.mp4

Aah, maybe this is preventing the map from knowing its new bounds?
image

IanMayo:
(have fixed conflicts. pushing now)

lebaphi:
> Hello @philb89 - it's not quite right. It _looks like_ the map control has been pushed to the right, and now sits underneath `Game Admin`. When we select two assets - the map should zoom to show both. But, the right-hand asset is under GAme Admin. Hmm, no. That's not the case - since the toolbar and attributions are visible.
> 
> Do you have any thoughts @lebaphi ?
> 
> https://i.gyazo.com/104c23b7e078ac00ae2b60a42773e1f9.mp4
> 
> Aah, maybe this is preventing the map from knowing its new bounds? image

I think it's a bug, the selected items should be centered on the map, I will check it :D

lebaphi:
Fixed @IanMayo 😄

IanMayo (approved):

PR #1913, opened 2022-09-30
Hotfix geometries in orders

### LINKED ISSUE #1863: Map display of `Orders`
Many `Orders` will include a spatial component. This will be a GeoJSON object of type `point`, `polyline` or `polygon`.  We will display shapes/features for these on the map.

So, we will create a `PlanningOrders` component that displays the currently filtered set of orders as shapes.  The component will display the data in a new `LayerGroup`.  The shapes will have an outline color according the current force.  We _may_ introduce some shading that is sensitive to the `Type` of orders (to be determined).

We may need to use the `leaflet.deflate` library to reduce very small shapes to just markers.

### LINKED ISSUE #1912: Refactor view-as into its own component
We should extend `MapControl` so that it takes children

Then, refactor the `view as` functionality into its own control, that returns a block of items (one per force).  In the future, this can be extended to use a popup menu when there are lots of forces.

Other controls can be passed in, too (apply filters, new orders item)

image

IanMayo:
Fixes #1863 
Fixes #1912 

I started off by just generating the geometry data, but then I moved on, to rendering too.


packages/components/src/local/organisms/map-planning-orders/styles.module.scss
```diff
@@ -0,0 +1,31 @@
+@import "../../../../../themes/dependencies/variables";
+
+.selected {
```
IanMayo: @IanMayo  - these styles need deleting. Also check for the other new components.

packages/components/src/local/organisms/planning-channel/index.tsx
```diff
@@ -219,18 +243,29 @@ export const PlanningChannel: React.FC<PropTypes> = ({
...
-        setSelectedAssets={setSelectedAssets}
-        forces={allForces}
-        viewAsCallback={setViewAsForce}
-        viewAsForce={viewAsForce}
         actionItems={mapActionItems}
         actionCallback={mapActionCallback}
-      />
+        toolbarChildren={
+          <>
+            <div className={cx('leaflet-control')} data-tour="filter-applied">
```
IanMayo: Generate an `ApplyFilter` component for this

packages/components/src/local/organisms/view-as/styles.module.scss
```diff
@@ -0,0 +1,31 @@
+@import "../../../../../themes/dependencies/variables";
+
```
IanMayo: ditch these styles

packages/components/src/local/organisms/view-as/types/props.d.ts
```diff
@@ -0,0 +1,8 @@
+export default interface PropTypes {
+    // current set of forces
```
IanMayo, line 2: stop these being optional.  If we're going to have a `view-as` component, then we'll have the data

IanMayo (commented):
some changes

PR #1917, opened 2022-10-03
1910: Adopt MapContainer

### LINKED ISSUE #1910: Adopt MapContainer
We're currently creating the map using `react-leaflet` `Map` control.  But, if we use `MapContainer` then we will be able to use `useMap()` hook in child components.

Plus, if we pass `children` into the map control using props, then we can create functionality outside the map container class.

So, step-1 is to switch `SupportMapping` to use `MapContainer`.

Step-2 is to move `SupportMapping` to `Map` - and remove any functionality that is specific to `PlanningChannel`.  To do this, we'll have to pass the two `LayerGroup` components in as `children`.

lebaphi:
Fixes #1910


lebaphi:
it seems I need to resolve some conflicts

IanMayo:
This is a big old PR @lebaphi  - Good piece of work.

But, I have a compiler error.  I've tried deleting the `build` folders, but I get this:
image

IanMayo:
> it seems I need to resolve some conflicts

don't worry about the conflicts - I can solve them.  Your best "value" is by solving other issues, thanks :-)

I do need you to sort the compile issue before I can review it properly though.

IanMayo (commented):
Ok, I _think_ I've fixed the conflicts.  But I'm a bit worried about `map` in `SupportMapping`.  It doesn't appear to have the functions expected in a map object.  I've worked around the problem - but I think it's a sign that `useMap()` isn't giving us the object we're expecting.

lebaphi:
> > it seems I need to resolve some conflicts
> 
> don't worry about the conflicts - I can solve them. Your best "value" is by solving other issues, thanks :-)
> 
> I do need you to sort the compile issue before I can review it properly though.



> Ok, I _think_ I've fixed the conflicts. But I'm a bit worried about `map` in `SupportMapping`. It doesn't appear to have the functions expected in a map object. I've worked around the problem - but I think it's a sign that `useMap()` isn't giving us the object we're expecting.

strange, let me check it

IanMayo:
For the local `compile` issue I have, Heroku has it too:
[build failure.txt](https://github.com/serge-web/serge/files/9699413/build.failure.txt)

IanMayo:
Thanks for the test fix @lebaphi - but for Heroku and I, we still have the compiler issue.

Here is the line of code:
image

Here are the two errors:
image

Each occurs in the word after the `type` control word.

Hmm, in `components\package.json` we specify `3.7.5` for `typescript`.   It _looks like_ the above `type` import [arrived with](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html) `3.8.0`

I've tried updating the `typescript` in all the `package.json` files to `3.9.1`, but that hasn't fixed it :-)

IanMayo:
Nearly there @lebaphi 

Just a strange rectangle behind the icons:
image

Hold on. I get that with `story`, but not `develop`.  It would be _nice_ to have it working in `storybook`.

IanMayo:
Hello @lebaphi - just one last item on this - the rectangles around the icons.

Are you able to get it done, so we can move you onto other issues?

lebaphi:
> Hello @lebaphi - just one last item on this - the rectangles around the icons.
> 
> Are you able to get it done, so we can move you onto other issues?

sure @IanMayo

lebaphi:
Fixes @IanMayo

IanMayo (approved):
I'm amazed how simple the map coordinates are :-)

PR #1922, opened 2022-10-09
1862: Provide drawing tools for new orders

### LINKED ISSUE #1862: Provide drawing tools for new orders
In the `Planning Channel`, players will trigger the creation of some `Orders` objects by selecting the type of new orders, then be lead through drawing a feature on the map.  Once the feature has been drawn the relevant `Orders` template will open, with the spatial data filled.

We'll use `leaflet.draw` for this feature.

For this we'll need these capabilities:
1. Leaflet map, with toolbar
2. Popup menu on toolbar, to select one of `A. Transit`, `B. Point Effect`,  and `C. Area Effect`.  These will trigger the drawing of `a. Polyline`, `b. Marker`, `c. Polygon`.  
3. Once the user has finished creating the feature, a popup dialog will open, letting them choose one of `Maritime`, `Land`, `Air` and `Other`.  Then, create and open the relevant template.  Note: the new feature will be stored in a `leaflet.draw` layer.  Once the template has been saved, the feature will be moved from the `leaflet.draw` layer to the existing `Orders` layer.
4. Users should be able to edit the spatial data for selected orders documents. So, we'll need a special `Edit` icon next to the `Spatial` field in the editor.  When this is clicked, the feature will be moved from the `Orders` layer to the `leaflet.draw` layer, and put into edit mode - to let the player drag the corners/nodes

**Note:** `leaflet.draw hasn't been updated for years, and may not work with leaflet 2.0.   leaflet-geoman looks like it is better maintained: https://geoman.io/leaflet-geoman

**Update:** I've made some progress in this in `map-draw-activity`, but have now got stuck.  I tried to introduce `react-leaflet-geoman-v2` via `yarn add` in `packages/components`.  It compiled ok, but threw an error when running `storybook`:
image

I could with some support in adding the `<FeatureGroup><GeomanControls>` element as demonstrated here:
https://github.com/TurtIeSocks/react-leaflet-geoman.   I guess we would have to position it at the "bottomright" instead, since our existing toolbar is at the top right.

I'm pretty sure that will go into `PlanningChannel`

lebaphi:
Fixes #1862 

https://user-images.githubusercontent.com/26705621/194742934-e245e73e-31c5-4c10-947c-79f45a21a78c.mov


IanMayo:
Brilliant!  But @lebaphi - could we have arrows spread along the line too?  So, if the player is zoomed in, they still see the direction of the line.

IanMayo:
Aaah, one other thing, please @lebaphi .  I know the JS you provided for polylinedecorator will get called in the end, but could you have a go at a `react` version of it?  So, there will be a `<LayerGroup>` with a `<PolylineDecorator>` (and `<Polyline>` if necessary) inside it?  It will have a set of points that come from a `state` object.

This could go in`<PlanningChannel>` - since that's where other `<LayerGroup>` objects are defined.

lebaphi:
let me check this 😃

IanMayo:
It's this code that I meant, @lebaphi 
image

lebaphi:
> It's this code that I meant, @lebaphi 
> 
> image
> 
> 

sure @IanMayo

lebaphi:
Hi @IanMayo , how does it look 😄  

https://user-images.githubusercontent.com/26705621/194764190-e3f21dc0-39d6-41d5-943e-0737549a1090.mov

IanMayo:
Build failing because of JS Heap size:
image

Tomorrow morning I can try hiding some more stories, and see if that fixes it. Any other ideas @lebaphi

IanMayo:
> Hi @IanMayo , how does it look 😄

It's great.   We will render lines in two places:
1. In geoman, when the line is being drawn
2. In Planning Map Orders - when all orders are being shown.

I _did_ think we only needed arrows for `2.`, but I now realise it would be useful in `1.` too.   I _think_ that it's in `2.` that we'll be declaring them as a react component.

lebaphi:
> Build failing because of JS Heap size: image
> 
> Tomorrow morning I can try hiding some more stories, and see if that fixes it. Any other ideas @lebaphi

I think we can increase the head size by this command `export NODE_OPTIONS=--max_old_space_size=8192`

lebaphi:
> > Hi @IanMayo , how does it look 😄
> 
> It's great. We will render lines in two places:
> 
> 1. In geoman, when the line is being drawn
> 2. In Planning Map Orders - when all orders are being shown.
> 
> I _did_ think we only needed arrows for `2.`, but I now realise it would be useful in `1.` too. I _think_ that it's in `2.` that we'll be declaring them as a react component.

actually, we need to get polyline reference to decorate it (add arrow) but we are using geoman lib to draw line so I don't get polyline's reference yet. My solution is to listen for the onCreate event when drawing the line in the geoman lib, capture the set of latlngs then remove the polyline layer (generated by geoman), hen let's redraw it with our react component.

I have created PolylineDecorator component, add set of latlangs then It will draw line with arrow.

IanMayo:
This is great, thanks @lebaphi 

You've implemented enough logic for us to "drive" the new feature.  I'm going to replace it with the "real" business logic. But, I'm going to merge the PR first, so that the new drawing editor is in the `develop` branch.

IanMayo:
Aah, one question. You've had to make this fix in a few places.  Can you explain why you had to do it please?
image

IanMayo:
Hmm, @lebaphi 

I made a change to put arrows along the line:
image


But, when I try to push the change, I get lots of lint errors.  I think you can see them if you run `yarn lint`.  Would you mind fixing them please?

lebaphi:
> Aah, one question. You've had to make this fix in a few places. Can you explain why you had to do it please? image

When we run the tests with jest, there are some nested components that need to be imported before running, but in the test environment they are undefined so the tests fail. So we need to mock these components and that's how we mock them.

lebaphi:
> Hmm, @lebaphi
> 
> I made a change to put arrows along the line: image
> 
> But, when I try to push the change, I get lots of lint errors. I think you can see them if you run `yarn lint`. Would you mind fixing them please?

sure @IanMayo

lebaphi:
Hi @IanMayo, I fixed all the lints and there are a lot of file changes, so do you think we should create a new PR or keep these changes in this PR?

IanMayo:
> Hi @IanMayo, I fixed all the lints and there are a lot of file changes, so do you think we should create a new PR or keep these changes in this PR?

Yes - the separate PR was a good call. Thanks :-)

IanMayo (approved):
The logic/implementation may change, but this successfully integrates the drawing tool

PR #1923, opened 2022-10-11
1862: Fix lints for Drawing Tool branch

lebaphi:


IanMayo:
Thanks @lebaphi - it looks like we need to use a special "trick" to sort the id of the elements:
image

IanMayo (approved):
Have commented out failing new-message snapshot

PR #1935, opened 2022-10-17
Dummy PR, to check Heroku

IanMayo:
We're getting lots of builds failing on Heroku.  They were intermittent, but are now happening more often.

This is a branch from a commit from a few weeks ago (16th September), to see if the build problem is because of something we have introduced.


IanMayo:
Ok. 16th Sept builds.  Getting Heroku to deploy again, just to test.

IanMayo:
Ok, this built several times.  That's interesting, but not useful in fixing current build.  That's being addressed in #1936

PR #1952, opened 2022-10-22
1934: Change labels on GeoMan tooltip

### LINKED ISSUE #1934: Change labels on GeoMan tooltip
When we enter a GeoMan drawing mode, a tooltip guides the user in what to do (add line, add polygon point, etc): `Click to place first vertex`

But, we do have domain-specific terms for the geometry/feature that the user is adding, such as `Route-in` or `Target Area`.  In `OrderDrawing`, it's in the `current.name` field, as manipulated in the `currentGeometry` useState hood.

GeoMan does include sets of translations that can be used to over-ride the displayed text.  I _think_ we just create a new custom translation that just introduces the term for the current `activity`.   The translation is created when the control-bar is created, but that's ok, because we re-generate it for each new activity element, eg: "route-in", "search area", "route-out".

animation of create feature

But, there may be another way of modifying the displayed text.

lebaphi:
Fixes #1934

### DIFF
--- a/packages/components/src/local/organisms/planning-channel/helpers/CustomTooltip.ts
+++ b/packages/components/src/local/organisms/planning-channel/helpers/CustomTooltip.ts
@@ -1,0 +1,35 @@
+export const CustomTooltip = {
+  tooltips: {
+    placeMarker: "Click to place marker",
+    firstVertex: "Click to start",
+    continueLine: "Click to continue drawing",
+    finishLine: "Click any existing marker to finish",
+    finishPoly: "Click first marker to finish",
+    finishRect: "Click to finish",
+    startCircle: "Click to place circle center",
+    finishCircle: "Click to finish circle",
+    placeCircleMarker: "Click to place circle marker",
+    placeText: "Click to place text"
+  },
+  actions: {
+    finish: "Finish",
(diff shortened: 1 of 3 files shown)


lebaphi:
Hi @IanMayo , I just created a concept on how to customize the tooltip text, so for this requirement: we recreate it for each new active element e.g. "route-in", "search" area", "route-out", I think we can create multiple translation files, then we can load the appropriate translation file for specific activity name.

IanMayo:
>  I think we can create multiple translation files, then we can load the appropriate translation file for specific activity name.

Yes, that was the only work-around I could think of.  But, we can't hard-code the titles, since they come from the game data.   That doesn't seem to be a problem though - we just clone the object and over-ride the specific command we're about to use.

IanMayo:
@lebaphi - I _Think_ you would clone  the translations in around this line:
https://github.com/serge-web/serge/blob/develop/packages/components/src/local/organisms/planning-channel/helpers/OrderDrawing.tsx#L102
In the relevant part of the `switch` block you would override the relevant part of the cloned translations, using the name from the `current` object.

IanMayo (approved):

PR #1953, opened 2022-10-23
1933: Issue - orders not getting down to PlanningMessagesList

### LINKED ISSUE #1933: Issue - orders not getting down to `PlanningMessagesList`
In the `SupportPanel` component, we have a child `PlanningMessagesList` - which will show the list of planning messages (orders).

But, that object is always empty.   This _may_ be related to the fact that we only render this table when the tab is active, but I _think_ it should receive the data when it gets rendered.

Clearly there is something about `state` that I don't understand :-)

This will be `Done` when `messages` in `PlanningMessagesList` has a non-empty `messages` object.

The code can be tested from any of the `PlanningChannel` stories in `StoryBook`.

lebaphi:
Fixes #1933


IanMayo:
Thanks for this @lebaphi - can you help me out by explaining what was happening?

lebaphi:
> Thanks for this @lebaphi - can you help me out by explaining what was happening?

Hi @IanMayo, ideally we don't want to re-render the SupportPanel component if the props don't change, but we have a problem at the `areEqual` function`!isEqual(prevProps, nextProps) `, it shouldn't have `!`, since props is an object we should do a deep comparison, there are 2 ways to do this:
- using `isEqual(obj1, obj2)`: in lodash docs says it's deep comparison but I don't see this, the result returns false
- JSON.stringify (prevProps) === JSON.stringify (nextProps): it works, the component does not re-render if the props don't change.

IanMayo:
> Hi @IanMayo, ideally we don't want to re-render the SupportPanel component if the props don't change, but we have a problem at the `areEqual` function`!isEqual(prevProps, nextProps) `, it shouldn't have `!`, since props is an object we should do a deep comparison, there are 2 ways to do this:

Hmm, @lebaphi - is there a chance that we're modifying an immutable array object higher up in the code?  So, while the array contents changed, the array object itself stayed the same - so react didn't recognise the change.

@lilitkarapetyan - is that a problem that happens with react state hooks?

IanMayo (approved):

PR #1974, opened 2022-10-28
Hotfix display new message content

IanMayo:
Hotfix to display new orders in `NewMessage` panel, and to make existing orders editable.

### DIFF
--- a/packages/components/src/local/molecules/json-editor/index.tsx
+++ b/packages/components/src/local/molecules/json-editor/index.tsx
@@ -1,16 +1,16 @@
-import React, { useState, useRef, useLayoutEffect, useEffect } from 'react'
+import React, { useEffect, useLayoutEffect, useRef, useState } from 'react'
 
 /* Import Stylesheet */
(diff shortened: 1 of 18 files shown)


packages/components/src/local/molecules/json-editor/index.tsx
```diff
@@ -128,12 +131,15 @@ export const JsonEditor: React.FC<Props> = ({
     document.addEventListener('keydown', handleKeyDown)
 
     if (nextEditor) {
-      const messageJson = expiredStorage.getItem(genLocalStorageId())
+      // only retrieve from expired content if we haven't been provided with message content
```
IanMayo, line 134: @lilitkarapetyan - in the code below, I only try to retrieve the message from storage if there isn't a `messageContent`.  Does that look ok to you?

lilitkarapetyan: Hello Ian, I think it's good.

PR #1995, opened 2022-10-30
Prevent assets being clickable when an activity is being planned

### LINKED ISSUE #1976: Prevent assets being clickable when an activity is being planned
When the player is planning a route, the click-handling goes a bit screwy around existing items (assets or prior orders).

We either need to get an understanding of this, or maybe configure GeoMan to only `snap` to items it has drawn

Note: in `PlanningForce` we can set `snapIgnore={true}` for the `<Marker>` element.  That prevents snapping.  But, it still isn't possible to start a line on top of an existing asset.

Hmm, it needs another fix.

lebaphi:
Fixes #1976 

https://user-images.githubusercontent.com/26705621/198883266-843b5632-e8a5-4d18-aabf-a65cea2c0c81.mov

### DIFF
--- a/packages/components/src/local/organisms/planning-force/index.tsx
+++ b/packages/components/src/local/organisms/planning-force/index.tsx
@@ -48,7 +48,7 @@ export const PlanningForces: React.FC<PropTypes> = ({ assets, selectedAssets, se
         className: styles['map-icon']
       })
     }
-  }, [])
+  }, [selectedAssets])
 
   return <>
     {
--- a/packages/components/src/local/organisms/planning-force/index.tsx
+++ b/packages/components/src/local/organisms/planning-force/index.tsx
@@ -1,6 +1,6 @@
 import cx from 'classnames'
 import L, { LatLng, latLng } from 'leaflet'
-import React from 'react'
+import React, { useCallback } from 'react'
 import * as ReactDOMServer from 'react-dom/server'
 import { LayerGroup, Marker, Tooltip } from 'react-leaflet-v4'
 import AssetIcon from '../../asset-icon'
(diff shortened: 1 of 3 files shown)


lebaphi:
Now we can start drawing everywhere we want, even on the same place of marker, but the path is behind the marker, do you want to make a path on top of the marker? @IanMayo

packages/components/src/local/organisms/planning-force/index.tsx
```diff
@@ -31,26 +28,54 @@ export const PlanningForces: React.FC<PropTypes> = ({ assets, selectedAssets, se
...
+      key: `asset-icon-${index}`,
+      position: loc,
+      icon: L.divIcon({
+        iconSize: [30, 30],
+        html: getAssetIcon(asset, isSelected, false),
+        className: styles['map-icon']
+      })
+    }
+  }, [])
+
```
IanMayo, line 52: @lebaphi   - since the above `useCallback` doesn't have any props, it won't get called if `selectedAssets` changes.  Am I mistaken

(this is the first time I've  worked with `useCallback` - so I'm guessing what is happening)

Yes, I've just tested this.   If we select an asset from the `own forces` tab, the map zooms in, but the asset doesn't start flashing.

IanMayo:
Works perfecly

> Now we can start drawing everywhere we want, even on the same place of marker, but the path is behind the marker, do you want to make a path on top of the marker? @IanMayo

The line being drawn is in front of the marker in the StoryBook.  This looks like the correct behaviour:
image

packages/components/src/local/organisms/planning-force/index.tsx
lebaphi: ah I just want to memoized the object function, so we don't need to add any props 😄

IanMayo:
> the path is behind the marker

@lebaphi - I'm happy to ignore that issue for now.

packages/components/src/local/organisms/planning-force/index.tsx
IanMayo: Without `selectedAssets` as a prop for `useCallback` we do not regenerate the icon to show that some assets are selected:

https://user-images.githubusercontent.com/1108513/198992283-2ddb3468-9153-41a7-8292-2b62d1b8d0c3.mp4

By putting `selectedAssets` in for the `useCallback` we do show the pulses for the selected assets:
screencast

IanMayo (approved):
Solution works

PR #1999, opened 2022-11-01
Save and Edit documents in My Orders panel

### LINKED ISSUE #1975: Save and Edit documents in `My Orders` panel
In the `PlanningChannel` we have the `My Orders` tab.  When the rows in this table are expanded, we use JSON-editor to display the contents.

While the orders are initially show read-only `[*]`, the owner of each message (`message.details.from.roleId === playerRoleId`) should have an `Edit button`.  They can then make document edits and then `Save` or `Cancel` the edit.

`[*]` `<JsonEditor>` is currently _ignoring_ the `disabled` property. Bug in #1973

lilitkarapetyan:
## 🧰 Issue

Fixes #1975 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations

### DIFF
--- a/packages/components/src/local/molecules/json-editor/index.tsx
+++ b/packages/components/src/local/molecules/json-editor/index.tsx
@@ -35,8 +35,7 @@ export const JsonEditor: React.FC<Props> = ({
   saveMessage,
   modifyForEdit,
   modifyForSave,
-  confirmCancel = false,
-  onCancel
+  confirmCancel = false
 }) => {
   const jsonEditorRef = useRef<HTMLDivElement>(null)
   const [editor, setEditor] = useState<Editor | null>(null)
@@ -83,18 +82,15 @@ export const JsonEditor: React.FC<Props> = ({
     setConfirmIsOpen(false)
   }
 
-  const onPopupConfirm = (event: React.MouseEvent<HTMLButtonElement>): void => {
+  const onPopupConfirm = (): void => {
     expiredStorage.removeItem(genLocalStorageId())
     initEditor()
     setConfirmIsOpen(false)
-    onCancel && onCancel(event)
   }
 
-  const openConfirmPopup = (event: React.MouseEvent<HTMLButtonElement>): void => {
+  const openConfirmPopup = (): void => {
     if (confirmCancel) {
       setConfirmIsOpen(true)
-    } else {
-      onCancel && onCancel(event)
     }
   }
 
--- a/packages/components/src/local/molecules/json-editor/index.tsx
+++ b/packages/components/src/local/molecules/json-editor/index.tsx
@@ -6,6 +6,7 @@ import 'bootstrap/dist/css/bootstrap.min.css'
 /* Import Types */
 import { Editor, TemplateBody } from '@serge/custom-types'
 import { configDateTimeLocal, usePrevious } from '@serge/helpers'
+import { Confirm } from '../../atoms/confirm'
 import Props from './types/props'
 
 import { expiredStorage } from '@serge/config'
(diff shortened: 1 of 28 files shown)


lilitkarapetyan:
Hello Ian, I'm still working on this PR.

packages/components/src/local/organisms/planning-messages-list/index.tsx
```diff
@@ -125,16 +104,39 @@ export const PlanningMessagesList: React.FC<PropTypes> = ({
...
+            const details: MessageDetails = {
+              channel: channel.uniqid,
+              from: {
+                force: selectedForce.name,
+                forceColor: selectedForce.color,
+                roleName: selectedRoleName,
+                roleId: playerRoleId,
+                iconURL: selectedForce.iconURL || selectedForce.icon || ''
+              },
+              messageType: template.title,
```
IanMayo: @lilitkarapetyan - I'd like use to store the template id in `messageType`, please.  That means Serge won't fail if the template title is changed.

packages/components/src/local/molecules/json-editor/index.tsx
```diff
@@ -190,18 +209,31 @@ export const JsonEditor: React.FC<Props> = ({
...
+      if (formId && !editButton) {
+        editor.disable()
+      } else if (disabled && !formId) {
         editor.disable()
       } else {
         editor.enable()
       }
     }
-  }, [editor])
+  }, [editor, editButton])
```
IanMayo, line 232: @lilitkarapetyan - this method refers to `formId`.  Should it include `formId` in the hook props?

lilitkarapetyan: Yes, because only components containing SaveMessageButton have a formId

IanMayo:
Hello @lilitkarapetyan - I've just been able to reproduce an issue I saw yesterday.

* yarn develop
* log in as `Mar-4`
* edit document owned by `Mar-4`
* change document `description` by adding a couple of words
* press `Save`
* data in table doesn't update

I checked the database, and the new record is in there.
I checked the console, and saw this:
image

So, it looks like we're receiving the new message twice, and ignoring the new version.

Note: I'm pretty sure I'm not just getting this on browser that does the `Save`. I think I get it on other logged in browsers, too.

Have you been able to successfully edit/save a document?

IanMayo:
Another issue, please @lilitkarapetyan 

When we save a document from `My orders`, it increments the `.message.Reference`.  

I'm pretty sure that in `checkReference` we should only find/generate a reference number if the `Reference` field is something like `null` or `''`.  I don't think we can check for `undefined`, because I think that would generate a `Reference` for templates that don't include a `Reference` field.

packages/helpers/src/tests/handle-initial-channel-messages.spec.ts
```diff
@@ -41,7 +41,7 @@ describe('handle initial channel creation', () => {
     // get the blue RFI channel
     const rfiChan = res.channels['channel-BlueRFI']
     expect(rfiChan).toBeTruthy()
```
IanMayo, line 43: thanks for the fixes @lilitkarapetyan 

_One day_ I look forward to `handle-initial-channel-messages` getting refactored & tidied up

packages/components/src/local/organisms/data-table/index.spec.tsx
```diff
@@ -20,6 +20,7 @@ const rfiData = newest.map((message: any) => {
     messageItem.details.channel,
     messageItem.details.from.roleName,
     messageItem.details.from.forceColor,
+    messageItem.details.from.forceId,
```
IanMayo, line 23: this one was a weird one to diagnose. Thanks!

IanMayo:
@lilitkarapetyan - there was an issue, where if we edit/save a message, the old and new instances of the message were displayed.  I've found the [fix](https://github.com/serge-web/serge/pull/1999/commits/f6376c0d8a5713d2a457d6cb64f40858a6ea37b2#diff-f97fc21549a839fe120adefe5de4fc716b655a6bcc52eb70d83868a04aaea726R42) for it, in `handle-channel-updates` - `handleNonInfoMessage`.  

The issue was because we're 1/2 through moving from `Reference` to `reference`.   Damn, I can't remember why I wanted to move to lower-case `reference`.  But, it's fortunate we did it, because currently `serge` is updating the `Reference` on every message sent, rather than only doing it for newly created messages.  I'll check we have that issue captured.

IanMayo (approved):

PR #2001, opened 2022-11-01
Selected Asset in Map should appear in the Panel

### LINKED ISSUE #1962: Selected Asset in Map should appear in the Panel
Once the user selects an asset through the panel on the left hand side, the screen adjusts and centers upon that asset, which is a great feature. I suggest we apply the same feature for a vice versa scenario: clicking on an asset in the map could automatically show the selected asset in the list, and scroll down until the selected asset is centered in the list or at least visible. 

The end result should show the asset clicked on the map and the centered list: 

Map SC 2

lebaphi:
Fixes #1962 

https://user-images.githubusercontent.com/26705621/199269124-4bd125dc-801e-4c99-ade4-71402aeba990.mov

### DIFF
--- a/packages/components/src/local/organisms/planning-assets/index.tsx
+++ b/packages/components/src/local/organisms/planning-assets/index.tsx
@@ -21,7 +21,8 @@ export const PlanningAssets: React.FC<PropTypes> = ({ assets, forces, playerForc
       const lastSelectedAssetId = selectedAssets[selectedAssets.length - 1]
       const elmRow = document.getElementById(lastSelectedAssetId)
       if (elmRow && !preventScroll.current) {
-        elmRow.scrollIntoView()
+        const smoothScroll: ScrollIntoViewOptions = {behavior: 'smooth'}
+        elmRow.scrollIntoView(smoothScroll)
       }
     }
     preventScroll.current = false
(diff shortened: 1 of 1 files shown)


packages/components/src/local/organisms/planning-assets/index.tsx
```diff
@@ -21,7 +21,8 @@ export const PlanningAssets: React.FC<PropTypes> = ({ assets, forces, playerForc
       const lastSelectedAssetId = selectedAssets[selectedAssets.length - 1]
       const elmRow = document.getElementById(lastSelectedAssetId)
       if (elmRow && !preventScroll.current) {
-        elmRow.scrollIntoView()
```
IanMayo: Hey @lebaphi  - how about this smooth scroll 😉

lebaphi: sure @IanMayo , let me updated 😄

IanMayo: Oh, @lebaphi  the highlight is on the wrong row. I've implemented the smooth scroll:
image

IanMayo (approved):
nice implementation

PR #2005, opened 2022-11-03
Move cancel drawing to the location of the new activity popup

### LINKED ISSUE #1964: Move `cancel drawing` to the location of the `new activity` popup
Once the user has selected an item from the `PlanningActivityMenu`, we do not need to show it.

So, we could pass `activityBeingPlanned` to the `PlanningActivityMenu`, and not show the popup control when something is being planned. Note: if we do that, 

When we are planning an activity, we do not need to show the `new activity` popup menu.

But, when we are planning an activity, we should show the `cancel planning` button in its place.

I tried this.  But, the `<item>` button wouldn't click while a line was being drawn.

lebaphi:
Fixes #1964 

https://user-images.githubusercontent.com/26705621/199728144-3dfb0a99-fc5b-4533-9d29-d7a56961e002.mov


IanMayo (approved):
Good solution, thanks.

PR #2008, opened 2022-11-04
Concept of turns in planning channel

### LINKED ISSUE #1982: Concept of turns in planning channel
The planning channel should `normally` only show orders for the current turn

But, it should be possible to view orders from the previous turn, with an adjudication outcomes from that turn.

Ok. Here is the UX:
* [x] At the top of "My Orders" tab insert a shallow panel. It will have a label: "Show orders for turn:" plus a drop-down of `Turn 1`, `Turn 2`, etc to `Turn 5 - Current`.
* [x] At the top of "Adjudication" tab insert a shallow panel.  It will have a label: "Show interactions for turn:" plus a drop-down of `Turn 1`, `Turn 2`, etc to `Turn 5 - Current`.

We then filter the `messages` in each tab according to the turn number.

lebaphi:
Fixes #1982

### DIFF
--- a/packages/components/src/local/organisms/support-panel/helpers/TurnFilter.tsx
+++ b/packages/components/src/local/organisms/support-panel/helpers/TurnFilter.tsx
@@ -1,5 +1,5 @@
 import { MenuItem, Select } from '@material-ui/core'
-import React, { ChangeEvent, useEffect, useState } from 'react'
+import React, { ChangeEvent, CSSProperties } from 'react'
 import styles from '../styles.module.scss'
 
 type TurnFilterProps = {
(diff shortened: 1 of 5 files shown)


lebaphi:
Hi @IanMayo , I've added the shallow panel on the top of these tabs, I'm not sure how to filter the message in each tab according to the turn number 🤔

IanMayo:
> Hi @IanMayo , I've added the shallow panel on the top of these tabs, I'm not sure how to filter the message in each tab according to the turn number 🤔

That's exactly what I was hoping you would do. I'll finish it off.

IanMayo:
Hello @lebaphi - I've implemented the filter handler.

Grr, I can't get things to work properly.  We're losing the state.   I had a couple of hours battling against it yesterday.  Could you step in to help out, please @lebaphi ?  When I choose a new turn from the menu, the control reverts to `show all turns`.

IanMayo (approved):
Working fine.

PR #2014, opened 2022-11-07
refactor how we customise templates

### LINKED ISSUE #2013: Refactor template customisers
We have a template-customiser to improve the drop-down lists in JSON-editor.

The method is steadily receiving more parameters.

Instead, we should declare an array of them, and work through the array when we need to _do-customise_.  The methods would be declared with their individual parameters, so they still match the `expected` API.

IanMayo:
Fixes #2013


packages/components/src/local/organisms/support-panel/index.tsx
```diff
@@ -156,6 +157,19 @@ export const SupportPanel: React.FC<PropTypes> = ({
     setTurnFilter(turn)
   }
```
IanMayo, line 159: @lebaphi - could you help review my JS please?

We did have a method that would take a `json-editor` template, and modify it (to build up the assets drop-down).

It was quite complex, with lots of params, and we need to add/modify some more drop-downs for it.  So,I've refactored it into a series (two so far) of methods that _effectively_ get called in a chain.

Could you check it makes sense please? Oh, and one last worry.  You see the `allOwnAssets` field in the code below?  Can you confirm that this should use the _current_ value of that state when the function gets called, and not the value of `allOwnAssets` when the page is first rendered?

PR #2026, opened 2022-11-09
2022 save geometry

### LINKED ISSUE #2022: Need to implement `Save Geometry`
We're allowing users to edit a geometry.  But, we still need to implement `Save`.

We `eventually` have to edit the live object in the `json-editor`.  But, it's significant that we're storing the GeoJSON in `hiddenStore`.  When users to `modifyForSave` it takes this GeoJSON and stores it.  If we _slide_ the edited value into there, we'll store the new geometry.

IanMayo:
Fixes #2022


IanMayo:
I have an issue with a callback. @lebaphi  is helping me.  Here are steps to reproduce:

https://user-images.githubusercontent.com/1108513/200845844-dbebd2ac-4b94-4df7-b677-a08ee0c7b35f.mp4

PR #2038, opened 2022-11-10
Dummy show zoom

IanMayo:
Trial build, to show current zoom level


IanMayo:
Close,  zoom value added in `develop`

PR #2075, opened 2022-11-13
Can't read read-only text-boxes and drop-downs because text is white

### LINKED ISSUE #2074: Can't read read-only text-boxes and drop-downs because text is white
In yarn-develop, when we create a new orders object, the disabled fields aren't legible.

image

Steps to reproduce:
- yarn develop
- log in as `mar-`
- create new orders for `Land Resupply`
- fill in line, polygon, line
- the `new orders` form will open (based on <NewMessage>)

lebaphi:
Fixes #2074 

Screen Shot 2022-11-13 at 9 57 33 PM

### DIFF
--- a/packages/themes/_forms.scss
+++ b/packages/themes/_forms.scss
@@ -59,6 +59,11 @@
       @include game-admin-input;
     }
   }
+
+  &:disabled,
+  &[readonly] {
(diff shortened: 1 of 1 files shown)


IanMayo (approved):
Good fix, thanks

PR #2120, opened 2022-11-15
2116 persistent edit orders

### LINKED ISSUE #2116: Editing current orders collapsed/reset when new orders submitted
We have some `tricks` in place to retain the contents of `json-editor` when there are unrelated changes in the channel.

When a new set of orders (in a new orders template) is being edited, we are able to replace the field contents on component refresh.

But, when an existing set of orders are being edited, the contents get lost when someone else from that force creates a new document in the channel.

We _could_ use a similar trick to retain content when an existing document is being edited.  But, I'd like us to consider another strategy.  How about we create the `json-editor` using a `createMemo()` call - that is only dependent on a object of state called `currentEditor`, that takes a document `id`.  On expand, we `setCurrentEditor(id)`, then create the memo, which only changes on changes to `currentEditor`. When the editor is saved or cancelled, we do `setCurrentEditor(undefined)` - and that's the only thing that changes the editor component.

I _think_ the memo would be on the `json-editor` component in `planning messages list`, and on `json-editor` in `message-creator`.

Video: https://i.gyazo.com/143802184fed6c5ac8c0ea7dba6cc106.mp4

IanMayo:
Attempts to fix #2116


packages/components/src/local/form-elements/message-creator/index.tsx
```diff
@@ -156,15 +167,21 @@ const MessageCreator: React.FC<PropTypes> = ({
...
     if (draftMessage) {
       const anyDraft = draftMessage as any
       if (anyDraft.message) {
-        setMessageContent(anyDraft.message)
+        // see if we have any cached content
+        if(formValue) {
+          setMessageContent(formValue)
+        } else {
+          setMessageContent(anyDraft.message)
+        }
```
lebaphi:
```
setMessageContent(formValue || anyDraft.message)
```

PR #2127, opened 2022-11-17
Provide JSON-Editor with component to view/render location data

### LINKED ISSUE #2006: Provide JSON-Editor with component to view/render location data
Our planning messages include `slugs` of JSON for the shapes(lines, polygons, etc).

JSON-Editor tries to build up a set of controls for this array data, but it  ends up with loads of ugly components:
image

I don't know the best solution, but this is the best I can think of:
- convert the JSON for `location` to a string
- specify in the template, that `location` is of `type`: `string`, but with a `format` of `location` [*]
- provide JSON-editor with a custom component for `format`: `location` entries
- display the geometries as `<ol>` `<li>` entries, showing the `properties.name`, and the `geometry.type`
- include an `edit` button, which will _eventually_ allow the player to edit the shapes

An example of providing a custom component to json-editor is given here: https://github.com/json-editor/json-editor/blob/8f2b5ec5ca5787b3cd1c6b83ee3ec2ef29aeb34c/src/editors/starrating.js

A story to develop this against is "ready to go" here: http://localhost:6611/?path=/story/local-molecules-jsoneditor--planning-message

[*] by doing this, json-editor no longer tries to create a tree of controls to represent the JSON data. I _think_ it also prepares JSON-Editor to use a custom `string` editor component.

Maybe an editor component like this:
image

lebaphi:
Fixes #2006


IanMayo:
Can I request a change of strategy please @lebaphi ?

I'd like us to keep the current workaround - in case I have to do a "rushed" demonstration.

This may be what you're doing, but I'd like the new component to gets used for `string` template objects with `"format": "location"`.

Aah, down in the details, if it's possible, I'd like the component to get re-rendered if anything changes in the form data.  The re-render function will receive a current copy of the form-data.

The reason for this is that:
- the location may show "route out", "do stuff", "route back".  If the player enters "Start-Date" and "End-Date" in the form, then I can calculate the actual start and end times of the 3 steps, and show them against them - to help the player understand when things are happening.

lebaphi:
it's just a draft PR @IanMayo, I will update later 😂

lebaphi:
Hi @IanMayo 

image

I think the new component works as the image above, we have the textarea with the Edit button. So I want to clarify some thing:

- _has function that can convert JSON content to human-readable:_ so we don't need the _modifyForEdit_ function to modify the location data, let the component do this.
- _has advanced function that can do lookup from JSON data in array of reference data:_ I'm unclear this
- _has "Edit location" button that can call method passing data, with callback function for results (or null for cancel):_ we will pass the form data in the callback function when edit button clicked? I see you mention to the cancel state, I guess when click on the Edit button, its label changes to Cancel, click on the Cancel will pass null in the callback function?

IanMayo:
> has function that can convert JSON content to human-readable: so we don't need the modifyForEdit function to modify the location data, let the component do this.

Exactly. `modifyForEdit` was a temporary workaround.  We will not need it for this new component.  We should _probably_ leave it in the API though, in case we need it for a future, unknown problem.

> has advanced function that can do lookup from JSON data in array of reference data: I'm unclear this

Ok, forget this.  I did think we need to allow the provision of a method that received the content and returned something human-readable.  But instead, I'll ensure the data object contains the human-readable contents - without `format()` helper. 

> has "Edit location" button that can call method passing data, with callback function for results (or null for cancel): we will pass the form data in the callback function when edit button clicked? I see you mention to the cancel state, I guess when click on the Edit button, its label changes to Cancel, click on the Cancel will pass null in the callback function?

Not quite.  It's like:
```
  edit: { data: PlannedGeometry[], callback: {(data: PlannedGeometry | undefined)} }
```

So. The component will receive a `edit` callback.  When the user clicks the `Edit` button, then `edit` is called, passing the current data, plus it's own callback function..   This opens GeoMan.  When the user clicks `Save` then `callback` is called with the new data.  If the user clicks `Cancel` then `callback` is called with `undefined`.

I hope this makes sense.

I welcome suggestions :-)

lebaphi:
https://user-images.githubusercontent.com/26705621/202726226-32bb2957-23c9-4c8b-917a-6a637c792096.mov

IanMayo:
Hi - the text area doesn't need to be editable.  So, it doesn't even need to be a text-area - if can be a div, I guess.   In the future I'd quite like to be able to include non-text stuff.   For the 3 `geometry` objects in the array for this location, we do (or will) have time in their properties.  So, I may want to show these 3 start/stop times on a small Gantt chart. To do that, we've got to move away from a text-area component.  All editing happens via the "Edit" button.

IanMayo:
/ok, @lebaphi - a bit more work here, please.  
* [x] the new `location-area` component shouldn't be a text-area, just a div/span
* [ ] I want to be able to transform the contents before they are rendered. So, _somehow_ I need to give a function to the `location-editor.ts` module. this callback will receive an array of `PlannedActivityGeometry` and return a string.  I want to specify this callback in `SupportPanel` - since that's where the supporting information is.

packages/components/src/local/molecules/json-editor/helpers/setupEditor.ts
```diff
@@ -21,36 +23,47 @@ const setupEditor = (editor: Editor | null, schema: any, ref: RefObject<HTMLDivE
...
+    //       editButton.name = 'editInLocation'
+    //       editButton.classList.add('btn', 'btn-secondary', 'json-editor-btn-add', 'json-editor-btntype-add')
+    //       editButton.style.height = '40px'
+    //       editButton.style.transform = 'translateX(-60px)'
+    //       editButton.addEventListener('click', () => editCallback())
+    //       locationSection.appendChild(editButton)
+    //     })
+    //   })
+
+    // test custom editor
```
IanMayo: at some point you'll have to delete this  test data please @lebaphi

lebaphi:
Hi @IanMayo,
>I want to be able to transform the contents before they are rendered. So, somehow I need to give a function to the location-editor.ts module. this callback will receive an array of PlannedActivityGeometry and return a string. I want to specify this callback in SupportPanel - since that's where the supporting information is.

Can we create a helper function to do this thing instead of add a function inside the location-editor component?

packages/components/src/local/molecules/json-editor/helpers/setupEditor.ts
lebaphi: so when I removed this block, we should update the schema to use format `location` and type `string` to use the new editor

IanMayo:
> Can we create a helper function to do this thing instead of add a function inside the location-editor component?

Hello @philb89 - sure, yes.   Let's pass in a helper function to `render` the contents of the `location-area` block.  I'd like it to return a `ReactElement` rather than plain text, so that we can include icons, if necessary, in the future.

If you create the function (that initially just returns the existing contents in a `<div>`, then I will modify it).  But, I'm pretty sure I need to create the function in `SupportPanel` - then the function gets passed down through the "tree" of components.

IanMayo:
Thanks @lebaphi - that looks exactly like I what I wanted :-)

IanMayo (approved):

packages/components/src/local/molecules/json-editor/helpers/setupEditor.ts
```diff
@@ -21,38 +23,40 @@ const setupEditor = (editor: Editor | null, schema: any, ref: RefObject<HTMLDivE
   const disableArrayDeleteLastRow = 'disable_array_delete_last_row'
   const promptBeforeDelete = 'prompt_before_delete'
```
IanMayo, line 24: @lebaphi - in these next sections, we should remove the un-used blocks, rather than comment them out, please.

PR #2132, opened 2022-11-18
Move `forces` into its own document in database

### LINKED ISSUE #2465: Move `forces` into its own document
In our system, we store game state in `infoMessage` documents.  The current game has 1000s of assets, and some very long templates.

The assets are in the `forces` structure.   During adjudication, the assets will change many times, as the state of the world changes.  Most of the other content will not change at all during the game.

This means we will have lots of copies of the wargame document, of which a lot will remain unchanged.

Instead, we should refactor the `forces` array into its own document type.  This will reduce the amount of times we store/distribute the static content for `templates`, `channels`, `activities`, `areas`, etc.

This change will affect these areas:
- loading wargames
- loading other game state
- storing updated wargames
- storing other game state
- maintaining game via admin pages

lilitkarapetyan:
Fixes #2465


lilitkarapetyan:
I'm still working on this PR

IanMayo:
Thanks for this @lilitkarapetyan - it looks good.  I have one question: will it work with a _legacy_ SQLite database, where the data is _all_ in the `wargame` document?   For the wargame that ran this week, I may still need to be able open/show the game state in Serge.

lilitkarapetyan:
Hello Ian, no it won't work with the old SQLite database, but can I make it work with the old database as well

IanMayo:
> Hello Ian, no it won't work with the old SQLite database, but can I make it work with the old database as well

Here is the challenge @lilitkarapetyan :
* from the game last week we have a large body of test data.  It uses the current "model" of having a really large `wargame` document
* the clients and I would like to run more tests against this data, including testing how adjudication should work.  So, I'll be running new versions of `Serge` software against that database file

So, I can think of these options:
1. update their SQLite database to separate out the two documents, then use updated Serge. Aah, there are actually several versions of the wargame document - and I've have to update all of them.
2. keep this PR open until we/they have finished testing the data (about 2 months, I guess). But, no-one likes a PR sitting around.
3. make this PR support their database format, but "retire" this legacy support when they have finished testing. Note: this isn't all that difficult.  The database will _either_ be in the new format or the old format.  So, `newState` will either receive an object with all the data, or it will receive just the `wargame`, plus a separate document of `settings`, which it will put into `newState`.

What do you think @lilitkarapetyan  - do you have a favourite option?

lilitkarapetyan:
> > Hello Ian, no it won't work with the old SQLite database, but can I make it work with the old database as well
> 
> Here is the challenge @lilitkarapetyan :
> 
>     * from the game last week we have a large body of test data.  It uses the current "model" of having a really large `wargame` document
> 
>     * the clients and I would like to run more tests against this data, including testing how adjudication should work.  So, I'll be running new versions of `Serge` software against that database file
> 
> 
> So, I can think of these options:
> 
>     1. update their SQLite database to separate out the two documents, then use updated Serge. Aah, there are actually several versions of the wargame document - and I've have to update all of them.
> 
>     2. keep this PR open until we/they have finished testing the data (about 2 months, I guess). But, no-one likes a PR sitting around.
> 
>     3. make this PR support their database format, but "retire" this legacy support when they have finished testing. Note: this isn't all that difficult.  The database will _either_ be in the new format or the old format.  So, `newState` will either receive an object with all the data, or it will receive just the `wargame`, plus a separate document of `settings`, which it will put into `newState`.
> 
> 
> What do you think @lilitkarapetyan - do you have a favourite option?

Does that mean we are going to complate this PR as originally planned or do we have to wait until they/we finish testing the data ?

lilitkarapetyan:
> > Hello Ian, no it won't work with the old SQLite database, but can I make it work with the old database as well
> 
> Here is the challenge @lilitkarapetyan :
> 
>     * from the game last week we have a large body of test data.  It uses the current "model" of having a really large `wargame` document
> 
>     * the clients and I would like to run more tests against this data, including testing how adjudication should work.  So, I'll be running new versions of `Serge` software against that database file
> 
> 
> So, I can think of these options:
> 
>     1. update their SQLite database to separate out the two documents, then use updated Serge. Aah, there are actually several versions of the wargame document - and I've have to update all of them.
> 
>     2. keep this PR open until we/they have finished testing the data (about 2 months, I guess). But, no-one likes a PR sitting around.
> 
>     3. make this PR support their database format, but "retire" this legacy support when they have finished testing. Note: this isn't all that difficult.  The database will _either_ be in the new format or the old format.  So, `newState` will either receive an object with all the data, or it will receive just the `wargame`, plus a separate document of `settings`, which it will put into `newState`.
> 
> 
> What do you think @lilitkarapetyan - do you have a favourite option?

Does that mean we are going to complete this PR as originally planned or do we have to wait until they/we finish testing the data ?

IanMayo:
Hello @lilitkarapetyan - if we choose options 1 or 3, we implement this PR.  If we choose option 2, we pause this PR.

Do you have a preferred option?

lilitkarapetyan:
I think the second version will be good because we will have more time to study and check the data, this is an important change

lilitkarapetyan:
of the other two options, I would prefer the first

IanMayo:
So, option 2 (`keep this PR open`) is your favourite solution @lilitkarapetyan ?

lilitkarapetyan:
Well,  it's not my favorite, but I think it's the best solution in this case because we're going to make an important change to our project.

lilitkarapetyan:
we can also do with the first option

lilitkarapetyan:
I haven't finished.

lilitkarapetyan:
Hello Ian, can you check this PR ?

lilitkarapetyan:
Hello @IanMayo, can you check this PR ?

IanMayo:
Hello @lilitkarapetyan - These are a set steps to force wargame changes.  They cause new versions of the `Forces` document to be created.  The updates appear to be working properly.

1. run `yarn develop`
2. login as game control: http://localhost:3000/?wargame=wargame-l6nngxlk&access=umpire
3. open `Adjudications` tab
4. click on `Get next`, a new `Interaction` will be created (this is a new document in SQLite, too)
5. make panel wide enough to see form, and scroll down the form to the `Health outcomes`
6. modify the health of `Red:22` to `0`
7. move back up to the `Perception Outcomes`
8. add a new row
9. select Asset `Red:22`, provide `dolphin` as the `Name`
10. scroll down to bottom, click on `Submit Adjudication`
11. switch to `Orders` tab
12. type `Red:22` into search box, `Red:22` should be at the top of the listing
13. it does now who a health of `0`

IanMayo:
@lilitkarapetyan - we weren't clearing the form on cancel.  I think I've fixed it, and tried to explain the new `clearForm` prop here: https://github.com/serge-web/serge/pull/2359/commits/9c9c285b7348accda92a9402aef0f84e5a9c0735

Does it look ok to you?

IanMayo:
Aah, the GameAdmin pages don't work, since they still expect to find a `Forces` object the `wargame` object in `RootState`.  
image


I guess we need to either change RootState (to put in the ForceData[] array, or need to put the `ForceData[]` array into the `wargame` object for `RootState`.  The first one seems like better engineering.

IanMayo:
Closing this PR.  It's no longer relevant after our switch to `core-mapping`

PR #2138, opened 2022-11-19
Add ruler control to map

### LINKED ISSUE #2133: Add ruler control to map
Lots of players this week wanted to measure distances on the map.

We should add a leaflet measure control.  I can't find a react component to do it, but there's a way of doing it (similar to what @philb89 ) has done before here:
https://stackoverflow.com/questions/67746299/react-leaflet-measurement-ruler-plugin

And here is a control that looks easy to use:
https://github.com/gokertanrisever/leaflet-ruler

lebaphi:
Fixes #2133 

https://user-images.githubusercontent.com/26705621/202845077-8f79e4e9-eef9-48fb-95ba-3937f16c2d12.mov


IanMayo:
thanks @lebaphi  - works well.

The "New Orders" popup button appears at the top left (when the current player is a Blue Player, and we're in `Planning Mode`.   So, could this new button appear with the other buttons on the top right-hand side?  Otherwise we'll have to have it at the bottom-left or bottom-right.

packages/components/src/local/organisms/planning-assets/index.tsx
```diff
@@ -71,7 +71,9 @@ export const PlanningAssets: React.FC<PropTypes> = ({
     components={{
       Body: (props): React.ReactElement => {
         if (props.columns.length && onVisibleRowsChange) {
-          onVisibleRowsChange(props.renderData)
```
IanMayo, line 74: hello @lebaphi  - can you teach me what this fix does please?

lebaphi: I think this hook calls immediately when the PlanningChannel is rendering -> reports an error, add setTimeout to make sure the PlanningChannel loads fully, then we call this hook

lebaphi:
> thanks @lebaphi - works well.
> 
> The "New Orders" popup button appears at the top left (when the current player is a Blue Player, and we're in `Planning Mode`. So, could this new button appear with the other buttons on the top right-hand side? Otherwise we'll have to have it at the bottom-left or bottom-right.

sure @IanMayo , let me update

lebaphi:
The ruler now is in the bottom right 😄 

image

IanMayo (approved):

PR #2149, opened 2022-11-24
New_solution_for_persistent_forms

lilitkarapetyan:
## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Thanks Lilit.  Data isn't lost, but we still have the flicker.

Could you investigate an alternate solution please?

I'm thinking of something like this:
* we only re-create the JSON editor if some test is passed, such as a new item is being edited

I'm happy to tidy this PR up, and merge it - but the users I spoke to last week are expecting a solution that doesn't involve flicker.  This is what they are used to: when they are instant-messaging with their family, the instant messenger doesn't flicker when a new message is received.  So, they know it is fixable, and so find the flicker annoying.

packages/components/src/local/molecules/json-editor/index.tsx
```diff
@@ -74,19 +73,25 @@ export const JsonEditor: React.FC<Props> = ({
...
+
   const OnSave = () => {
     saveMessage && saveMessage()
     expiredStorage.removeItem(genLocalStorageId())
+    expiredStorage.removeItem(getEditLocalStorgaeid())
   }
 
   const onPopupCancel = (): void => {
-    expiredStorage.removeItem(genLocalStorageId())
+    // removePlanning
```
IanMayo, line 71: @lilitkarapetyan  - do you remember what the `removePlanning` comment is for?

packages/components/src/local/molecules/json-editor/index.tsx
```diff
@@ -74,19 +73,25 @@ export const JsonEditor: React.FC<Props> = ({
     return memoryName
   }
 
+  const getEditLocalStorgaeid = (): string => `edit-${genLocalStorageId()}`
```
IanMayo: @lilitkarapetyan - we have two methods with similar names (`getEditLocalStorageId` and `genLocalStorageId`).  Can you put in comments to indicate what they are both for, please?

packages/components/src/local/molecules/json-editor/index.tsx
```diff
@@ -230,12 +235,14 @@ export const JsonEditor: React.FC<Props> = ({
   }, [disableArrayToolsWithEditor && disabled])
 
   useLayoutEffect(() => {
```
IanMayo, line 202: @lilitkarapetyan - I tried to understand what was happening in this method before, and couldn't get it.

My favourite outcome would be to simplify the API/props for the `json-editor`.  I feel we've got some props that do the same job as each other.  If we can't do that, I'll need better comments on the props, and in the method please.

packages/components/src/local/molecules/json-editor/index.tsx
```diff
@@ -263,7 +270,10 @@ export const JsonEditor: React.FC<Props> = ({
                 : null
             }
           </>
-          : !disabled ? <Button color='secondary' onClick={() => { setBeingEdited(true) }} icon='edit'>Edit</Button>
```
IanMayo, line 270: Oh dear, `JSON.stringify('edit')` is too complex for me. Won't `JSON.stringify('edit')` just return `"edit"` ?

lilitkarapetyan:
> Thanks Lilit. Data isn't lost, but we still have the flicker.
> 
> Could you investigate an alternate solution please?
> 
> I'm thinking of something like this:
> 
>     * we only re-create the JSON editor if some test is passed, such as a new item is being edited
> 
> 
> I'm happy to tidy this PR up, and merge it - but the users I spoke to last week are expecting a solution that doesn't involve flicker. This is what they are used to: when they are instant-messaging with their family, the instant messenger doesn't flicker when a new message is received. So, they know it is fixable, and so find the flicker annoying.

Hello Ian, I don't think the problem is with jsonEditor because by default when we don't have anything created with jsonEditor it flashes again when we get a new message, I will work on that and try to fix the flashing.

packages/components/src/local/molecules/json-editor/index.tsx
lilitkarapetyan: Should we delete the chached message ?

packages/components/src/local/molecules/json-editor/index.tsx
lilitkarapetyan: Ok I will add a comment in the code.

packages/components/src/local/molecules/json-editor/index.tsx
lilitkarapetyan: I will look into it.

lilitkarapetyan: I don't remember where we use it either, I guess we can remove it from the `jsonEditor`

lilitkarapetyan:
I will try improve it so that when updating a message the scrollbar is in the edit area

packages/components/src/local/organisms/planning-messages-list/index.tsx
```diff
@@ -54,7 +54,7 @@ export const PlanningMessagesList: React.FC<PropTypes> = ({
     }
   }
 
-  useEffect(() => {
+  useLayoutEffect(() => {
```
IanMayo, line 53: Hey @lilitkarapetyan - THIS WORKS PERFECTLY!

Can you explain how it fixes it, please?

lilitkarapetyan:
I'm still working on this PR

IanMayo:
Hello Lilit, thanks for keeping this moving.  But, as I said last week, me preferred outcome is to identify a solution that avoids cache-ing data on table updates.  My preferred outcome is that we use state to prevent JSON-Editor (or its react parents) from being refreshed.

lilitkarapetyan:
> Hello Lilit, thanks for keeping this moving. But, as I said last week, me preferred outcome is to identify a solution that avoids cache-ing data on table updates. My preferred outcome is that we use state to prevent JSON-Editor (or its react parents) from being refreshed.

Hello Ian, okay, I will work on it now

packages/components/src/local/organisms/planning-messages-list/index.tsx
```diff
@@ -4,45 +4,70 @@ import { MessageDetails, MessagePlanning, PerForcePlanningActivitySet, PlannedAc
...
   !7 && console.log('planning selectedOrders: ', selectedOrders, !!setSelectedOrders, messages.length)
 
   const [myMessages, setMyMessages] = useState<MessagePlanning[]>([])
+  const roles: string[] = []
   useEffect(() => {
     const showOrdersForAllRoles = !onlyShowMyOrders
     const myForceMessages = messages.filter((message: MessagePlanning) => isUmpire || message.details.from.forceId === playerForceId)
     const myRoleMessages = myForceMessages.filter((message: MessagePlanning) => showOrdersForAllRoles || message.details.from.roleId === playerRoleId)
-    setMyMessages(myRoleMessages)
+
```
IanMayo: Hey @lilitkarapetyan - is this next block where the new magic happens ?

IanMayo:
Hello @lilitkarapetyan - could you update me on where we are with this issue please?

lilitkarapetyan:
> Hello @lilitkarapetyan - could you update me on where we are with this issue please?

Hello Ian, I'm going to graduate

lilitkarapetyan:
I finished it

IanMayo:
Hello @lilitkarapetyan - I'm keen to merge this (hooray!), but I'm just doing a bit of last-minute testing.  We only use `cacheMessage` in one place, `<MessageCreator>`. This component is used in creating new `Orders` messages in the `Planning` tab.  I've tried changing the value of `cacheMessage` to false in that component - and things still seem to work. 

I've pushed the change. Would you mind double-checking?  

That's because if we don't need to use `cacheMessage={true}` anywhere, we can remove a lot of the `expiredStorage` code.

IanMayo:
Aah, I've found an issue @lilitkarapetyan 

When I ask to filter for "Orders" that belong to me, the list doesn't update, but when I cancel the filter, the table grows by one (as it re-processes the first element).  I don't think it matters if we lose live edits when a filter is applied/removed, since the user is deliberately changing the contents of the page.
https://i.gyazo.com/40eb053789a671fabd3058f4adfdfe67.mp4

lilitkarapetyan:
> Hello @lilitkarapetyan - I'm keen to merge this (hooray!), but I'm just doing a bit of last-minute testing. We only use `cacheMessage` in one place, `<MessageCreator>`. This component is used in creating new `Orders` messages in the `Planning` tab. I've tried changing the value of `cacheMessage` to false in that component - and things still seem to work.
> 
> I've pushed the change. Would you mind double-checking?
> 
> That's because if we don't need to use `cacheMessage={true}` anywhere, we can remove a lot of the `expiredStorage` code.

Hello Ian, I will into it now

lilitkarapetyan:
Hello Ian, I will into it now

IanMayo:
I'm not sure how this problem arose, but we have a layout issue with the `Orders` tab:
image

lilitkarapetyan:
> I'm not sure how this problem arose, but we have a layout issue with the `Orders` tab: image

Ok, I will look into it

IanMayo:
Hello @lilitkarapetyan - I like this piece of work. Thanks for refactoring out those helpers.

`json-editor` has a `cacheMessage` prop.  As far as I can see, that is never true.  If I'm right about that, we never call `expiredStorage.setItem` in `json-editor\index.tsx`.  If that's right, we never need to call `expiredStorage.getItem` or `expiredStorege.removeItem` in `json-editor\index.tsx`.   If that's right, we don't need to refer to `expiredStorage` in `json-editor\index.tsx` at all - and we can remove it.

Yes, there are a few _if I'm right_'s above.  Would you mind checking it please?

lilitkarapetyan:
> 



> Hello @lilitkarapetyan - I like this piece of work. Thanks for refactoring out those helpers.
> 
> `json-editor` has a `cacheMessage` prop. As far as I can see, that is never true. If I'm right about that, we never call `expiredStorage.setItem` in `json-editor\index.tsx`. If that's right, we never need to call `expiredStorage.getItem` or `expiredStorege.removeItem` in `json-editor\index.tsx`. If that's right, we don't need to refer to `expiredStorage` in `json-editor\index.tsx` at all - and we can remove it.
> 
> Yes, there are a few _if I'm right_'s above. Would you mind checking it please?
Hello Ian , Thanks I will check it and remove `expiredStorage` from `JsonEditor`

IanMayo:
> Thanks I will check it and remove expiredStorage from JsonEditor

The `expiredStorage` fix was a great workaround, that made a monster problem go away.  So, it feels _scary_ to delete it.  But, if we really can remove it, then great :-D

IanMayo:
Note: this bug should be included in this PR please, @lilitkarapetyan : https://github.com/serge-web/serge/issues/2164

IanMayo (approved):
tested and working

PR #2163, opened 2022-12-14
2157: Duplicate row feature in JSON-Editor

### LINKED ISSUE #2157: Duplicate row feature in JSON-Editor
For some type of "Own Asset" forms, we wish to be able to "duplicate" a row.

This is done with `enable_array_copy" in the JSON-Editor.

Test this, and verify we can see the button.  This is the only time when we want `copy` - so I guess a flag has to _somehow_  come from the template.

Hmm, in one editor (template) we want this `copy` button for some arrays, but not for others.  So, _maybe_ it's a case of a having a flag to enable it at the template level, then hiding the button for all arrays that don't have a particular flag.

lebaphi:
Fixes #2157


IanMayo:
Just this one `codacy` error, please @lebaphi 
image

packages/themes/_btns.scss
```diff
@@ -272,3 +272,9 @@ button {
 .btn-hide {
   display: none !important;
 }
+
```
IanMayo, line 275: @lebaphi  - could you explain what problem this next block overcomes, please?

lebaphi: Hi @IanMayo, I just adjusted the button's color when it's disabled for ease of viewing, without this block, this button looks like:

image

IanMayo: @lebaphi - is there a reason why we don't make it hidden when disabled?
image

lebaphi: ah, i think disable should just update the state of the button instead of hiding it, it makes more sense. But if you want to hide this button, it's ok.

IanMayo: Thanks for trying to "do the right thing" @lebaphi  .   In this instance, the `Delete` button isn't shown when the form is disabled, so I think it's logical to also not show the `Copy` button:
screencast

So, ok to remove?

lebaphi: > So, ok to remove?

sure @IanMayo

PR #2173, opened 2022-12-20
2172 increase volume of assets

### LINKED ISSUE #2172: Increase volume of mock assets
There are some performance issues with large numbers of assets (it can take a second or so to make them all appear when `Select all` is selected).

Hmm, part of this issue _could_ be that we're re-downloading all of the SVG, rather than cache-ing them.

Produce a higher volume of blue and red assets, so we can investigate performance.

Try 2500 and 3000 assets

IanMayo:
Fixes #2172


IanMayo:
Re-do, based upon `develop`

PR #2176, opened 2022-12-21
Fix issue when integrate with json-editor v2.8.0

### LINKED ISSUE #2171: Bug in `develop` - JSONEditor not found
Aaah, err, I think I've introduced a bug into develop, while integrating the new location-editor code.

image

Steps to reproduce:
- run `develop` branch
- log in as `MTG-1`

It looks like we have this new method of referring to JSONEditor, but JSONEditor is undefined in the above issue:
image

Note: as I look at the codebase, there are quite a few instances of `JSONEditor.defaults`, rather than `JSONEditor.JSONEditor.defaults`.  Could we investigate which is the correct pattern to use?

lebaphi:
Fixes #2171 

Hi @IanMayo, I just made a temporary fix to get it working, to support custom editor like location I need to spend more time to research. The class `AbstractEditor.extend` no longer exists.

### DIFF
--- a/packages/client/src/Components/jsonSchemaEditor/index.js
+++ b/packages/client/src/Components/jsonSchemaEditor/index.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react'
 import jsonMetaSchema from './data.json'
-import JSONEditor from '@json-editor/json-editor'
+import { JSONEditor } from '@json-editor/json-editor'
 import Editor from './jsonEditor'
 import Preview from './editorPreview'
 import '@fortawesome/fontawesome-free/css/all.css'
@@ -11,8 +11,8 @@ class SchemaEditor extends Component {
   constructor (props, content) {
     super(props, content)
 
-    JSONEditor.JSONEditor.defaults.options.iconlib = 'fontawesome5'
-    JSONEditor.JSONEditor.defaults.options.theme = 'bootstrap4'
+    JSONEditor.defaults.options.iconlib = 'fontawesome5'
+    JSONEditor.defaults.options.theme = 'bootstrap4'
     this.updateMetaSchema = this.updateMetaSchema.bind(this)
     this.updatePreviewSchema = this.updatePreviewSchema.bind(this)
     this.updateOptions = this.updateOptions.bind(this)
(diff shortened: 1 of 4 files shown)


IanMayo:
> The class AbstractEditor.extend no longer exists.

Oh dear @lebaphi - that sounds scary.

lebaphi:
> > The class AbstractEditor.extend no longer exists.
> 
> 
> 
> Oh dear @lebaphi - that sounds scary.

no worries @IanMayo, I think there is an alternative solution for it

IanMayo:
Thanks for this @philb89 - is there a particular StoryBook I can use to view the new editor, is it via `yarn develop`?

lebaphi:
> Thanks for this @philb89 - is there a particular StoryBook I can use to view the new editor, is it via `yarn develop`?

Hi @IanMayo , currently not, I think we should remove this block first
image

Then we can create a new storybook for the location editor 🤔

IanMayo:
> currently not, I think we should remove this block first

Ok, but I _think_ we still need some processing like that.  We need to store the current GeoJSON data in the form (somewhere - maybe in the `location-editor`, or `DOM`), and we need to re-format the GeoJSON in human-readable form. 

We have these requirements. When the form contains a `location-editor`:
1. store the `PlannedActivityGeometry[]` JS array 
2. call a `render` function to convert the array to human readable `<ul>` list, display this as read-only in the `div`
3. provide an `Edit` button, when the form is editable
4. when the `Edit` button is clicked, provide the raw `PlannedActivityGeometry[]` array to the callback
5. when `Save` is clicked during location edit, store the new `PlannedActivityGeometry[]`, and re-do the `Render` function
6. when `Save` is clicked for the JSON-Editor, push the `current` value of of `PlannedActivityGeometry` back into the location field of the results object.

IanMayo:
Note: this isn't relevant for this immediate issue, but note that in the future, we may have a table of `location` objects to edit.  So, we'll get this sorted first, and in the future come up with a way of the form containing multiple rows that contain a location-editor.  The drop-down in the below image will be replaced with a new `location-editor` with an `edit` button.

image

PR #2184, opened 2022-12-25
2180: Use icons for copy/delete/add in JSONEditor

### LINKED ISSUE #2180: Use icons for copy/delete/add in JSONEditor
For some of our forms, there is a fight for screen space.

Using the JSON editor button icons instead of labels could save some space.

Could you try to switch to button icons please @philb89 ?

lebaphi:
Fixes #2180


IanMayo:
Hello @philb89 - SERGE will be running without an Internet connection, so we can't download the material icons at run-time. We have to include them in the build.  I'm pretty sure we do have at least one set of icons included in the build already (Font-awesome, I think), so _maybe_ you can direct JSON-Editor to use them?

lebaphi:
Yes @IanMayo, I think I'm using the icon from `iconlib: 'fontawesome5',` we don't download the material icons at run-time
image

IanMayo:
> Yes @IanMayo, I think I'm using the icon from `iconlib: 'fontawesome5',` we don't download the material icons at run-time image

My mistake. I re-did the build, and I saw it was using local icons :-)

PR #2192, opened 2022-12-31
More fixes for LocationEditor 

### LINKED ISSUE #2183: More fixes for LocationEditor
Some steps to complete the location editor are recorded [here](https://twitter.com/RishiSunak/status/1605964002143522816)

Copied below:

Ok, but I _think_ we still need some processing like that.  We need to store the current GeoJSON data in the form (somewhere - maybe in the `location-editor`, or `DOM`), and we need to re-format the GeoJSON in human-readable form. 

We have these requirements. When the form contains a `location-editor`:
1. store the `PlannedActivityGeometry[]` JS array 
2. call a `render` function to convert the array to human readable `<ul>` list, display this as read-only in the `div`
3. provide an `Edit` button, when the form is editable
4. when the `Edit` button is clicked, provide the raw `PlannedActivityGeometry[]` array to the callback
5. when `Save` is clicked during location edit, store the new `PlannedActivityGeometry[]`, and re-do the `Render` function
6. when `Save` is clicked for the JSON-Editor, push the `current` value of of `PlannedActivityGeometry` back into the location field of the results object.

lebaphi:
Fixes #2183


packages/components/src/local/organisms/planning-channel/index.tsx
```diff
@@ -531,17 +531,21 @@ export const PlanningChannel: React.FC<PropTypes> = ({
...
-        coords.forEach((val: [number, number]) => {
-          const pos = L.latLng(val[0], val[1])
-          localBounds = localBounds === undefined ? L.latLngBounds(pos, pos) : localBounds.extend(pos)
-        })
-      } else if (geom.coordinate) {
-        const val = geom.coordinate as [number, number]
-        const pos = L.latLng(val[0], val[1])
-        localBounds = localBounds === undefined ? L.latLngBounds(pos, pos) : localBounds.extend(pos)
-      }
+      // currently I saw the geom.coordinates has the same format with geom.coordinate, so I guess there is something wrong with the data
```
IanMayo: note: I _think_ we receive `coordinate` if there is only one item, e.g. if it's a point rather than a line/polygon

packages/components/.eslintrc.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     'react'],
   "ignorePatterns": ["*.scss.d.ts", '*.snap'],
   'rules': {
+    "space-before-function-paren": [2, "never"],
```
IanMayo: @lebaphi - is the above change going to cause lots of linter updates in other files?

lebaphi: Don't worry @IanMayo, it just disables this rule, doesn't reformat other files

packages/components/src/local/organisms/planning-channel/index.tsx
lebaphi: hmm, can you help to re-check it 🤔 , I think the `geom.coordinates` handles data for all types: `Array of [number, number]` and `[number, number]`

image

packages/components/.eslintrc.js
lebaphi: I cleaned up some code and removed it. So just ignore it.

IanMayo:
hello @philb89 - I'm afraid the "save location edits" is broken.  When I edit "Blue-3 EW Attack", it's the formatted HTML that gets stored in the saved message, not the GeoJSON.  

**Update:** No, that's wrong.  Saving an edit works, it's cancelling a location edit that fails.  If I edit EW-Attack, then cancel the location-edit, but still save the updated message, then the plain text gets stored. This is because there is nothing in `pendingLocationData`.  So, the handler for `cancel edit location` should push the original GeoJSON to the save handler.  Experimenting with this.

Note: The existing solution relied on the `pendingLocationData` state object.    In `PlanningMessagesList`, we used to intercept the `saveMessage`, and replace the location (pretty HTML) data with the stored `pendingLocation` data. Somehow we have lost the code that updates `pendingLocationData`.

But, I do think we have  an opportunity for a more robust solution now.  It would work like this:
- location editor component has hidden text component `hiddenStore`.  On load (setValue(), the GeoJSON from `location` gets put into there, before the text area gets written with the human-readable text.  When `Edit` is clicked for the location component, a callback is provided that, when called, updates the contents of `hiddenStore` with the new GeoJSON, and updates the human-readable text.  When `getValue()` is called for location-editor, it returns the GeoJSON from hiddenStore, not the human-readable text.

This avoids the need for `pendingLocationData` in the parent classes.

IanMayo (approved):
Working solution.

Improved solution proposed here: #2195

lebaphi:
> hello @philb89 - I'm afraid the "save location edits" is broken. When I edit "Blue-3 EW Attack", it's the formatted HTML that gets stored in the saved message, not the GeoJSON.
> 
> **Update:** No, that's wrong. Saving an edit works, it's cancelling a location edit that fails. If I edit EW-Attack, then cancel the location-edit, but still save the updated message, then the plain text gets stored. This is because there is nothing in `pendingLocationData`. So, the handler for `cancel edit location` should push the original GeoJSON to the save handler. Experimenting with this.
> 
> Note: The existing solution relied on the `pendingLocationData` state object. In `PlanningMessagesList`, we used to intercept the `saveMessage`, and replace the location (pretty HTML) data with the stored `pendingLocation` data. Somehow we have lost the code that updates `pendingLocationData`.
> 
> But, I do think we have an opportunity for a more robust solution now. It would work like this:
> 
> * location editor component has hidden text component `hiddenStore`.  On load (setValue(), the GeoJSON from `location` gets put into there, before the text area gets written with the human-readable text.  When `Edit` is clicked for the location component, a callback is provided that, when called, updates the contents of `hiddenStore` with the new GeoJSON, and updates the human-readable text.  When `getValue()` is called for location-editor, it returns the GeoJSON from hiddenStore, not the human-readable text.
> 
> This avoids the need for `pendingLocationData` in the parent classes.

Hi @IanMayo, I wonder if we need `hiddenStore` text element, as you can see in `location-editor.ts` we have `this.locations` variable to store GeoJSON. So when Load(setValue()), I also set GeoJSON from `location` to `this.locations` on line 78, we also work to render GeoJSON in human readable text from code line 81 -> 99, we also return `this.locations` value in getValue() function.

IanMayo:
Yes, @lebaphi - you are right.  We are using `this.locations` instead of `this.hiddenStore`

So, if you are suggesting we don't need `hiddenStore`, then I agree :-D

PR #2226, opened 2023-01-08
Order Editing - don't let features get deleted

### LINKED ISSUE #2222: Order Editing - don't let features get deleted
In the `OrderEditing` module, players are able to delete nodes in lines/areas by right-clicking on them.

But, if we have less than two points in a line, the line disappears, and can't be redrawn. 

The `preventMarkerRemoval` flag would prevent the above, but we would need to dynamically updated it, according to if a line has two elements or not.

Steps to reproduce:
- yarn develop
- log in as MTG1
- go to Orders tab
- find `ASW Barrier`
- click to Edit document
- click to Edit locations, map will zoom in on locations
- right click on points on line to delete

We don't want user to be able to completely remove the line.

https://user-images.githubusercontent.com/1108513/211052735-517ab7c7-ddc0-4308-aeb7-d404c6d49922.mp4

lebaphi:
Fixes #2222

### DIFF
--- a/packages/components/src/local/organisms/planning-channel/helpers/OrderEditing.tsx
+++ b/packages/components/src/local/organisms/planning-channel/helpers/OrderEditing.tsx
@@ -112,7 +112,10 @@ export const OrderEditing: React.FC<OrderEditingProps> = ({ saved, activityBeing
 
       layerToEdit.on('pm:edit', e => {
         const pointsOnLine = get(e, 'layer._rings[0]') as Point[]
-        const preventRemove = !!(e.shape.toLowerCase() === 'line' && pointsOnLine.length <= 2)
+        const preventRemove = !!(
+          (e.shape.toLowerCase() === 'line' && pointsOnLine.length <= 2) ||
+          (e.shape.toLowerCase() === 'polygon' && pointsOnLine.length <= 3)
+        )
         const options = get(e, 'layer.pm.options')
         const newOptions = {
           ...options,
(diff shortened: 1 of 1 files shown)


IanMayo:
Hmm, I think I can see how this would work. But, I think it needs to be updated on modified.  I could start out with 4 points, and this would allow node removal. But, I could then delete 3 of the points.

I _Think_ we need to track changes - and control the property according to the number of points on the line.

IanMayo:
Hi, I just tested this in the deployment, and it's not a valid solution:
- log in as MTG1
- switch to Orders tab
- filter to "My orders"
- open "ASW Barrier"
- click "Edit"
- go down to "Location", click on "Edit"
- start to delete points from a route line (via right-click)
- it is possible to get down to two points, and then delete one of them. which makes the line disappear

I _think_ we have to implement this function, and return false if it's a line and there are two points, or if it's a polygon and there are 3 points (I forgot about the polygon instance):
image

packages/components/src/local/organisms/planning-channel/helpers/OrderEditing.tsx
```diff
@@ -112,7 +112,10 @@ export const OrderEditing: React.FC<OrderEditingProps> = ({ saved, activityBeing
 
       layerToEdit.on('pm:edit', e => {
         const pointsOnLine = get(e, 'layer._rings[0]') as Point[]
-        const preventRemove = !!(e.shape.toLowerCase() === 'line' && pointsOnLine.length <= 2)
+        const preventRemove = !!(
```
IanMayo, line 115: Hopefully I am wrong, but I _don't think_ this will fix it.  I _think_ this logic is only called when editing starts.  In the above reproducible steps, there are enough points at the start of editing, it's during editing that problem occurs.

IanMayo (approved):
Working perfectly

Issue #2227, opened 2023-01-09
Generate STRIKE interaction


IanMayo:
We've identified the need for `short-circuit` interactions.

These aren't subject to the normal interaction generation, but are when an activity needs it's own adjudication - where it has a game outcome that is independent of other activities (ISTAR, Strike, EW, Movement, others).

Provide a first sample of this - via a STRIKE interaction.

IanMayo:
Done.

PR #2229, opened 2023-01-10
2227 strike interaction

IanMayo:
Supports #2227


IanMayo:
Part-way implementation.

Merging, so other work (#2233) can develop based upon this.

PR #2248, opened 2023-01-14
2236: Dynamically regenerate sub-type based upon filtered type data

### LINKED ISSUE #2236: Dynamically regenerate `sub-type` based upon filtered `type` data
In the `Assets` tab, we can now filter by `Platform Type` and `Sub-Type`.  Well, it's in #2235 

Both a currently drop-downs containing all of the values found in the data.  But, the sub-types column contains all values - even those which are no longer relevant.

We may/may-not wish to this for other columns in the future, but I'd like `sub-type` drop-down to only contain the relevant.

This could either be through:
1. When we load the data, store sub-types with the platform-type they're found under.  Then once we've filtered on platform-type, we can get the selected platform-type values, and filter sub-type to only show those
2. When a filter has been applied, run through the visible rows and collate a list of unique sub-types

Ok, I think the answer here is that the filters should be created from the current set of rows, not from the parent dataset (of assets).  So, in PlanningAssets we already have the callback for `onVisibleRowsChange`. I think this needs to trigger the update of the filter lists.

lebaphi:
Fixes #2236


IanMayo (approved):
Nice solution, thanks @lebaphi

PR #2254, opened 2023-01-14
2253 persistent tables

### LINKED ISSUE #2253: Material table - filters reset on change tab
Our material tables aren't persistent.  They are only "alive" when their tab is alive.

This means that when we change to another tab, and then return, any filters have been lost.

I _think_ we should get around this by storing the state of the filters for each table in the parent `SupportPanel` class (unless there is a better solution).

So, for example, `planning-assets` will have props for `filters` and `setFilters()`.   When it loads, if `filters` is empty, it will generate its columns object, and store either the columns or the currently applied filters in the parent via `setFilters`.  When the player changes tab, and comes back again, it will receive the stored value of `fliters` - and the original set of filters will be restored.

IanMayo:
Fixes #2253


IanMayo:
Close. Duplicate of #2256

PR #2256, opened 2023-01-15
2253: Material table - filters reset on change tab

### LINKED ISSUE #2253: Material table - filters reset on change tab

lebaphi:
Fixes #2253


IanMayo:
Oops @lebaphi  - we've both worked on this issue.  I have this PR for it:
https://github.com/serge-web/serge/pull/2254

I _think_ we've both taken the same strategy. Would you mind checking mine, please?

lebaphi:
> Oops @lebaphi - we've both worked on this issue. I have this PR for it: #2254
> 
> I _think_ we've both taken the same strategy. Would you mind checking mine, please?

Hi @IanMayo , yes we've both taken the same strategy 🤣

lebaphi:
But I have updated the fix for the subtypes filter on `custom-filter-row.tsx`

IanMayo:
> Hi @IanMayo , yes we've both taken the same strategy 🤣

Yours is neater.  I'll pull over a couple of my extra changes and merge it.

IanMayo:
> if (asset.attributes && ((filter.length && filter.includes(asset.platformTypeId)) || !filter.length)) {

Could you explain how the above line works please?

lebaphi:
> > if (asset.attributes && ((filter.length && filter.includes(asset.platformTypeId)) || !filter.length)) {
> 
> Could you explain how the above line works please?

hi @IanMayo, that means we build a subTypes list base on corresponding selected platformType or build all subTypes if no platformType is selected

packages/components/src/local/organisms/planning-assets/helpers/custom-filter-row.tsx
```diff
@@ -24,7 +24,7 @@ const CustomFilterRow: React.FC<CustomFilterRowProps> = (props): React.ReactElem
         return
       }
       force.assets.forEach((asset: Asset) => {
-        if (asset.attributes && ((filter.length && filter.includes(asset.platformTypeId)) || !filter.length)) {
+        if (asset.attributes && (!filter.length || filter.includes(asset.platformTypeId))) {
```
IanMayo, line 27: you ok with this refactoring @lebaphi ?

IanMayo (approved):

PR #2282, opened 2023-01-19
Avoid get content from internet

### LINKED ISSUE #2266: Avoid getting content from Internet
For the wargame, the PCs won't be an Internet connection.

So, we need to make sure we package all content.  It looks like we're pulling 3 sets of content from the Internet:
image

(network tab when running `Storybook`)

lebaphi:
Fixes #2266


IanMayo (approved):
Great - from the Network tab in Dev Tools, it now looks like no content is coming from the Internet.

PR #2285, opened 2023-01-21
1977: Adopt alternate symbol generator

### LINKED ISSUE #1977: Adopt alternate symbol generator
An open-source project exists that renders NATO mapping symbols as SVG: https://github.com/spatialillusions/milsymbol

This could greatly open up the range of asset/entity symbols available to Serge game designers.

But, it's a broad change. We should only do it outside a game work-up period.

@lebaphi - I'd like us to do a trial of this alternate component, please.  There's lots of logic in how we create each icon type, and add the `adornments`, but I can provide that input once we've established if it is workable or not.

I _guess_ one way of doing this is to modify `AssetIcon` so that for some icon type, we provide a MilSymbol instead of the  existing `imageSrc`

lebaphi:
Fixes #1977 

Hi @IanMayo, I introduce new component `SymbolAssetIcon` which uses new library `milsymbol`, usage in `collate-assets.tsx`, line 154, this is just basic usage.

Screen Shot 2023-01-21 at 9 35 52 AM


IanMayo:
Thanks @lebaphi - it looks suprisingly simple.

Could you prove we can use these icons the map too, please?  So, when you select the yellow asset in the table,it will pan to it on the map, and we should see the yellow icon there too, p!ease.

IanMayo:
Great, it works.  I _guess_ the change we need is in `<AssetIcon>`, since those icons are used in both forms and on the map.

The current `just change one asset` strategy is fine :-)

I guess other steps are that I need to introduce a new property for the `PlatformTypeData` class - for the `milSymbol` to use for that type of platform.

Note: I'm just pushing a commit that starts to use `sidc` codes stored in the `platform-type` data.  I'm also looking on UpWork for an `expert` who can help us generate the SIDC codes.

lebaphi:
> I guess other steps are that I need to introduce a new property for the PlatformTypeData class - for the milSymbol to use for that type of platform.

Yes @IanMayo 

image

It seems that the new component `SymbolAssetIcon` does not work as the AssetIcon does, but it works when I call directly from the img tag. I'm still investigating.

IanMayo:
Aah, @lebaphi -  I've started to look for someone on UpWork who understands the SIDC codes. But, instead, is it possible to create the icons using plain-English words for the force color, icon type, health, etc?

(I think I saw an example like that)

That would avoid the need to understand SIDC.

lebaphi:
> Aah, @lebaphi - I've started to look for someone on UpWork who understands the SIDC codes. But, instead, is it possible to create the icons using plain-English words for the force color, icon type, health, etc?
> 
> (I think I saw an example like that)
> 
> That would avoid the need to understand SIDC.

Hi @IanMayo , does it help ? https://spatialillusions.com/milsymbol/docs/milsymbol-2525c.html#heading-11

IanMayo:
> Hi @IanMayo , does it help ? https://spatialillusions.com/milsymbol/docs/milsymbol-2525c.html#heading-11

Wow, that's a monster document!

Yes, that's the kind of thing I/we need.

I can't see how we create a icon using "HAWK" - it seems we need that full SIDC.

Aah, but we may need to modify the call to provide force - I think that may be affiliation.  Could you try experimenting with providing an SIDC, but then providing a modifier option to color the force?  Blue is `friendly`, Red is `hostile`, and Green is `neutral`.  

It would also be useful to learn how to use the `Condition` [options](https://spatialillusions.com/milsymbol/docs/milsymbol-2525c.html#heading-3) - so we can use the lib to indicate dead/damaged, etc.

lebaphi:
> > Hi @IanMayo , does it help ? https://spatialillusions.com/milsymbol/docs/milsymbol-2525c.html#heading-11
> 
> Wow, that's a monster document!
> 
> Yes, that's the kind of thing I/we need.
> 
> I can't see how we create a icon using "HAWK" - it seems we need that full SIDC.
> 
> Aah, but we may need to modify the call to provide force - I think that may be affiliation. Could you try experimenting with providing an SIDC, but then providing a modifier option to color the force? Blue is `friendly`, Red is `hostile`, and Green is `neutral`.
> 
> It would also be useful to learn how to use the `Condition` [options](https://spatialillusions.com/milsymbol/docs/milsymbol-2525c.html#heading-3) - so we can use the lib to indicate dead/damaged, etc.

okay, let me check 👌

IanMayo:
Aah, I think I can work out the codes from looking at the sources here:
https://spatialillusions.com/milsymbol/docs/milsymbol-APP6b.html

The second letter is the affiliation (force).  So, 'F' for Friend, 'H' for Hostile, 'N' for Neutral.

Aah, the fourth letter is the condition:  `C` for ok, `D` for damaged, `X` for destroyed.

Progress!!

IanMayo:
Hold on, let me simplify the task :-D

I'd like you to concentrate on finding out how we display the icon on the Leaflet map, please.  This seems to be a demo of how to do it in Leafet:
https://www.spatialillusions.com/milsymbol/example.html#leaflet-divicons

_Hopefully_ we'll only need the milSymbol code in one place.

Once we've got that done, I'll concentrate on all that SIDC stuff.

IanMayo:
Hey, looking good @philb89 

Now that we have proven it is possible,  could you have a look and a think about if we can simplify/refactor the implementation

Also, since it's essentially just plain text, is it an option to create a dictionary (cache) of symbols?  I guess the index would be `sidc` + `stringify(options)`. That would give a performance increase with lots of icons.

lebaphi:
Hi @IanMayo, I think we are caching the icon by using useMemo at this ine 🤔 

image

IanMayo:
> Hi @IanMayo, I think we are caching the icon by using useMemo at this line 🤔

Yes, that cache's individual symbols - to speed up rendering. But, we still create lots and lots of symbols.

A cache would work like this

Currently:
- we need a symbol for [sidc, options]
- we create a symbol for [sidc, options]

Proposed:
- we need a symbol for [sidc, options]
- we ask cache for symbol for [sidc, options]
- cache looks checks in dictionary for entry with key [sidc, options]
    - if found
        - return it
    - else 
        - create symbol for [sidc, options] 
        - store it with key [sidc, options]
        - return it

There may only be, say 40 different symbols in the game (changing for sidc and color) - even though there could be several thousand assets in the game.

Note: we would probably still use the `useMemo` you provided above, to help render speed.

IanMayo:
I may be mistaken @lebaphi  - but it looks like we create the cache in two places: support-panel and planning-channel.   I _guess_ we should only create it once.
Oops, `new LRU` appears in three places.

lebaphi:
> I may be mistaken @lebaphi - but it looks like we create the cache in two places: support-panel and planning-channel. I _guess_ we should only create it once. Oops, `new LRU` appears in three places.

Hi @IanMayo, I think there is a place :D I tested and the cache is only initialized once

IanMayo:
Thanks for this @philb89 

I'm just having a play.  We have these options:
image

If `sizeCalculation` always return `1`, then the `maxEntrySize` value is ignored.

Also:
image

How about we cache the results of `icon.toDataURL()` instead of the symbol?  I've tried doing `stringify(xx).length` on `icon` and on `icon.toDataURL()`.  The size of the icon is 3-4k, and the size of the dataURL is 0.8-2k.  So, we'd use less memory, and reduce the browser processing requirement (since we only to `.toDataURL()` once per symbol.

lebaphi:
> Thanks for this @philb89
> 
> I'm just having a play. We have these options: image
> 
> If `sizeCalculation` always return `1`, then the `maxEntrySize` value is ignored.
> 
> Also: image
> 
> How about we cache the results of `icon.toDataURL()` instead of the symbol? I've tried doing `stringify(xx).length` on `icon` and on `icon.toDataURL()`. The size of the icon is 3-4k, and the size of the dataURL is 0.8-2k. So, we'd use less memory, and reduce the browser processing requirement (since we only to `.toDataURL()` once per symbol.

Hi @IanMayo ,yes we can store `icon.toDataURL()` instead of symbol, it will reduce memory size. I just thought caching the symbol would give us more control over the icon.

IanMayo:
Hello @philb89 - if you switch to the cacheing the DATA URL, we can merge this tomorrow, and I'll try to come up with SCIDs for all of our platform types.

(Sorry guys, I tagged Lilit, I should have tagged Phi. Now fixed)

IanMayo (approved):

PR #2293, opened 2023-01-25
Server API call

lilitkarapetyan:
## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Thanks @lilitkarapetyan ,
a couple of extra bits, please:
- include the turn number in each row
- make it return a unique list of entries (de-dupe)
- provide a client side API call to get this data (wargame API?)
- Call the client side API from Adjudicate Messages List component on gameDate state change,  and output the results to console

lilitkarapetyan:
> 

Okay Ian, I will work on it now

packages/server/providers/pouchdb.js
```diff
@@ -181,6 +181,36 @@ const pouchDb = (app, io, pouchOptions) => {
...
+  app.get('/:wargame/turns', (req, res) => {
+    const databaseName = checkSqliteExists(req.params.wargame)
+
+    if (!databaseName) {
+      res.status(404).send({ msg: 'Wrong Wargame Name', data: null })
+    }
+
+    const db = new PouchDB(databaseName, pouchOptions)
+
+    db.find({
```
IanMayo, line 193: @lilitkarapetyan - can you remind me why we do `not-equal` for `wargameSettings`?  At the start of the game, this is a valid description of game time and game turn length.

packages/client/src/api/db/index.tsx
```diff
@@ -137,6 +136,18 @@ export class DbProvider implements DbProviderInterface {
     })
   }
 
+  getTurnPeriods (): Promise<TurnPeriod> {
```
IanMayo: @lilitkarapetyan -  I think this should return an array of TurnPeriod objects, please.

packages/server/providers/pouchdb.js
```diff
@@ -181,6 +181,36 @@ const pouchDb = (app, io, pouchOptions) => {
...
+      selector: {
+        messageType: INFO_MESSAGE,
+        _id: { $ne: wargameSettings }
+      },
+      fields: ['data', '_id', 'gameTurn']
+    }).then((result) => {
```
IanMayo, line 199: can we de-dupe the array before returning it, please @lilitkarapetyan ?

packages/client/src/ActionsAndReducers/playerUi/playerUi_ActionCreators.ts
```diff
@@ -147,6 +147,10 @@ export const failedLoginFeedbackMessage = (dbName: string, password: string, tur
   }
 }
```
IanMayo, line 154: Note: @lilitkarapetyan - I think this should be returning an array of TurnPeriods

Oh, and hopefully the actual `TurnPeriod[]` data, not `Promise<TurnPeriod[]>`

Oh, and a better method name would be `getTurnPeriods`

packages/components/src/local/organisms/support-panel/types/props.d.ts
```diff
@@ -35,6 +35,7 @@ export default interface PropTypes {
   saveCachedNewMessageValue?: (editMessage: any, messageOption: string) => void
   getCachedNewMessagevalue?: (value: string) => string
   clearCachedNewMessage?: (data: string[]) => void
+  onTurnPeriods?: (gameDate: string, currentWargame: string) => void
```
IanMayo: @lilitkarapetyan - how about we intercept the call up at `factory.tsx` and inject the wargame at that point?

Then we don't need to pass the `gameDate` or `currentWargame` around.

Aah, I think the method should be called `getTurnPeriods`, rather than `onTurnPeriods`.

packages/components/src/local/organisms/adjudication-messages-list/types/props.d.ts
```diff
@@ -90,6 +96,11 @@ export default interface PropTypes extends Omit<ForcesInChannelProps, 'icons' |
   /** descriptions of platform types (used for perception drop-down) */
   platformTypes: PlatformTypeData[]
 
+  onTurnPeriods?: (gameDate: string, currentWargame: string) => void
+  /**
+   * get the turnaround periods data
+  */
+
```
IanMayo: the above comment should go _before_ `onTurnPeriods`, please.

packages/components/src/local/organisms/support-panel/index.tsx
```diff
@@ -469,6 +470,8 @@ export const SupportPanel: React.FC<PropTypes> = ({
                   onMarkAllAsRead={onReadAll}
                   mapPostBack={mapPostBack}
                   channel={channel}
```
IanMayo, line 489: as specified before, if we intercept `onTurnPeriods` in `factory.tsx` we can insert the wargame name at that point.

packages/client/src/api/wargames_api/index.ts
```diff
@@ -890,3 +890,14 @@ export const duplicateAnnotation = (dbName: string, currentAnnation: IconOption)
     return updateWargame({ ...res, data: updatedData }, dbName)
   })
 }
+
```
IanMayo, line 892: this method doesn't need the `getData` parameter.  It just needs the wargame name.

Plus, should it return the actual `TurnPeriod[]` instead of `Promise<TurnPeriod[]>

packages/server/providers/couchdb.js
```diff
@@ -236,6 +236,36 @@ const couchDb = (app, io, pouchOptions) => {
...
 
+  app.get('/:wargame/turns', (req, res) => {
+    const databaseName = checkSqliteExists(req.params.wargame)
+
+    if (!databaseName) {
+      res.status(404).send({ msg: 'Wrong Wargame Name', data: null })
+    }
+
+    const db = new CouchDB(couchDbURL(databaseName))
+
```
IanMayo, line 247: We also need to de-dupe the TurnPeriod array in this implementation, please

packages/custom-types/turn-period.d.ts
```diff
@@ -0,0 +1,8 @@
+import { GameTurnLength } from "./turn-length";
+
+export interface TurnPeriod {
+  gameTurnTime: GameTurnLength,
+  gameDate: string,
+  gameTurn: number,
+  _id?: string
+}
```
IanMayo: I _don't_think_ we need an `_id` in the above type.  It's not coming directly from a PouchDb document, so I don't think we have an _id to use.

packages/server/providers/pouchdb.js
lilitkarapetyan: Hello Ian, I will add it.

IanMayo:
Ok @lilitkarapetyan - I just logged in as `game designer` and stepped forward a few turns. This is what I received:
```
{
  "msg": "ok",
  "data": [
    {
      "gameDate": "2022-05-01T00:00:00+00:00",
      "gameTurnTime": {
        "millis": 259200000,
        "unit": "millis"
      },
      "gameTurn": 1,
      "_id": "2022-09-25T19:29:28.917Z"
    },
    {
      "gameDate": "2022-05-04T01:00",
      "gameTurnTime": {
        "millis": 259200000,
        "unit": "millis"
      },
      "gameTurn": 2,
      "_id": "2023-01-27T16:52:38.201Z"
    },
    {
      "gameDate": "2022-05-04T01:00",
      "gameTurnTime": {
        "millis": 259200000,
        "unit": "millis"
      },
      "gameTurn": 2,
      "_id": "2023-01-27T16:52:44.042Z"
    }
  ]
}
```

I see we get two message for each turn (one for when we switch to adjudication, the other when we switch to planning).  I think we should remove the `_id` field, and de-dupe them. That will give us the information we require.

lilitkarapetyan:
Hello Ian, I'm still making cgenges.

IanMayo:
Thanks for the update Lilit - and for passing the data objects to the right place.

But, in the `TurnPeriod` object, I don't think we need `phase` or `_id`.   Plus, once these fields are removed, we need to remove duplicates from the dataset.  In a _real_ wargame, as the umpires are adjudicating the turns, there will be lots of updates to the wargame document.  These will clutter up the list with unnecessary duplicates.

IanMayo:
Thanks @lilitkarapetyan  - we now just need the de-duplicate.

Now that we have removed `phase` and `_id`, can you see how the same `turn-2` message in two times?

image

lilitkarapetyan:
> 

Yes

IanMayo (approved):

Issue #2294, opened 2023-01-25
As a data-scientist I wish to produce reports based upon SERGE data


IanMayo:
This task is to develop 3 new export buttons in Serge.  

**Change of plan**:  actually, it will be easier to export the data from within the client side of the game, since the data is all already present there.  So, if the current player is from an `umpire` force (`force.umpire === true`), we add an `export` button (`faFileExport`) to the existing buttons in the turn overview panel:

![image](https://user-images.githubusercontent.com/1108513/214620892-13df493e-3d0f-4712-8bae-dce447b1049b.png)

This trigger the data to be collated (via a call to `handleExport`), and the `.xlsx` will start to download.  Note: I've done the non-UI processing in this branch #2296. I just need @lilitkarapetyan  to do convert the data to a multi-sheet Excel workbook, and handle the download button.  This is in `GameChannels.tsx`, in the method `generateFile` in branch `2294_collate_data_for_export` in PR #2296

* **Assets**: one row per asset, with these columns: 
    * uniqId
    * name
    * force-name
    * platform-type-name
    * sub-type (from attributes)
    * health
    * C2-state (from attributes)
    * location
    * other attributes
    * turn-number
    * turn start/end date
* **Adjudication Outcomes**:  this is the set of completed (details.interaction.complete === true) interaction messages. It will have 3 worksheets (`Perception`, `Health`, `Movement`), with some common columns, then type-specific fields:
    * Common 
        *  document id,
        * order1-name (lookup from details.interaction.order1)
        * order2-name
        * start time (details.interaction.startTime)
        * end time (details.interaction.endTime)
        * event (details.interaction.event)
    * Type specific (one per tab)
        * perception outcome fields _or_
        * health outcome fields _or_
        * movement outcome fields
* **Other Messages** one row per message, with one worksheet per message-type
    * document id,
    * title
    * channel (details.channel)
    * from-force (details.from.forceName)
    * from-role (details.from.roleName)
    * messageType (details.messageType)
    * timeStamp (details.timeStamp)
    * turnNumber (details.turnNumber)
    * one column per field for the contents of the `message.message` body, including arrays when necessary

IanMayo:
Solution delivered.

PR #2296, opened 2023-01-25
2294 collate data for export

IanMayo:
Backend processing for #2294 

Actually, I've put the UI in there, too. Well, at least some of it.


IanMayo:
Great, thanks @lilitkarapetyan  - I think we have some tidying to do, but this will let them get started.

Issue #2291, opened 2023-01-24
Switch to one planning-force per force


IanMayo:
For the clustered icons we show the cluster icon shaded in a color.

Currently the clusters are green/yellow/amber depending upon the cluster size.

Instead, we should to per-force clusters, and shade the clusters according to the force color.

Note: we should still not cluster `selected` assets.

Note 2: @lilitkarapetyan has mastered over-riding the cluster function here:
https://github.com/serge-web/serge/pull/2310/files#diff-0cd02d4e329c4ead1df64feffa5033f1d7088832d26603502eb94f3da176f71fR20

I guess we have these requirements:
- component receives:
    - asset-rows for this view
    - selected assets
    - filtered assets (may be contents of filtered or current assets lists)
- if subject force === player force or player is umpire then
    - collate all assets
    - also render range-rings (phase 2)
    - do not cluster selected assets
- else
    - decide perception of asset
    - render asset according to perception 
    - do not cluster selected assets

PR #2310, opened 2023-02-01
Clustering_defer_asset_icon_creation

### LINKED ISSUE #2280: Clustering - defer asset icon creation
The cluster library we're using allows the creation of the icon to be deferred:
https://github.com/Leaflet/Leaflet.markercluster#customising-the-clustered-markers

Using this, icons will only actually be created when they need to be displayed, which will simplify the component tree.

lilitkarapetyan:
Fixes #2280

### DIFF
--- a/packages/components/src/local/organisms/planning-force/index.tsx
+++ b/packages/components/src/local/organisms/planning-force/index.tsx
@@ -1,9 +1,7 @@
 import cx from 'classnames'
-import L, { LatLng, latLng, LeafletMouseEvent } from 'leaflet'
+import L, { LatLng, latLng, LeafletMouseEvent, MarkerCluster } from 'leaflet'
 import 'leaflet.markercluster/dist/leaflet.markercluster'
-import 'leaflet.markercluster/dist/MarkerCluster.css'
-import 'leaflet.markercluster/dist/MarkerCluster.Default.css'
-import React, { useContext, useEffect, useState } from 'react'
+import React, {useContext, useEffect, useState } from 'react'
 import * as ReactDOMServer from 'react-dom/server'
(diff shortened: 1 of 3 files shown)


IanMayo:
I'm away from my desk right now, so can't check the detail. But the code looks like you are providing a function to create the `cluster` icon.  I was hoping to provide a function to only create the `asset` icon when it's actually required.

But, I think I have caused the problem.  In the issue I link to the demo code to create the cluster icon - I mis-read the code.  I have posted `create icon using a function` as an issue in the MarkerCluster repo.

**Update**: ok, I've checked.  This PR generates custom marker clusters, not custom icons.

But, the code will be useful in #2291 - where we will shade the clusters according to the force.

packages/components/src/local/organisms/planning-force/index.tsx
```diff
@@ -19,9 +17,26 @@ const PlanningForces: React.FC<PropTypes> = ({ assets, selectedAssets, setSelect
   const [rawMarkers, setRawMarkers] = useState<AssetRow[]>([])
   const { assetsCache } = useContext(SupportPanelContext)
 
+  const createIcon = () => {
+    return {
+      iconCreateFunction: function (cluster: MarkerCluster) {
```
IanMayo, line 22: @lilitkarapetyan - I won't be merging this PR, since it's the Markers I wanted to defer, not the clusters.

But, I do want to create cluster markers using the existing force colors.

This code shows how to specify a color using CSS.  But, can you give me the JS code to make the marker a specific color?  eg: `#f0f`

IanMayo: No, forget that @lilitkarapetyan  - I've found how to do it :-D

PR #2315, opened 2023-02-02
2291b per force asset layers

### LINKED ISSUE #2291: Switch to one planning-force per force

IanMayo:
Fixes #2291


IanMayo:
Oh dear @lilitkarapetyan - I've got myself stuck here.

I've refactored the code so that we now do per-force clustering.  But, when we update the set of assets (after we have filtered the table) I'm sure the re-render leaves the previous icons on the map.

The expected behaviour is that when I apply a filter for the sub-type, and only two blue assets are shown in the table, then only two assets should be shown on the map.

I'm pretty sure that `PlanningForces` is drawing the initial set of 20 icons on page load (`Blue:1` to `Blue:20`, then on each change it is _ALSO_ showing the current set of icons for the current filter (which could be `Blue:377`).

screencast

Video also [here](https://i.gyazo.com/730256985a7b7216edf390fd8f4a9ce3.mp4)

I have a minor worry that we are adding the cluster group to the map too many times (line `91` below), but that doesn't seem related:
https://gyazo.com/730256985a7b7216edf390fd8f4a9ce3

lilitkarapetyan:
> Oh dear @lilitkarapetyan - I've got myself stuck here.
> 
> I've refactored the code so that we now do per-force clustering. But, when we update the set of assets (after we have filtered the table) I'm sure the re-render leaves the previous icons on the map.
> 
> The expected behaviour is that when I apply a filter for the sub-type, and only two blue assets are shown in the table, then only two assets should be shown on the map.
> 
> I'm pretty sure that `PlanningForces` is drawing the initial set of 20 icons on page load (`Blue:1` to `Blue:20`, then on each change it is _ALSO_ showing the current set of icons for the current filter (which could be `Blue:377`).
> 
> screencast [ screencast ](https://camo.githubusercontent.com/592437ad3edeba1a4e6032494c8901dd561b4425759b2ec152f83c439e3af0a4/68747470733a2f2f692e6779617a6f2e636f6d2f37333032353639383561376237323136656466333930666438663461396365332e676966) [ ](https://camo.githubusercontent.com/592437ad3edeba1a4e6032494c8901dd561b4425759b2ec152f83c439e3af0a4/68747470733a2f2f692e6779617a6f2e636f6d2f37333032353639383561376237323136656466333930666438663461396365332e676966)
> 
> Video also [here](https://i.gyazo.com/730256985a7b7216edf390fd8f4a9ce3.mp4)
> 
> I have a minor worry that we are adding the cluster group to the map too many times (line `91` below), but that doesn't seem related: https://gyazo.com/730256985a7b7216edf390fd8f4a9ce3

Hello @IanMayo, okay I will investigate it now.

IanMayo:
Thanks for the fix @lilitkarapetyan  :-D

But, I didn't like `SupportPanel` knowing about the `MarkerCluster`. I was also scared of relying on the array length to trigger updates (since the filters could change, but the number of assets displayed could stay the same).

But, you fix helped me to understand the issue (the marker cluster layer getting added too many times).  So, I've come up with this other way of fixing it.  It seems to work properly - since the marker cluster only gets added to the map once.  I had to move the `useMap()` call up into the parent class, but that still works ok.

Does this new fix look ok to you?
https://github.com/serge-web/serge/pull/2315/files#diff-0cd02d4e329c4ead1df64feffa5033f1d7088832d26603502eb94f3da176f71fR44

Thanks,
Ian

lilitkarapetyan:
Yes it looks good.

IanMayo:
Oops, I'm mistaken.  My solution isn't deleting the existing icons.  Will investigate further.

The issue was we keep on generating new MarkerCluster instances on each re-render.  So, the new fix is to look for an existing marker cluster in the `map` - and re-use that, if present: https://github.com/serge-web/serge/pull/2315/commits/0634e214ffc4e49f80f001aeeb009eb0e61ead4c#diff-0cd02d4e329c4ead1df64feffa5033f1d7088832d26603502eb94f3da176f71fR44

PR #2316, opened 2023-02-02
Refactor range ring component

### LINKED ISSUE #2269: Refactor range rings
Range rings for SAM sites are currently being rendered in a separate control for the subject assets.

Render the rings alongside the icons.  So - we show rings when the relevant icon is shown.

That _probably_ means they need to go into `PlanningForces`

Include this with #2291 

Ok, this is what I understand for this:
- we create asset icons in the `PlanningForces` component
- that component has two icon create methods:
    - `getRawMarkerOption`
    - `getClusteredMarkerOption` 
- when the icon is shown as `raw` we should certainly include the range ring. We should track the `rawMarkers` state, and create a `rawRings` array of circle shapes that we offer for rendering at the same level as `MarkerCluster`.
- when the icon is in a cluster we should not show the ring, but when the cluster is expanded, or when the marker doesn't need to be clustered, we should show the ring.  Note: I investigated this, and couldn't find a collection of which icons were shown as clustered, and which weren't :-(

IanMayo:
Fixes #2269


IanMayo:
Close. This is implemented in the other branch

Issue #2334, opened 2023-02-07
Players refer to standard areas


IanMayo:
As a player I wish to **Select a polygon from an existing list** in order to **refer to standard areas**

As players are doing high level planning of the next turn, they start referring to standard areas.

Some areas may be on land, and some on sea.  The whole current map may be split into, say 10 polygons.

While generating their orders, players may wish to select an existing polygon instead of drawing one.

So, we will introduce a `areas` array in the `WargameData` object.  Like the other elements in the `WargameData` object this will have a parent `wrapper` element called `areas` that contains an array of `ArrayData` elements.  These array elements will have these parameters:
```
  // user label for this shape
  name: string
  // which forces use this shape
  usedBy: Array<ForceData['uniqid']>
  // the shape itself (probably a Polygon, but could be line or point)
  feature: GeoJSON.Polygon
``` 

This will be supported in the `OrderDrawing` and `OrderEditing` pages.

The will receive a new prop called `areas: AreaData[]`.  This will have already been filtered by the parent class for this force.

![image](https://user-images.githubusercontent.com/1108513/217224141-dff9db1d-9ba8-4df9-aba1-56d751a4a6e6.png)

If the next shape to be drawn is a polygon, then a new drop-down list will appear next to the `Cancel editing` button.  This will contain a list of areas for this force. Selecting one will add that polygon to the set of shapes, and move onto the next one.

IanMayo:
Ok. I've got a long way with this (#2336), but am now stuck. 
![screencast](https://i.gyazo.com/99252e01cdc7bd790fe9407d1c7f6ec3.gif)

I'd appreciate some help with these bits, please @lebaphi :
* [x] For `StandardAreaMenu` move `select` control to the top-left, but place it vertically beneath the `cancel planning` button. Note: if we use `position:topLeft` they get stacked on each other
* [x] In `OrderDrawing`, if player selects item from menu, cancel planning polygon
* [ ] In `OrderDrawing`  generate planning activity using selected polygon
* [x] In `OrderDrawing` it looks like `GeoMan` is intercepting the `StandardAreaMenu` click, and treating it as the start of the polygon
* [ ] Aaah, err, in the current implementation we get a new control each time we create new orders. I'm sure it should be deleted 
* [ ] Also try to allow use of `Standard Areas` in `OrderEditing`, too
* [ ] Swap the `Standard Area Menu` popup to use some other icon. Shapes? The letters `S/A`?

[...]

PR #2336, opened 2023-02-07
2334 standard areas

### LINKED ISSUE #2335: View standard areas
We should let players view the set of standard areas.

In `PlanningChannel` add a new layer component to view standard areas.

If the current force has any standard areas (`myAreas && myAreas.length`) then add a toggle button to the right-hand toolbar that toggles the display of standard areas.

IanMayo:
Supports #2334 
Fixes #2335


IanMayo:
Thanks for the CSS fix @philb89 

Could you also apply the fix so it works in Storybook too, please?

It's this story, but the `areas` select control still sits under the `Cancel planning` button:
http://localhost:6611/?path=/story/local-organisms-planningchannel--with-areas

I'd also appreciate your help with the other bits, please :-)

PR #2360, opened 2023-02-10
Players refer to standard areas

### LINKED ISSUE #2334: Players refer to standard areas

lebaphi:
Fixes #2334 

* [x] For `StandardAreaMenu` move `select` control to the top-left, but place it vertically beneath the `cancel planning` button. Note: if we use `position:topLeft` they get stacked on each other
* [x] In `OrderDrawing`, if player selects item from menu, cancel planning polygon
* [x] In `OrderDrawing`  generate planning activity using selected polygon
* [x] In `OrderDrawing` it looks like `GeoMan` is intercepting the `StandardAreaMenu` click, and treating it as the start of the polygon
* [x] Aaah, err, in the current implementation we get a new control each time we create new orders. I'm sure it should be deleted 
* [x] Also try to allow use of `Standard Areas` in `OrderEditing`, too
* [x] Swap the `Standard Area Menu` popup to use some other icon. Shapes? The letters `S/A`?


IanMayo:
Hello @philb89  - thanks for helping move this forward.

I now realise that I don't think we can use `standard` areas in `Order Editing`.  When a player is creating a new set of orders `OrderDrawing` - we know the steps when a polygon could be used, and we can offer the list of standard areas at those steps.  But, with `Order Editing`, the player is looking at the full set of shapes for that set of orders, and we don't know which shape should be swapped for the standard area.

Hmmm, if there is only one polygon for the set of shapes, then we can swap that polygon for the standard area.  

Aah, ok.  We only show the `Standard Areas` menu in `Order Editing` if there is just one polygon in the set of shapes. I can implement that :-)

One last bit:
image

We're not quite moving the logic forward properly.   I'm in this story:
https://serge-phileba-2334-stan-raxggo.herokuapp.com/storybook/?path=/story/local-organisms-planningchannel--with-areas

I select a blue icon form the map, so it's selected, and flashing. Then I click the button to start the process of creating a new set of orders. I select "Land" and "Patrol".  I asks me to enter the line. I enter the line. Then it wants a polygon, so I select one of the standard areas.  Then - it should ask me to enter the line for the "Route back".  If you add a polygon (instead of selecting the existing area) - then you will see this happen.   But, in the current logic, things _stop_ when I select an existing area.

BTW - fixing the above (moving edit process forward) is more important than `OrderEditing` right now.  In tomorrow's rehearsal I'd like players to start creating orders using standard areas - but we need the process to work for that to happen  :-)

PR #2418, opened 2023-02-15
2391: Quick way of clearing filters

### LINKED ISSUE #2391: Quick way of clearing filters
It can take quite a few mouse-clicks for players to clear filters.

We _could_ add a `Clear filters` button. But,  a quicker way of implementing this would be to let the users cycle through `Hide filters` and `Show filters`.  We should intercept the `Hide filters` button, and insert code to clear any filters in place.

So. to clear all filters for a table, the player would click on `hide filters` then `show filters`, and any filters will have been cleared.

lebaphi:
Fixes #2391


IanMayo (approved):
Tested & working

PR #2428, opened 2023-02-16
2380: Op-Boxes in categories

### LINKED ISSUE #2380: Op-Boxes in categories
We've realised there are about 60 standard areas - so we need to organise them a little better. So, they are being grouped into categories - with color shading.  I've changed our mock data to fit the new `AreaCategory` data structure.

* [x] introduce `category` of area includes name, color, list of areas
* [ ] render select menu using tiered data

Over to you now please @lebaphi 

The `StandardAreaMenu` now shows a single list of area items, each with the category name.  But, we really want a tiered menu.  I was able to implement the tiered menu, but the menu-click was failing.  I think you used a trick [here](https://github.com/serge-web/serge/pull/2336/files#diff-e7637e7308bb915e168456de488b9acb633b7585f9ba0420a6426bd6bd667603R284) to prevent that.

lebaphi:
Fixes #2380

### DIFF
--- a/packages/components/src/local/organisms/planning-channel/helpers/StandardAreaMenu.tsx
+++ b/packages/components/src/local/organisms/planning-channel/helpers/StandardAreaMenu.tsx
@@ -49,31 +49,17 @@ const StandardAreaMenu: React.FC<StandardAreaMenuProps> = ({ areas, handler, sho
 
   /** generate the tree of activities */
   const getItems = (areas: AreaCategory[]): SelectItem[] => {
-    // TODO: this produces nested list, but I can't click them :-(
-    // const aItems = areas.map((category) => {
-    //   return {
-    //     label: category.name,
-    //     value: category.name,
-    //     items: category.areas.map((area) => {
-    //       return {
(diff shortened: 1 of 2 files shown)


packages/components/src/local/organisms/planning-channel/helpers/StandardAreaMenu.tsx
```diff
@@ -49,40 +49,24 @@ const StandardAreaMenu: React.FC<StandardAreaMenuProps> = ({ areas, handler, sho
...
-    //       }})
-    //   }
-    // })
-    // return aItems
-
-    const flatList: SelectItem[] = []
-    areas.forEach((category) => {
-      // note: we had a crash when Serge encountered an areas block in the "old" structure.
-      // note: so, check the areas element exists
-      category.areas && category.areas.forEach((area) => {
```
IanMayo, line 69: @lebaphi  - could you include the above check that `category.areas` exists before running on the map please? 

There is a legacy wargame on the client network with the `previous` structure for areas.  We don't need to render it, but I don't want it to crash Serge (like it did on Tuesday morning :-o )

PR #2430, opened 2023-02-16
2379: Fix compile error

### LINKED ISSUE #2379: Material table for own forces has two sliders
The real data has really long platform names, like: `UK F-35A Lightning CTOL_Sqn_5`

With these long names we have two scroll bars in material table. 

Will try to generate some mock data to reproduce this.

We probably just need to block use of scroll bar on _Some_ element on the page.

I did think this was related to long platform names.  But, I now see it happens with existing data in develop.  To make the unwanted scrollbar appear, just keep making the `Own Forces` panel wider and wider, and it will appear:
image

lebaphi:
Fixes #2379

### DIFF
--- a/packages/components/.stylelintrc
+++ b/packages/components/.stylelintrc
@@ -3,15 +3,16 @@
     "stylelint-config-sass-guidelines",
     "stylelint-config-css-modules",
     "stylelint-config-prettier"
-    ],
+  ],
   "plugins": [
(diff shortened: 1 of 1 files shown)


IanMayo:
Closing this PR, fixed in this commit: https://github.com/serge-web/serge/pull/2426/commits/9f0e39fc768cd3eec06086515455b53b66250828

Issue #2432, opened 2023-02-16
Performance Problem with `Select Assets`


IanMayo:
We are experimenting with increasing the number of rows displayed in the `PlanningAssets` table:
![image](https://user-images.githubusercontent.com/1108513/219419096-a7d25de2-377a-40cf-a57b-41db19c14298.png)

But, with 500 rows visible, it can take 5+ Seconds for a row to be marked as `Selected`.  See the video from phi here:
https://github.com/serge-web/serge/issues/2386#issuecomment-1432997117

I _think_ the delay is related to the icons on the map.  Could you investigate please, and see if you can find the bottleneck?

Issue #2386, opened 2023-02-14
Allow longer list of orders


IanMayo:
The Order tab (Planning Messages List) shows at most 50 rows.  Allow 100 and 250 rows (check if performance of assets table is ok with 250 rows).

IanMayo:
Related to #2384

lebaphi:
Hi @IanMayo, I tried to add more 150, 200 and 250 rows and here is the result:

https://user-images.githubusercontent.com/26705621/219356389-e55a9912-897b-4fff-b7c0-b802529f39c4.mov


As you can see, the more rows on the table, the longer it takes to click on 1 content (200 content ~ 5-6 seconds). I think there might be a performance issue here 🤔

IanMayo:
I'm pretty sure the delay is map-related.   If I switch off firing out the selection change, it's always under a second or so:
![image](https://user-images.githubusercontent.com/1108513/219367743-2f166d3b-a3b3-4319-ac90-44cdd1475a91.png)

Let's treat the map delay as a separate issue for now.  So could you please:
- (temporarily) stop firing the external `onSelectionChange`
- investigate if/where there is a performance delay in this table - see if we're unnecessarily calling something? getting icons from the network? 
- if the table is set to more rows than are present, empty <tr>` are created that appear as empty white-space.  So please (fancy css?) find a way of not displaying empty `<tr>`s.


![image](https://user-images.githubusercontent.com/1108513/219368170-b0abbd5d-ae13-4ebc-859e-f2d9abe341aa.png)

IanMayo:
Done.

PR #2437, opened 2023-02-17
2432: Improve performance when selecting asset

### LINKED ISSUE #2384: Increase default number of rows for assets page
For the Planning Assets table, use a default of 100 assets.

Check table performance with 250, 500, 1000 items.

@lebaphi has established that the performance really slows down in the assets table with lots of rows.

I've been playing and have (slightly) reduce the number of events that get triggered from the table.

I have a suspicion the table is doing lots and lots of rendering.  We have the `render-icon` and `render-platform` - which both do a bit of processing.

lebaphi:
Fixes #2384 #2386 #2432 

Hi @IanMayo, I just made some changes to reduce rendering, it seems the performance has improved (it only takes 1~2 seconds when this table loads 250 rows)

https://user-images.githubusercontent.com/26705621/219589492-360f1a60-bfd5-4e76-aea5-6ca28c2a37bf.mov

But I found that the material table itself has an performance issue when enable/disable the filter, it makes 3~5s to applied, I'm investigating it.


IanMayo:
Brilliant!

It looks like it needs a bit more review - will do that later.

IanMayo:
@lebaphi - you've lost some of my changes 🤣 

For the material tables, I have added a new tool button for `planning-messages`, and provided two-state icons for `planning-messages` and `planning-assets`.  The original bits are in red in `Files Changed`.

packages/components/src/local/organisms/planning-channel/index.tsx
```diff
@@ -366,23 +370,16 @@ export const PlanningChannel: React.FC<PropTypes> = ({
...
-    // only update selected assets if we're not planning an activity
-    if (!activityBeingPlanned) {
-      setSelectedAssets(localSelectedAssets)
-    }
-  }, [localSelectedAssets])
-
   useEffect(() => {
     // produce the own and opp assets for this player force
     const forceCols = getForceColors(allForces)
     const platIcons = platformIcons(platformTypes)
```
IanMayo, line 376: hello @lebaphi -  your code has dropped the extra parameter for the next two calls - that will prevent some functionality from working.

lebaphi:
Close due to this PR https://github.com/serge-web/serge/pull/2444

PR #2441, opened 2023-02-19
2416: Persistent Planning Channel layout/status

### LINKED ISSUE #2416: Persistent Planning Channel layout/status
* [x] Current tab
* For tables:
    * [x] Sort on any columns
    * [x] Number of rows visible

lebaphi:
Fixes #2416 

* [x] Current tab
* For tables:
    * [ ] Whether filters are shown
    * [ ] Which filters are applied
    * [ ] Sort on any columns

### DIFF
--- a/packages/components/src/local/organisms/planning-assets/index.tsx
+++ b/packages/components/src/local/organisms/planning-assets/index.tsx
@@ -1,9 +1,12 @@
 import { faSearchMinus, faSearchPlus } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import MaterialTable, { Column, MTableBody, MTableBodyRow } from '@material-table/core'
+import { expiredStorage } from '@serge/config'
 import cx from 'classnames'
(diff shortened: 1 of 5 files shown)


IanMayo:
I just tried to view the deployment for this, but it's crashing with this error:
image

Hmm, you haven't changed `ArrowHead` - it's probably nothing to worry about.  No. I've checked out the branch - and it's failing with same error (but with a little more detail in the trace):
image

Very strange 😮

lebaphi:
> I just tried to view the deployment for this, but it's crashing with this error: image
> 
> Hmm, you haven't changed `ArrowHead` - it's probably nothing to worry about. No. I've checked out the branch - and it's failing with same error (but with a little more detail in the trace): image
> 
> Very strange 😮

yep, I also see this error, and I'm not sure the root cause, I don't touch this file 🤣

IanMayo:
> yep, I also see this error, and I'm not sure the root cause, I don't touch this file 🤣

I _think_ this is a case where you need a fresh branch off develop, and make your changes one at a time - and track when the problem starts.

lebaphi:
Close due to https://github.com/serge-web/serge/pull/2446

PR #2446, opened 2023-02-19
2416: Persistent Planning Channel layout/status

### LINKED ISSUE #2416: Persistent Planning Channel layout/status

lebaphi:
Fixes #2416 

* [x] Current tab
* For tables:
    * [ ] Whether filters are shown
    * [ ] Which filters are applied
    * [ ] Sort on any columns

### DIFF
--- a/packages/config/src/globals.ts
+++ b/packages/config/src/globals.ts
@@ -192,3 +192,11 @@ export const CONTROL_ALL = 'control-all:'
  */
 export const UNSENT_CHAT_MESSAGE_TYPE = 'chat'
 export const UNSENT_SELECT_BY_DEFAULT_VALUE = 'BY_DEFAULT_VALUE'
+
+export const SUPPORT_PANEL_LAYOUT = {
+  OPENING_TAB: 'opening_tab',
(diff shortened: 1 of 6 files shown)


packages/config/src/globals.ts
```diff
@@ -191,4 +191,12 @@ export const CONTROL_ALL = 'control-all:'
  * @param event - name of event that just happened
  */
 export const UNSENT_CHAT_MESSAGE_TYPE = 'chat'
-export const UNSENT_SELECT_BY_DEFAULT_VALUE = 'BY_DEFAULT_VALUE'
\ No newline at end of file
+export const UNSENT_SELECT_BY_DEFAULT_VALUE = 'BY_DEFAULT_VALUE'
+
+export const SUPPORT_PANEL_LAYOUT = {
```
IanMayo, line 196: @lebaphi - could you change these constants by putting `serge.planning` at the front of the string?  

That may help us navigate the cookies in the future.

lebaphi: sure @IanMayo, I will update in part 2 of this PR 😄

PR #2501, opened 2023-02-24
2366: When adding a Standard Area to a planning order - show the area

### LINKED ISSUE #2366: When adding a `Standard Area` to a planning order - show the area
For polygons in orders, players can either add them by drawing them, or add them from the drop-down list.

If they draw them, they are able to see the shape on the map.  But, if they select it from the list it isn't shown.

It would be good if _Somehow_ we could add the new shape to the drawing layer - so it gets drawn and the player can review what they are doing.  It also helps the player plan their `route back`.

We _probably_ have to push the `Layer` for the new shape into the GeoMan layer in `OrderDrawing`.   Hmm, if we do that, _maybe_ we no longer need to insert the shape ourselves - maybe the `GeoMan`  handler will jump onto it.

lebaphi:
Fixes #2366


IanMayo:
Hey - this is scary stuff @philb89  - well done :-)

Can you explain why we need two copies of `leaflet.select.js`?  Is it just for storybook to work?

IanMayo:
Ooh, we're not quite there.

I just started to create orders for Land > Patrol.  I drew the `route out`, selected a standard area (which appeared on the map :-) ), but then it didn't invite me to draw the `route back`.  This _used_ to work.

I think we need to create a `pendingGeometries` object to move things forward, and `onCreate` used to handle that.

lebaphi:
> Hey - this is scary stuff @philb89 - well done :-)
> 
> Can you explain why we need two copies of `leaflet.select.js`? Is it just for storybook to work?

Hi @IanMayo, I update the leaflet.select.js to support new property `showSelectedIcon: true`

image

then we can determine which area is selecting.

lebaphi:
> Ooh, we're not quite there.
> 
> I just started to create orders for Land > Patrol. I drew the `route out`, selected a standard area (which appeared on the map :-) ), but then it didn't invite me to draw the `route back`. This _used_ to work.
> 
> I think we need to create a `pendingGeometries` object to move things forward, and `onCreate` used to handle that.

ah sorry, I think I forgot to uncomment these lines 🤣

IanMayo:
> then we can determine which area is selecting

I noticed the change.  I'm pretty sure we don't need to display the radio button.  We're treating this as a "command button" menu - that ends with a click.  We're not providing it to allow a user to investigate a tree of data.

Also:
> Can you explain why we need two copies of leaflet.select.js? Is it just for storybook to work?

lebaphi:
> > then we can determine which area is selecting
> 
> I noticed the change. I'm pretty sure we don't need to display the radio button. We're treating this as a "command button" menu - that ends with a click. We're not providing it to allow a user to investigate a tree of data.
> 
> Also:
> 
> > Can you explain why we need two copies of leaflet.select.js? Is it just for storybook to work?

okay, let me update

IanMayo:
One remaining issue, please.

The new rectangle is left drawn on the map when we finish creating the shapes.  I think we need to (optionally) remember it in state, and delete it when we finish.

Demo:  https://i.gyazo.com/ff89650eea2207523dc874c55fda21a6.mp4

IanMayo (approved):
Good solution, thanks.

IanMayo:
Aah, hold on. There is a remaining issue.  We need to remove the `Standard Area` select control when the standard area has been selected.

Demo here: https://i.gyazo.com/01a67d7620e6c037fb74acd3d2537621.mp4

Could you also check that the `standard areas control` and the `standard polygon` are removed if the player `Cancels` planning the orders?

IanMayo (changes requested):
Changes requested [here](https://github.com/serge-web/serge/pull/2501#issuecomment-1445033866) please.

lebaphi:
> Changes requested [here](https://github.com/serge-web/serge/pull/2501#issuecomment-1445033866) please.

sure @IanMayo, I will update after dinner 👍🏻

IanMayo:
Working fine :-)

IanMayo:
Sorry @lebaphi -  I think this still isn't quite right.  As soon as we have planned the polygon (either by selecting a standard area, or drawing a polygon) - then the standard-areas menu should disappear. I know we don't want it there when the `route back` line is being drawn.  The logic to control this is in the `standardPolygons` state.  When this is zero or undefined, the we should not show the standard area menu.  Looking at it another way, we should only show the standard area menu when `standardPolygons` is contains items.

IanMayo:
Working perfectly :-)

Issue #2467, opened 2023-02-22
Run-time console errors/warnings


IanMayo:
We currently have some react errors and material-table warnings as Serge starts.

We should address the technical debt, and fix them.

Maybe it's best for @lilitkarapetyan  to do the `react` ones, and @philb89  to do the `material-table` ones.

```
Warning: React has detected a change in the order of Hooks called by ChannelTabsContainer. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://fb.me/rules-of-hooks

   Previous render            Next render
   ------------------------------------------------------
1. useState                   useState
2. useState                   useState
3. useRef                     useRef
4. useContext                 useContext
5. useContext                 useContext
6. useState                   useState
7. useState                   useState
8. useEffect                  useEffect
9. useEffect                  useEffect
10. useEffect                 useEffect
11. undefined                 useContext
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    in ChannelTabsContainer (at GameChannels.tsx:169)
    in div (at GameChannels.tsx:168)
    in div (at GameChannels.tsx:166)
    in GameChannels (at GameChannelsWithTour.tsx:169)
    in div (at GameChannelsWithTour.tsx:168)
    in div (at GameChannelsWithTour.tsx:167)
    in GameChannelsWithTour (at PlayerUi.tsx:117)
    in PlayerUi (created by Connect(PlayerUi))
    in Connect(PlayerUi) (at PlayerUiWrapper.jsx:9)
    in PlayerStateProvider (at PlayerUiWrapper.jsx:7)
    in PlayerUiWrapper (at RouterMain.jsx:14)
    in RouterMain (at App.jsx:20)
    in ThemeProvider (at App.jsx:19)
    in Provider (at App.jsx:18)
    in App (at src/index.jsx:5)
console.<computed> @ index.js:1
overrideMethod @ react_devtools_backend.js:2655
printWarning @ react-dom.development.js:88
error @ react-dom.development.js:60
warnOnHookMismatchInDev @ react-dom.development.js:14715
updateHookTypesDev @ react-dom.development.js:14674
useContext @ react-dom.development.js:15895
useContext @ react.development.js:1493
usePlayerUiDispatch @ PlayerUi.tsx:46
factory @ factory.tsx:32
ChannelTabsContainer @ ChannelTabsContainer.tsx:73
[...]

PR #2504, opened 2023-02-24
run_time_console_errors_warnings

lilitkarapetyan:
Supports #2467


lilitkarapetyan:
I fixed the `React has detected a change in the order of Hooks called by ChannelTabs Container` error

IanMayo:
> I fixed the `React has detected a change in the order of Hooks called by ChannelTabs Container` error

ok - so this PR just covers one of the errors, right?

IanMayo (approved):

lilitkarapetyan:
> > I fixed the `React has detected a change in the order of Hooks called by ChannelTabs Container` error
> 
> ok - so this PR just covers one of the errors, right?

Yes.

PR #2511, opened 2023-02-25
2487: Filter only updated the second time it is opened

### LINKED ISSUE #2487: Filter only updated the second time it is opened
I've implemented `dependent filters`.

This means - when a player applies a filter, the other filters get updated to only show values that are currently present in the table.

It works.  But, the new filter values are only shown for the second time the filter is opened.

screencast

lebaphi:
Fixes #2487


IanMayo:
Aah, that `timeout` is a useful fix.

But @lebaphi  - I don't think we need the `custom-filter-row` any more.  I've written alternate code that populates the lists from the current data - it's in the `visibleColumns` state handler in `planning-assets`

So, we don't need the `sub-type` handling from that component.  But, we _may_ need the filter cache-ing.  Would you mind investigating please?

lebaphi:
sure @IanMayo

IanMayo:
@lebaphi  - could you also have a think about our user of the custom-filter-row please?  

I think the code in `planning-assets` is more quicker and more accurate, since it only offers values in filters that relate to the data shown in the table.

lebaphi:
Yep Ian, I prefer to use the default filter too, it would better than the customize filter which might be slower (depending on our implementation)

lebaphi:
Hi @IanMayo , I removed the custom filter for the subtype, keeping only the filter caching on the `CustomFilterRow` element. So about this `Filter is only updated when it is opened for the second time`, I guess this is a matter of the material table library. The data looks correct, but the table doesn't re-render when the column data changes, it only re-render when open filter for 2nd time. I'm investigating on this.

https://user-images.githubusercontent.com/26705621/221418868-c1dfcd29-24dd-4e44-8531-73daeda38f56.mov

IanMayo:
Yes, that's the problem we 're addressing

>  I'm investigating on this.

But, I thought your `setTimeout` fix yesterday fixed it.  Didn't it?

Just checked.  In yesterday's build, the sub-type filter does update properly.  Hmm, maybe the way of updating columns in  `custom-filter` is  better than the way I'm updating them in `planning-assets`.  Could you investigate please?
screencast

lebaphi:
I've pushed new fixes, can you check this @IanMayo :D

IanMayo:
Fix works :-)

I don't know how, but it works :-D

lebaphi:
> Fix works :-)
> 
> 
> 
> I don't know how, but it works :-D

ahh, I just call setLocalProps inside custom-filter-row component to make it re-render 😂

PR #2535, opened 2023-02-28
2533: Filter changes (store tab and role details)

### LINKED ISSUE #2533: Filter changes (store tab and role details)
1. When we remember our filters, we need to remember which tab they relate to.   

Steps to reproduce:
* log in as blue player
* leave the `Own forces` filters empty
* set a `Type` filter for `Other Forces`
* refresh the page.  See the filter is applied to own forces

I suspect we need to introduce multiple `filter` cookies, that include the name of the tab.  From within the `PlanningAssets` component, the `opFor` boolean flag can be used to determine if it is the `Own Forces` or `Other Forces` tab.

2.  When we remember Support Panel preferences, remember the role they related to.

Steps to reproduce:
* log in as `Adjudicator 1`
* select `Adjudiction` tab
* switch to `Blue` user
* note the app opens with the `Adjudication` tab open - which should not happen

lebaphi:
Fixes #2533

### DIFF
--- a/packages/components/src/local/organisms/planning-channel/index.tsx
+++ b/packages/components/src/local/organisms/planning-channel/index.tsx
@@ -765,12 +765,13 @@ export const PlanningChannel: React.FC<PropTypes> = ({
   }
 
   const onSupportPanelLayoutChange = (key: string, value: string) => {
-    const panelState = JSON.parse(expiredStorage.getItem(selectedRoleId) || '{}')
+    const refKey = `support.planning.${selectedRoleId}`
+    const panelState = JSON.parse(expiredStorage.getItem(refKey) || '{}')
     panelState[key] = value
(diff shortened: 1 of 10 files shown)


packages/components/src/local/organisms/planning-channel/index.tsx
```diff
@@ -765,11 +765,15 @@ export const PlanningChannel: React.FC<PropTypes> = ({
   }
 
   const onSupportPanelLayoutChange = (key: string, value: string) => {
```
IanMayo, line 767: hello @philb89 :-) 
I Iike the new strategy, but can we make the key "support.planning.[rolename]", then we just call the children "filterApplied", "supportPanelWidth" and "visibleColumns".

lebaphi: sure Ian 👍🏻

IanMayo (approved):

PR #2538, opened 2023-03-01
enter_location_for_movement_outcomes

### LINKED ISSUE #2107: Enter location for `movement` outcomes (right-click handler)
One of the potential outcomes of an interaction is a `movement` - it's particularly valuable in `Transit` activities.

**Note:** we have a temporary workaround for this, in #2198 / #2189 

Here is an alternate way of supporting this:
- a right-click on the map silently stores the lat/long in an array on the clipboard.  This can then be pasted into the box

If we can't write to the clipboard (because our app won't be running on an `https` connection - so the browser won't allow clipboard access), then let the user copy it themselves by making the right-click handler popup the coords in an alert:
- https://stackoverflow.com/a/6055620/92441

**Note:** it _looks-like_ this is how we catch events with `React-LEaflet`: https://react-leaflet.js.org/docs/api-map/

lilitkarapetyan:
Fixes #2107

### DIFF
--- a/packages/components/src/local/organisms/planning-channel/index.tsx
+++ b/packages/components/src/local/organisms/planning-channel/index.tsx
@@ -5,7 +5,7 @@ import {
 } from '@serge/custom-types'
 import { clearUnsentMessage, forceColors as getForceColors, ForceStyle, getUnsentMessage, platformIcons, saveUnsentMessage } from '@serge/helpers'
 import cx from 'classnames'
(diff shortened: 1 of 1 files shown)


IanMayo:
thanks @lilitkarapetyan  - I've made a change, to show the cursor location, rather than my location 🤣

IanMayo (approved):

PR #2542, opened 2023-03-03
missile_lists_dont_show_headers_properly

### LINKED ISSUE #2539: Missile lists don't show headers properly
We're using a JS snippet to convert headers (entries beginning with '###) into opt-groups, so they are not selectable.

The demo is here, but see the example below: https://pmk65.github.io/jedemov2/dist/demo.html

image

But, we don't transform the data.  The code to do it is in `json-editor/index.tsx`

image

My _guess_ is that we do the JS trick when the form loads, but when we're starting with a blank form, the list of own assets is empty, so there is no select control to transform.

Maybe we need to intercept the "row-add" event, somehow.

Aah, yes - the above is right.  When we `Edit` an existing "Strike" document, the select control is already generated, and gets transformed as below:
image

Aah, but if we added new rows in the `Edit` form, they would not get adapted.

lilitkarapetyan:
Fixes #2539

### DIFF
--- a/packages/components/src/local/molecules/json-editor/index.tsx
+++ b/packages/components/src/local/molecules/json-editor/index.tsx
@@ -86,6 +86,27 @@ export const JsonEditor: React.FC<Props> = ({
     const newDoc = modifyForSave ? modifyForSave(fixedDate) : fixedDate
     if (!isEqual(JSON.stringify(newDoc), originalMessage)) {
       storeNewValue && storeNewValue(newDoc)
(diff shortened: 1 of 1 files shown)


IanMayo:
@lilitkarapetyan - can you tell me which issue this is fixing, please?

lilitkarapetyan:
@IanMayo editor change event triggers after clicking on "Add Asset" button, so I do same actions during that event.

IanMayo (approved):
Works great

PR #2552, opened 2023-03-06
2416: Persistent Planning Channel layout/status (p3)

### LINKED ISSUE #2416: Persistent Planning Channel layout/status

lebaphi:
Fixes #2416


IanMayo (dismissed):
Tested - works

IanMayo (approved):
Works fine

PR #2565, opened 2023-03-07
2558_Improved_select_control_for_asset_lists

### LINKED ISSUE #2558: Improved `<select>` control for asset lists
Players are complaining about how difficult it is to find assets in the "own force" and "opp force" assets lists.

They would like to be able to search for the asset they need.

The `json-editor` page lists several components that can be used for the `select` control: https://github.com/json-editor/json-editor#optional-requirements

We should find one that works with our snapshot of JSON-editor, and change the template to use it.

I guess these are the requirements
- allow user to start typing asset name, show matching entries
- only allow single selection
- allow for separate array item `value` and `label`.

These are probably the steps:
1. Find working example of third-party select control 
2. Integrate new control into story:
    a. `New Message` - `Land Activity` has a suitable control
    b. Open the above story, and find the `JustLandActivity` story
    c. Find the `Land Activity` template that is used in this story .  It's in `planning-message-templates.mock.ts`.  The select control with separate value and label starts at line `935`.
    d. Push the branch to a PR, so Ian can show the deployed version to the players
3. Ian can then update real templates to use this new control

lilitkarapetyan:
Fixes #2558


IanMayo:
Hello @lilitkarapetyan - thanks for this.

I tried to use it in a real game, but the new control isn't appearing. 

I've modified the wargame to specify `selectize` for the `Maritime--Patrol` template, so it looks like this:
image

I've pushed this updated wargame to the branch.

Steps to reproduce:
- log in as blue player, in planning phase
- select a blue asset
- select the `Maritime` -- `Patrol` orders
- draw the necessary lines & shapes
- the Maritime Patrol template will open
- `Own Assets` should use `selectize`

lilitkarapetyan:
Hello , I will work on it now.

IanMayo:
Nice work. I've tried the current control, but it only shows the first two matching elements:
image

Is it configurable to show more possible answers?

lilitkarapetyan:
> 

Yes, I am working on it now.

IanMayo:
@lilitkarapetyan  - when I run this on the Heroku deployment, and I create "maritime patrol", I don't get selectize.  I see you have a commit to "update wargamedb".  Is there a chance that this change has removed the custom template I created for `maritime patrol`

Hmm I don't see `selectize` for maritime-patrol in storybook either:
image

lilitkarapetyan:
> @lilitkarapetyan - when I run this on the Heroku deployment, and I create "maritime patrol", I don't get selectize. I see you have a commit to "update wargamedb". Is there a chance that this change has removed the custom template I created for `maritime patrol`
> 
> Hmm I don't see `selectize` for maritime-patrol in storybook either: image

No I didn't update the wargames doc yesterday I just added the one you made

IanMayo:
The list is extended - great.

But, I the "text input" box isn't there - so users can't start typing in the name to get to the item of interest:
image

In testing it, I've seen that "Land - Strike" uses the Selectize control - that's quicker to open than "Land Patrol", since it doesn't require shapes.

lilitkarapetyan:
I'll take a look at it.

IanMayo:
@lilitkarapetyan  - I still don't see the box where I can start typing the item name into:
image

lilitkarapetyan:
Hello @IanMayo. I have checked and it works. You can write in the selection area.
Screenshot_2

I have also checked with "View deployment" button.

Issue #2595, opened 2023-03-09
Performance logging


IanMayo:
* remove logging for linePolygon interactions
* insert time logging around get next interaction

IanMayo:
Done

PR #2596, opened 2023-03-09
2595 interaction performance

IanMayo:
Supports #2595


IanMayo:
Interaction processing has speeded up due to removing lots of activity interactions.  But, the diagnostics in this issue should be useful.

PR #2614, opened 2023-03-10
trial trimming interactions to recent ones

### LINKED ISSUE #2606: Fix performance lag for "Get next"
Last night the performance of "Get next" dragged out to 40 seconds.

The team couldn't resolve the performance issue.

I ran an experiment this morning to remove adjudications recorded more than 24 hours ago. The delay went from 40 seconds to 3 seconds.  This clean-up can happen repeatedly, and should reliably resolve the issue.

For each interaction to be considered, we check it's id against all previous interactions. As that list gets longer, it takes longer to process.

But, before adopting that strategy - experiment with sorting the list, then using a binary search.

IanMayo:
Fixes #2606


IanMayo:
nope, this didn't fix it.

PR #2617, opened 2023-03-11
2129: Drop-down lists for columns in Orders table

### LINKED ISSUE #2129: Drop-down lists for columns in `Orders` table
For the `Orders` table we just have free-text filters for the columns. Some of these should be drop-down lists - particularly `author/role` and `activities`.

The code is in `PlanningMessagesList`.  Well ,it's in the call to `genData.toColumn()`.  I can see that we're collating the data for the `lookup`, but they don't seem to work in the app.

Can be tested in this story:
http://localhost:6611/?path=/story/local-organisms-planningmessageslist--default

lebaphi:
Fixes #2129

### DIFF
--- a/packages/components/src/local/organisms/adjudication-messages-list/index.tsx
+++ b/packages/components/src/local/organisms/adjudication-messages-list/index.tsx
@@ -11,6 +11,7 @@ import { ADJUDICATION_OUTCOMES, INTER_AT_END, INTER_AT_RANDOM, INTER_AT_START, P
 import { Asset, ForceData, InteractionDetails, INTERACTION_SHORT_CIRCUIT, LocationOutcome, MessageAdjudicationOutcomes, MessageDetails, MessageInteraction, MessagePlanning, MessageStructure, PerceptionOutcome, PlannedActivityGeometry, PlannedProps } from '@serge/custom-types'
 import { findAsset, findForceAndAsset, forceColors, ForceStyle, formatMilitaryDate, hexToRGBA, incrementGameTime } from '@serge/helpers'
 import { area, length, lineString, LineString, polygon, Polygon } from '@turf/turf'
+import cx from 'classnames'
 import dayjs, { Dayjs } from 'dayjs'
 import { Geometry } from 'geojson'
 import _ from 'lodash'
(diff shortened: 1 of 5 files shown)


IanMayo:
Hello @philb89 - thanks for this.

We're heading for some performance problems - there are currently around 5000 documents in the database (orders and adjudications).  So, I need to be wary of introducing something that introduces a performance problem.  Can you explain how the `needToUpdate` works please?

Aah, @lebaphi - DeepScan thinks the logic isn't working as expected.

lebaphi:
yep @IanMayo, if the lookUp has changed, this function returns true and we can update the table header.

packages/components/src/local/organisms/adjudication-messages-list/index.tsx
```diff
@@ -330,24 +333,28 @@ export const AdjudicationMessagesList: React.FC<PropTypes> = ({
...
-          { title: 'Order 1', field: 'order1', render: (row: AdjudicationRow) => renderOrderTitle(true, row) },
-          { title: 'Order 2', field: 'order2', render: (row: AdjudicationRow) => renderOrderTitle(false, row) },
-          { title: 'Activity', field: 'Reference' },
-          { title: 'Duration', field: 'period' }
-        ]
+      console.time('LLOG_Collate Adjudicate Message Filters')
+      const umpireForce = forces.find((force: ForceData) => force.umpire)
+      // TODO: the column definitions should use the data collated in the column summary (below)
+      // provide more sophisticated column definition lookups
+      const summaryData = umpireForce && getColumnSummary(forces, umpireForce.uniqid, false, [])
```
IanMayo: @lebaphi  - I think that `getColumnSummary` could be expensive.  Can we only run it once we know we need to update the columns?
Hmm, no we can´t.  Never mind.

IanMayo: Hello @philb89 - I've just been having a look at the above code.  We don't actually _use_ the `summaryData` above.  The columns aren't dependent on the data, and the only `lookup` we use is the `Y/N` for `important`.

So, the `needToUpdate` will never change.

I think that means that we only need to run this once, on application load.

Does that look right?

(I want to try to improve the performance, but I'm nervous of breaking something 🤣  )

lebaphi: okay @IanMayo. if the table data does not change, we can ignore it 😄

PR #2655, opened 2023-03-26
2508: Co-located markers don't show as selected

### LINKED ISSUE #2508: Co-located markers don't show as selected
When a marker is `selected`, we don't cluster it.

But, when markers are co-located, we do cluster them, to prevent them stacking.

This means we do see co-located selected assets.

We need to modify `planning-force` to make clusters pulse if they contain one or more selected assets.

In `planning-force`, this function generates cluster icons: `iconCreateFunction`.  It receives a set of markers.  So, we need to `inject` some field into a marker to show if they are selected.  We then need to check the set of markers in the above method, and decide if any are currently selected.  If they are - then we add styling to the marker cluster to make it pulse.

To test this we need data at exactly the same location. I'll investigate this.

lebaphi:
Fixes #2508 

https://user-images.githubusercontent.com/26705621/227768682-07b80946-d456-4f50-95dd-1c4570240cb6.mov

### DIFF
--- a/packages/components/src/local/organisms/planning-force/index.tsx
+++ b/packages/components/src/local/organisms/planning-force/index.tsx
@@ -50,12 +50,13 @@ const PlanningForces: React.FC<PropTypes> = ({
 
   const createClusterIcon = (): MarkerClusterGroupOptions => {
     return {
-      iconCreateFunction: function (cluster: MarkerCluster): L.DivIcon {
+      iconCreateFunction: (cluster: MarkerCluster): L.DivIcon => {
         const markers = cluster.getAllChildMarkers()
+        const selectedCount = markers.filter(m => m.options.attribution === 'true')
         const size = markers.length / 5 + 60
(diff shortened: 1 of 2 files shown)


packages/components/src/local/organisms/planning-force/index.tsx
```diff
@@ -99,9 +100,8 @@ const PlanningForces: React.FC<PropTypes> = ({
         if (asset.position) {
           if (!clusterIcons || selectedAssets.includes(asset.id) || (currentAssets && currentAssets.includes(asset.id))) {
             raw.push(asset)
-          } else {
-            clustered.push(asset)
           }
+          clustered.push(asset)
         }
```
IanMayo, line 105: @lebaphi - doesn't the above code mean that assets could get put into the `raw` _and_ the `clustered` lists?  That would mean they were rendered twice.

IanMayo (approved):

PR #2667, opened 2023-04-07
2663: Let players filter orders by asset involved

### LINKED ISSUE #2663: Let players filter orders by asset involved
Players are trying to review what happened in the game.

They want to use Serge to do that. 

The first thing they want to do is find all of the orders that relate to an asset they are interested in.

image

So, we will add an (initially hidden) `Assets` column to the orders table.  In the collated data, this cell will contain a comma-separated of the "own forces" and "opp forces" involved in that set of orders.  Note: in version 1 of this it will be ok to show the asset-id.  In the longer term we'll show the asset name.  The will take a free text search.  So the player will be able to search for orders relating to a specific asset.

lebaphi:
Fixes #2663


IanMayo (approved):
Tested and working

Issue #2697, opened 2023-11-21
TSX API for Core Mapping


IanMayo:
Come up with TSX API for the mapping component.

Consider:
- things it has to know about: 
    - current force/role
    - current turn
    - current phase
    - FeatureCollection 
    - channel definition
    - ForceStyle data (names/colors)
- things it must provide:
    - list of asset ids/names

IanMayo:
Done,for now.

PR #2701, opened 2023-11-23
2697 initial renderers

IanMayo:
Supports #2697 

image


IanMayo:
Could you help me out with these improvements please @lebaphi: 
* [x] we're missing the icon for `marker-shadow.png` (see console)
* [ ] I don't want to re-create the array of renderers on each new message/features update. Do you think we can supply the `features` object down in the TSX? 
image
* [x] our two renderers are going to override some GeoJSON behaviours.  Currently, they just have distinct `filter` behaviours.  But, I don't want that logic in this class.  I'd like it in two child classes, in a `renderers` sub-folder. Could you refactor it please?  I _guess_ each class could **extend** `GeoJSON`.

lebaphi:
Hi @IanMayo , I pushed a small changes of creating some renderer classes, can you review it :D
> I don't want to re-create the array of renderers on each new message/features update. Do you think we can supply the features object down in the TSX?

yep, I think so.

IanMayo:
Thanks for the refactoring. It's almost what I was looking for.  But, I'd like to avoid re-creating the renderers when we get a new message (features).  Really, we only need to generate the renderers on a `channel` update.

Is there a way where we pass the `features` object down in the TSX?  Then we're not re-creating the renderer, we're just updating a prop.

client/src/Components/local/atoms/core-mapping/helper/core-renderer-helper.tsx
```diff
@@ -0,0 +1,18 @@
+import { FeatureCollection, GeoJsonProperties, Geometry } from 'geojson'
+import { RENDERER_CORE, RENDERER_MILSYM } from 'src/custom-types'
+import { CoreRenderer } from '../renderers/core-renderer'
+import { MilSymbolRenderer } from '../renderers/milsymbol-renderer'
+import { ICoreRenderer } from '../renderers/I-core-renderer'
+
+export class CoreRendererHelper {  
+  static from (type: string, features: FeatureCollection<Geometry, GeoJsonProperties>): ICoreRenderer {
+    switch (type) {
+      case RENDERER_CORE:
```
IanMayo, line 10: Hello @lebaphi - in the next line, why don't we just pass `features` with the constructor.  We wouldn't need `from` then.  Or is this a programming practice that's new to me?

>  return new CoreRenderer().from(features)

(though it doesn't matter if we pass features in the TSX at the end of `core-mapping`.

lebaphi:
> Thanks for the refactoring. It's almost what I was looking for. But, I'd like to avoid re-creating the renderers when we get a new message (features). Really, we only need to generate the renderers on a `channel` update.
> 
> Is there a way where we pass the `features` object down in the TSX? Then we're not re-creating the renderer, we're just updating a prop.

I've updated code a bit, hope this is version that you were looking for @IanMayo

PR #2708, opened 2023-12-01
Add new drawn entities to our FeatureCollection

### LINKED ISSUE #2694: MilSymbol
Add the ability to drop new [MilSymbol](https://github.com/spatialillusions/milsymbol) items onto the Leaflet map.

These do not need to be configurable (yet). Any symbol will do.

Ok @lebaphi - I think we're ready for this one.

When geoman wants to `add a point` we need to:
* intercept the call
* add the extra metadata to look like a MilSymbol feature (like in the other mock data)
* add the item to the `features` array
* re-render `features`

IanMayo:
Fixes #2694


IanMayo:
Nice work @lebaphi 

I think the next thing to sort out is to update `features` when something is moved.  I suspect only the `leaflet` object gets updated.

client/src/Components/local/atoms/core-mapping/helper/map-controls.tsx
```diff
@@ -30,9 +85,7 @@ const MapControls: React.FC<GeomanControlProps> = ({ onCreate, onChange }) => {
       snappingOption: true
     }}
     globalOptions={{}}
-    onCreate={e => onCreate(e as unknown as PM.ChangeEventHandler)}
-    onChange={e => onChange(e as unknown as PM.ChangeEventHandler)}
-
+    onCreate={e => map.removeLayer(e.layer)}
```
IanMayo, line 97: Hello @lebaphi - I guess the above line is removing the new item from the GeoMan data since it is going to "arrive" via an updated `features` prop.  But, I _think_ we have to _wipe_ the GeoMan data on each `features` update - so it should not be necessary to `remove` the layer above.

client/src/Components/local/atoms/core-mapping/index.tsx
```diff
@@ -36,22 +38,120 @@ const CoreMapping: React.FC<PropTypes> = ({ messages, channel, bounds }) => {
...
+        setFeatures(cloneDeep(features))
+        console.log('features: ', features.features)
+      }
+    }  
   }
 
-  const onChange = (e: PM.ChangeEventHandler) => {
-    console.log('onChange Event Fired', e)
+  const onChange = (id: number, latlng: LatLng[]) => {
+    console.log('onChange Event Fired', id, latlng)
```
IanMayo: @lebaphi - do you know when `change` is fired?  I don't think I can trigger it.

client/src/Components/local/atoms/core-mapping/helper/map-controls.tsx
lebaphi: okay, I will remove this line

client/src/Components/local/atoms/core-mapping/index.tsx
lebaphi: Actually, I'm trying to listen the onChange callback when dragging marker/features but got no luck 😄

lebaphi:
@IanMayo I pushed the changes to update the onChange listener, there is a small problem that when you drag a marker to change its coordinates then create a new marker, the old marker moves back to the old position

IanMayo:
> @IanMayo I pushed the changes to update the onChange listener, there is a small problem that when you drag a marker to change its coordinates then create a new marker, the old marker moves back to the old position

Sure. That looks like a sign that we're not updating the `Feature` on GeoMan drag events.  Leaflet is maintaining it's own `private` copy of the data, that gets loaded into a GeoJSON layer.  We need to intercept these events, change our `features`, then re-generate the  GeoJSON layer.

I'll have a play...

IanMayo:
Merging  - so tidy  development can continue.   There remains an outstanding issue related to deleting markers.

PR #2710, opened 2023-12-12
2704: implement basic UI for control panel

### LINKED ISSUE #2704: Control panel for Core Mapping
We require some general controls for the Core Mapping channel.

Here is a mockup:
https://docs.google.com/presentation/d/1LdTKv9BTrJdnfshJeEz6SpmMML8D_kbNc8emZ-Ugw6Q/edit#slide=id.g29d5206fb52_0_40

(more detail on following slide)

The aim here is for them not to be focussed on a particular wargame, style of data, or set of properties [*].  So, the UI will largely be build using retrospection from the current features on the map

* [x] create left-hand control panel. It's width (as a %age, I guess) should be persistent
* [x] create top `Filters` panel.  It should have a `filters` heading.  It should start with zero filters, but should have an `add` button to add a new filter.   The `add` button will popup a list of known properties (I don't yet know where these will come from. Either by doing a "live" scan of current data, or on each new update to `features`
* [x] create central `Items` panel. This will be a list of the currently visible items.  For `MilSymbol` it will show the symbol, for `CoreRenderer` item it will be a generic `shape` icon, shaded using the `color` for that item
* [x] create lower `properties` panel.   If one item is selected in the `items` listing, show it's properties here.  We're display text-boxes or drop-downs depending on what we can determine from the property.  We won't show properties with a leading underscore.

[*] - ok, we'll allow a _few_ predictable properties. `sidc` for MilSymbols and `color` for all features.

lebaphi:
Fixes #2704


IanMayo:
~~Grr, I commented earlier, but I cant see them now.~~

~~1. The color shades in the mockup were just to help discussion. I'm happy for them to take the standard UI template shade (presumably grey)~~
~~2. The `+` icons for the drag handles seem non-standard. I _think_ the standard icon is either a group of dots, or a set of parallel lines.~~

~~Other than that - I'm quite happy with the progress, and look forward to contributing to the logic when I get chance :-D~~

Aah, the comment is here :-D 
https://github.com/serge-web/serge/issues/2704#issuecomment-1851762818

IanMayo:
Thanks for this so-far @philb89 

Can I suggest a next step?  How about filling in the items in the `list` component from the current set of features?  I'm happy to just show the `label` value in the list, to start with.

Then we can start to work on auto-generating the filters, and applying them to the list.

lebaphi:
> Thanks for this so-far @philb89
> 
> Can I suggest a next step? How about filling in the items in the `list` component from the current set of features? I'm happy to just show the `label` value in the list, to start with.
> 
> Then we can start to work on auto-generating the filters, and applying them to the list.

it sounds great @IanMayo , I'm going to work on this

IanMayo:
Hey, nice work @lebaphi 

These are the next steps:
- the "properties" panel needs to be responsive to the selected item.  So, we need to look at the properties for the item and show one row per property.  Aah, we also need to look at the properties defined for the channel, to see if it is a text field or an enum, to know whether to show a text field or a drop-down.
- similarly, when the user chooses to add a new field, we need to check for a `property` with that id in the channel, and either show a text box or a drop-down list, that we use to filter the items.

IanMayo:
Some linter issues and a CSS-conflict to resolve, please @lebaphi :
image

lebaphi:
> Some linter issues and a CSS-conflict to resolve, please @lebaphi : image

Yep, I'm working on it @IanMayo

IanMayo:
There is an issue present where the `id` of `milSymbols` are added to the `attribution` at the bottom of the map. It's present in the normal Storybook, but it's more obvious in the bulk example:
image

IanMayo:
Aah and another thing please @lebaphi - for the `filters` in the top section, could you inspect the  `baseProps` and `additionalProps` for each filter and see if it's possible to provide a drop-down list in the filter?

lebaphi:
Hi @IanMayo, I just pushed the fixes 😄

IanMayo:
> Hi @IanMayo, I just pushed the fixes 😄

Nice work, thanks @lebaphi

IanMayo (approved):
Initial implementation working well

PR #2717, opened 2023-12-20
Playwright test phase1

Tristina1788:
Add the first phase for automation test with playwright 
- Add log console "Hello world"
- Verify login passed and failed
- Update readme.md with automation part


e2e/playwright_test/playwright.config.ts
```diff
@@ -0,0 +1,90 @@
+import { defineConfig, devices } from '@playwright/test';
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
```
IanMayo: Hello @Tristina1788 - this `require` code is commented out.  Should we just delete it, or might we need it in the future?

IanMayo:
The `DeepScan` linter found this issue:
image

It's probably worth investigating how we should handle retries when running on CI.

Tristina1788:
> It's probably worth investigating how we should handle retries when running on CI.

We can use this config for retry failed test. Sometimes the test maybe failed because of connection or element is slower than expected. It will helps us run again. Now I set it 0 to not take time for retry.

IanMayo (approved):
tested and working

PR #2738, opened 2023-12-31
Should be able to Restrict access in new participation row

### LINKED ISSUE #2722: Should be able to Restrict access in new `participation` row
## 🐞 Bug description
User can't select restrict access with default force. User have to select force again to enable to select restrict access.

## 🔢 Steps to Reproduce

1. Login app as admin
2. Create a new game
3. From 'Forces' tab, create some forces ( White, Blue, Red)
4. From 'Channels' tab, click 'Add Channel'
5. Notice it has a new row with default force
6. I select restrict access for that force but It doesn't show any options.
7. I click on the force and select that default force
8. Notice I can select restrict access

More details from below screencast

From @IanMayo - it looks like the `roles` drop-down doesn't get populated until the force drop-down gets modified.  We should `trigger` a forced update on load, so that it initialises the roles dropdown.

## ⚙️ Expected behavior
At step 6, it should show all restrict access of the default force

**UPDATE FROM @ianmayo** - better still, let's remove the default row that gets added when we add a new channel. Separate to that - we still need to check that the restrict access values are present for any new participation.

## 🖥️ Screencasts

https://github.com/serge-web/serge/assets/107697044/61462fea-0f92-4f5c-b9f4-254be5292008

## 💻 System information

 - Browser : Chrome

lebaphi:
Fixes #2722


IanMayo (approved):
tested and working

PR #2768, opened 2024-01-16
WIP: 2745 Core Mapping - support for generating message deltas

### LINKED ISSUE #2745: Core Mapping - support for generating message deltas
To reduce the size of the MappingMessage objects, we should start working with deltas.

We'll support this by `MappingMessage` having either a `featureCollection` or a `featureDelta`.

When the shapes on the map are changed (C, U or D), we should calculate the delta of the feature collection, and send this as a new message for that channel.

So, the wargame database will contain an initial `featureCollection` and then a series of `featureDelta` messages.

Note: let's use this library to generate the diffs: https://github.com/cujojs/jiff

**Update:** -  unless you prefer a different implementation @lebaphi   We could use one that includes TypeScript definitions, if you prefer that.

lebaphi:
Fixes #2745


client/src/Components/local/atoms/core-mapping/index.tsx
```diff
@@ -107,11 +109,59 @@ const CoreMapping: React.FC<PropTypes> = ({ messages, channel, playerForce, play
...
+          },
+          since: newMessage._id,
+          delta
+        }
+        postBack(deltaMessage)
       }
-      postBack(newMessage as MappingMessage)
     }
   }
+
```
IanMayo, line 131: @lebaphi - could we please generate the diff using an JSON-Patch library?  Then we can be confident of how the diff is written/implemented. This also means apps other than Serge may be able to handle/generate them.
The possible libraries are here:
https://jsonpatch.com/#javascript

lebaphi: okay @IanMayo

lebaphi: I just added the jiff lib to generate the diff but it seems the lib throws an error when applying the patch, I'm still investigating it. @IanMayo

lebaphi: I decided to use this lib https://www.npmjs.com/package/fast-json-patch, it seems better in terms of performance and it solves the error of jiff lib. The only small problem I have is that after the c/u/d feature and reloading the page, all the features are gone. I'm still working on it but I welcome you to check this out. @IanMayo

client/src/Helpers/handle-channel-updates.ts
```diff
@@ -111,10 +113,25 @@ export const handleNewMessageData = (
       channel.messages = []
     }
     channel.messages.unshift(payload)
-  } else {
+  } else if (payload.messageType === CUSTOM_MESSAGE) {
     handleNonInfoMessage(res, payload.details.channel, payload, playerId)
+  } else {
+    const messageType = get(payload, 'messageType')
```
IanMayo: @lebaphi  - can we please keep knowledge of deltas, and how to create/merge them inside `core-mapping`?

IanMayo:
One first observation.  The `delta` should be of the `featureCollection`, not the whole message:
image

client/src/Helpers/handle-channel-updates.ts
IanMayo: To do the above, I _think_ we just call `handleNonInfoMessage` (as before), and then handle the new `messages` in `CoreMapping`

client/src/Components/local/atoms/core-mapping/index.tsx
IanMayo: I think I've got it. Our current `delta` handling is in `handleNewMessage`.  That method doesn't get called when we load an existing wargame.
So, I think the `delta` handling has to happen in `CoreMapping`.  Then it will run on initial load aswell as on new messages.

client/src/Helpers/handle-channel-updates.ts
lebaphi: I see, I will try this 👍

client/src/Components/local/atoms/core-mapping/index.tsx
```diff
@@ -103,16 +112,23 @@ const CoreMapping: React.FC<PropTypes> = ({ messages, channel, playerForce, play
         timestamp: timestamp,
         turnNumber: 1
       }
-      const newMessage: MappingMessage = {
-        details, 
-        featureCollection,
-        messageType: CUSTOM_MESSAGE,
-        _id: timestamp
+
```
IanMayo, line 115: Aah, @lebaphi - there is another bit of logic we can use here.  If there **isn't** a `lastMessage` then we send the whole `newFeatureCollection` as a `MAPPING_MESSAGE`.

At some point in the future we will need to consider when we send these whole messages.  But, when do come up with something, we will just have to clear `lastMessage` - and know that the full `featureCollection` will be sent.

PR #2776, opened 2024-01-21
if there is no last message, save the whole mapping message

lebaphi:
Fixes #2768


IanMayo (approved):
Verified by code inspection.

PR #2777, opened 2024-01-22
Support text editing in Core Mapping channel

### LINKED ISSUE #2761: Support text editing in Core Mapping channel
The GeoMan demo page allows text labels to be added:
https://www.geoman.io/demo

Our core-mapping interface doesn't currently support this.

We need to add the ability to add and edit text labels, with these changes:
- the origin of the text label should be at the centre of the label, in the demo it appears to be at the top-left
- the padding around the text doesn't seem to be even on all sides, it should be
- we should be able to control the color of the backdrop (via feature properties)
- _maybe_ we should be able to control the font size  (via feature properties)

https://github.com/serge-web/serge/assets/1108513/d180fe5d-7084-405f-87cb-d46eeadbf197

lebaphi:
Fixes #2761 

There is some minor issues but it will work like this I think :D

https://github.com/serge-web/serge/assets/26705621/e5f6b120-eaa3-4473-962c-5da5a048fbfb


IanMayo:
Yes - I see we're really close.  In the GeoMAN demo, doesn't it start off with being an editable component (so the user immediately types in the text)?  Can we reproduce that?

lebaphi:
> Yes - I see we're really close. In the GeoMAN demo, doesn't it start off with being an editable component (so the user immediately types in the text)? Can we reproduce that?

Yep, I'm working on this ✌️

lebaphi:
Hi @IanMayo, I just pushed the fixes, can you help to check this 😄

IanMayo:
Yes, nicely one thank @lebaphi 

There are a few more requirements recorded in the issue, please.

IanMayo:
Nice progress :-)

I don't think we need both `TextValue` and `Label`.  If possible, could we just have the `Label` field?
image

IanMayo (approved):
Tested and working

client/src/Components/local/atoms/core-mapping/index.tsx
```diff
@@ -202,6 +203,24 @@ const CoreMapping: React.FC<PropTypes> = ({ messages, channel, playerForce, play
           }
         }
       }
+      case 'Text': {
+        const loc = (e as any).layer._latlng as L.LatLng
+        return {
+          type: 'Feature',
+          properties: {
```
IanMayo, line 210: @hello @lebaphi the `text` element should be of type RENDERER_CORE, rather than RENDERER_MILSYM - since MILSYM is just for entities from the MilSymbol library, please.

IanMayo (changes requested):
Just the one change please - to move `text` from MILSYM to CORE.

IanMayo (approved):
Tested and working

PR #2778, opened 2024-01-23
2000_Clarify_strategy_for_messageTyp

### LINKED ISSUE #2000: Clarify strategy for `messageType`
In our message definitions, we use `messageType` to represent the type of message:
image

But, when we actually send a message, the `messageType` details gets dropped, and it gets sent as `CUSTOM_MESSAGE`:
image

Note: we do also have a `messageType` in the `message.details` - but this is being used as the placeholder for the template id to be used to edit the message.  The `MessageDetails` element is common to all message types, but the template only belongs in the `CUSTOM_MESSAGE` structure. So, we should both rename `details.messageType` to `details.templateId` AND we should move the field up into `MessageCustom`

Here is what I _think_ we have to do:
- simplify `MessageCustom`, since it contains some legacy design decisions:
    - remove `gameTurn` (since its in MessageDetails)
    - remove `feedback`  (since we have custom message type for that)
    - remove `infoType` (since we handle that in `messageType`, too)
    - we should do _something_ with `isOpen` - since that doesn't belong in the database, it's just for this client
- modify `postNewMessage` and `saveMessage` to take a `messageType` field. (Hey, do we need both methods? Can we just use one?)
- modify `postNewMessage` to call `checkReference` if there is a `message.message.reference` field - regardless of the `messageType` (it looks like it is currently doing an extensive `allDocs()` server round-trip before sending every new message)
- avoid two instances of `postNewMessage`.  Do we need both? If we do, one should be renamed to better explain its use.
- modify calls to `saveMessage` to introduce `messageType`
- modify mock/sample data to use correct message type (esp. PT-9 wargame in SQLite)

lilitkarapetyan:
Fixes #2000 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations

### DIFF
--- a/client/src/custom-types/message.d.ts
+++ b/client/src/custom-types/message.d.ts
@@ -43,8 +43,6 @@ export interface MessageDetails {
   //  * extra detail for managing an interaction
   //  */
   // interaction?: InteractionDetails
-  /** ID of template for this message */
-  messageType: TemplateBody['_id']
   /** time message sent */
   timestamp: string
   /** turn when this message was sent */
@@ -86,6 +84,7 @@ export interface CoreMessage {
   readonly details: MessageDetails
   /** whether this message has been read on the current client */
   hasBeenRead?: boolean
+  readonly templateId?: TemplateBody['_id']
 }
 
 /** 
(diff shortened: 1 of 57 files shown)


IanMayo (dismissed):
Tested and working

client/src/custom-types/message.d.ts
```diff
@@ -130,6 +130,8 @@ export interface CollaborationDetails {
 
 export interface MessageCustom extends CoreMessage {
   readonly messageType: typeof CUSTOM_MESSAGE
+    /** Details related to the message, excluding the message type. */
```
IanMayo, line 132: @lilitkarapetyan - I _think_ we can remove `messageType` from `MessageDetails`, now that we are storing the correct` messageType` in the parent `CoreMessage`.

I've experimented with the change, and see it throws up quite a lot of errors. But, I'm pretty sure it's the right thing to do.

IanMayo (approved):
Tested and working

lilitkarapetyan:
Fixes #2000

PR #2784, opened 2024-01-25
Core-Mapping: zoom controls (including wheel)

### LINKED ISSUE #2782: Core-Mapping: zoom controls (including wheel)
The zoom in/out controls are obscured by the left-hand panel.  Can we move the controls to the bottom-right of the screen?

The mouse-wheel doesn't work for scroll. It does in the `default` leaflet implementation:
https://leafletjs.com/

Can you get the mouse-wheel working please?

lebaphi:
Fixes #2782


IanMayo (approved):
Tested and working.

PR #2789, opened 2024-01-29
2779_Refactor_for_FeedbackMessage_move_from_to_core_message

### LINKED ISSUE #2779: Refactor - for FeedbackMessage, move `from` to core message
I just submitted a FeedbackForm, including a "From" value.

```
{
    "details": {
        "channel": "Feedback",
        "from": {
            "force": "Blue",
            "forceColor": "#3dd0ff",
            "forceId": "f-blue",
            "iconURL": "",
            "name": "Mr Ian",
            "roleId": "blue-mtg-1",
            "roleName": "MTG 1"
        },
        "messageType": "Chat",
        "timestamp": "2024-01-23T10:27:10.034Z",
        "turnNumber": 2
    },
    "message": {
        "content": "some game feedback"
    },
    "messageType": "FeedbackMessage"
}
```

Note the `name` field in the `details`. This field is declared in the `MessageDetailsFrom` typoe.  Since this is message of type `FeedbackMessage` we should move `name` to an optional field within the `MessageFeedback`  type definition.  We should update the rest of the code accordingly. Note: we may need to change the api of `postFeedback`.  Hmm, maybe we should build of a `FeedbackMessage` in the `InsightForm` - and just send the message like any other message - rather than via a custom `postFeedback` API.

lilitkarapetyan:
Fixes #2779 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


lilitkarapetyan:
fixes #2779

client/src/Views/MessagesListInsightsChannel.jsx
```diff
@@ -24,7 +24,7 @@ const MessagesListChatChannel = ({ messages, markAllAsRead }) => {
             <div className='info-wrap'>
               <Badge pill variant='primary'>{message.details.from.force}</Badge>
               <Badge pill variant='secondary'>{message.details.from.role}</Badge>
-              {message.details.from.name && <Badge pill variant='warning'>{message.details.from.name}</Badge>}
+              {message.name && <Badge pill variant='warning'>{message.name}</Badge>}
               <span>{formatFullDate(message.details.timestamp)}</span>
```
IanMayo, line 31: Hello @lilitkarapetyan - I think the above bit isn't quite right.  The `name` is optional when submitting feedback, but I think the Game Admin will be interested in both fields. 

Aah, it also highlights another issue.  There isn't a `role` in `messages.details.from`.  This problem is falling through the cracks because this file is `jsx` rather than `tsx`.  Can you convert it to `tsx` please?

lilitkarapetyan: Yes, I will work on it now.

client/src/Views/MessagesListInsightsChannel.tsx
```diff
@@ -1,18 +1,21 @@
...
+interface Props { messages: MessageFeedback[], markAllAsRead: () => void }
+
+const MessagesListChatChannel: React.FC<Props> = ({ messages, markAllAsRead }) => {
   const { currentWargame, selectedForce, selectedRole } = usePlayerUiState()
   const selectedForceId = selectedForce ? selectedForce.uniqid : ''
-  const isMessageRead = (message) => isMessageReaded(currentWargame, selectedForceId, selectedRole, message._id)
+  const isMessageRead = (message: MessageFeedback) => isMessageReaded(currentWargame, selectedForceId, selectedRole, message._id)
   const isMessageUnreadExists = messages.some(message => !isMessageRead(message))
-
+
```
IanMayo, line 18: Hello @lilitkarapetyan - I want to drop the human-readable `force` and `role` from the `messageDetailsFrom` block.  This is because they can be determined from the `forceId` and `roleId` fields.   It will also mean that if a role-name changes during a wargame, the `from` for old messages will use the new name, rather than the one at the time when the message was sent.

So, this class will have to collate a list of wargames and roles (a dictionary, I guess) and lookup the names of the force and role in order to display them.

lilitkarapetyan: I wanted to seek clarification regarding the proposed changes to the Message Details From interface. Below are the modifications that have been suggested: 
```
export interface MessageDetailsFrom {
  /** id of sending force */
  readonly forceId?: ForceData['uniqid']
  /** CSS color shade for this force */
  readonly forceColor: ForceData['color']
  /** role of the individual that wrote the message */
  readonly roleId: Role['roleId']
  /** URL of icon to display for this force */
  iconURL: ForceData['iconURL']
  /** user-name, as typed into Feedback/insights form */
  name?: string
}
```  
Before proceeding with these changes, could you please confirm if my understanding aligns with the intended task?

IanMayo: Thanks for double-checking @lilitkarapetyan 

1. I don't think we need `name` since it has moved to `FeedbackMessage`
2. let's drop `forceColor` and `iconURL` - since both can be determined from the wargame data

Oh dear, now we're down to just `forceId` and `roleId`.  We already need `roleId` to be unique across the whole wargame.  So, we could get away with just `roleId`.  If that's the case, we can remove the `MessageDetailsFrom` structure, and just have `from: Role['roleId']` in `MessageDetails`.

That will be quite a big refactoring, but you've done larger recently. 

Does it seem ok with you?

lilitkarapetyan: I'm okay with the suggested changes, and I think they will make things work better and be easier to manage.

IanMayo: Have captured this issue here: https://github.com/serge-web/serge/issues/2813

IanMayo (approved):
Tested and working

PR #2801, opened 2024-01-31
2797 - Core Mapping - lat/long editor

### LINKED ISSUE #2797: Core Mapping - lat/long editor
We need to allow users to enter text values for new locations of point markers.

I suggest that for any feature that is a `Point`, we introduce a new Lat/Long editor that allows editing of the `coordinates` array. This will be two adjacent text controls.

Note: The `human` convention is for Lat/Long. The `GeoJSON` convention is for `Long/Lat` so I'm afraid we'll just have to keep swapping them.

### LINKED ISSUE #2807: Core Mapping - custom editor control for lat/long
We have a PR in progress that introduces lat/long as two new properties, shown on separate lines in the property editor.

But, these two properties are so closely associated with each other that I'd like them to be shown as two text fields side-by-side, with `tab` key moving from the first to the second.

lebaphi:
Fixes #2797 
Fixes #2807


IanMayo:
I'm impressed with the simplicity of this solution @lebaphi 

But...
- I don't like the "duplication" we have where lat/long are stored in properties and in the coordinates.   So, I think we need our code to generate mock lat/long properties when an item is about to be shown in the properties editor, and save-then-remove them as the item is being saved

(note: I'd quite like the two text boxes to be on the same row in the UI, but I'm happy to defer that since we have larger issues to resolve. I'll create an issue for it)

client/src/Components/local/atoms/core-mapping/index.tsx
```diff
@@ -195,8 +195,8 @@ const CoreMapping: React.FC<PropTypes> = ({ messages, channel, playerForce, play
             _type: RENDERER_MILSYM,
             sidc: 'SFG-UCI----D',
             size: 'M',
-            lat: loc.lat,
-            lng: loc.lng,
+            lat: mockLat, // mock lat
```
IanMayo: @lebaphi - I think it's _unhealthy_ to store the values as properties - it just feels like it will cause trouble downstream.  As a workaround, could we inject them as special properties before we call `setSelectedProps` in `mapping-panel.tsx`?  We would then extract them again (if present) in `onPropertiesChange`.

lebaphi: I don't quite understand your idea, to show the Lat/Lng editor in the properties panel, it has to be defined in `properties`, and it should show the coordinates of the marker?

image

IanMayo: Sure.  I don't like to see the data stored in 2 places (coordinates and properties).  I can see that's the only way to expose it in the properties panel.  So, instead of it being permanently stored in properties, I'd like us to only have it in there when the properties are being used to build the properties window.  So, immediately before we call `setSelectedProps()` we would check if this is a `point` feature, and insert lat/long properties into the array.  Similarly, on `onPropertiesChange` we would update the `coordinates`, remove the lat/long properties, and pass on the reduced set of properties.

That's all quite complex/convoluted.  I _guess_ an alternate strategy could be for the `PropertiesPanel` to have an optional LatLng value.  If this is provided, introduce a `LatLng` as a first property in the UI.  The `onPropertiesChange` would gain an optional latLng field, populated if there is a LatLng present.

IanMayo (approved):
Tested and working.

PR #2812, opened 2024-02-01
Core Mapping - Double-click to select

### LINKED ISSUE #2810: Core Mapping - Double-click to select
Users can use filter(s) to get to features of interest, then select them and edit their properties.

It would be useful to be able to click on a feature on the map, and it be selected in the mapping bar.

lebaphi:
Fixes #2810 

Hi @IanMayo, currently I think we can support click event listener because the marker does not support double click event.

### DIFF
--- a/client/src/Components/local/atoms/core-mapping/index.tsx
+++ b/client/src/Components/local/atoms/core-mapping/index.tsx
@@ -320,10 +320,6 @@ const CoreMapping: React.FC<PropTypes> = ({ messages, channel, playerForce, play
     }
   }
 
-  const onSelect = (id: number | string) => {
-    setSelectedFeature(id)
-  }
-
   const getExtraFilterProps = (): PropertyTypes[] => {
     const rendererObjects: BaseRenderer[] = channel.renderers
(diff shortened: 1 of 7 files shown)


client/src/Components/local/atoms/core-mapping/index.tsx
```diff
@@ -320,10 +320,6 @@ const CoreMapping: React.FC<PropTypes> = ({ messages, channel, playerForce, play
     }
   }
```
IanMayo, line 322: are you ok with this simplification @lebaphi ?

IanMayo:
> Hi @IanMayo, currently I think we can support click event listener because the marker does not support double click event.

Sure - single-click is fine.

client/src/Components/local/atoms/core-mapping/index.tsx
lebaphi: > are you ok with this simplification @lebaphi ?

sure, it's completely fine @IanMayo

IanMayo (approved):

PR #2815, opened 2024-02-01
2808 Don't hard-code force colors

### LINKED ISSUE #2808: Don't hard-code force colors
In `core-renderer.tsx` we have hard-coded force ids and colors.

Instead, we should lookup the data from the `ForceStyle` object provided in the props.

lebaphi:
Fixes #2808


IanMayo:
Sorry @lebaphi - I didn't really solve your problem there.

image

I didn't realise they weren't being passed into the parent class.  I'll try to hook things up for your in the morning :-D

IanMayo (approved):
Tested and working

PR #2820, opened 2024-02-04
Core Mapping - flicker on messages in other channel

### LINKED ISSUE #2791: Core Mapping - flicker on messages in other channel
I have the `core mapping` channel open.  If I send a `Game Admin` messages in the right-hand panel, the map flickers.

We should only re-render the core mapping channel when:
1.  a new message has been received for that channel
2. the mapping features actually have changed

lebaphi:
Fixes #2791


IanMayo:
Hello @lebaphi - I understand the approach, but it would be great if it could be handled higher up the component tree?

If the most recently received messages aren't for this channel, then we don't need to re-render this component.

(I'm hoping this will help avoid future issues, such as losing edits in "live" edit forms).

IanMayo (approved):

PR #2843, opened 2024-02-11
Should have the way to know which item is selected from left panel with item on map

### LINKED ISSUE #2835: Should have the way to know which item is selected from left panel with item on map
## 🐞 Bug description
When user select an items on left panel, notice no item is high light on map. We should have the way to make user know the selected item is which item on map. So that user can update properties of correct item.

From @IanMayo: Yes, good observation @Tristina1788 .  @lebaphi - could we introduce some glow/shadow/other on the map to highlight selected item?  We'll obviously have to pass the current selection to the map.  I know we can only currently select one item at a time, but I _guess_ it we should start off by passing an array of selected items, to allow for future growth.  This will start to remind you of the previous mapping :-D

## 🔢 Steps to Reproduce
2. Click on channel 'Core Mapping'
3. Select a item on left panel
image

5. Notice user don't know the selected item is which item on map

## ⚙️ Expected behavior
When user select an items on left panel, the item should be highlight on map

## 🖥️ Screencast

https://github.com/serge-web/serge/assets/107697044/3107484c-c821-4822-9806-9e5bdbee6dd8

## 💻 System information
- Browser : Chrome

lebaphi:
Fixes #2835

### DIFF
--- a/client/src/Components/local/atoms/core-mapping/helper/marker-helper.ts
+++ b/client/src/Components/local/atoms/core-mapping/helper/marker-helper.ts
@@ -17,7 +17,7 @@ export const loadDefaultMarker = (): void => {
 export const createDivIcon = (iconHTML: string, healthColor: string, isSelected: boolean): L.DivIcon => {
   return L.divIcon({
     html: `
-      <div class="${cx({ [styles['asset-icon']]: true, [styles['selected-icon']]: isSelected })}">
+      <div class="${cx({ [styles['asset-icon']]: true, [styles['pulse-milsymbol']]: isSelected })}">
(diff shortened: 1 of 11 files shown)


lebaphi:
Hi @IanMayo, the box-shadow doesn't work for shapes so I added scale animation for them.

https://github.com/serge-web/serge/assets/26705621/bb0c40a5-ddd8-4a01-802b-5a55027c38f9

IanMayo:
Thanks @lebaphi - can you think of a highlighting strategy that is consistent for both types of renderer?

IanMayo (approved):
Tested and working

PR #2853, opened 2024-02-14
'Important' filter gets wrong result with default 'Yes'

### LINKED ISSUE #2847: 'Important' filter gets wrong result with default 'Yes'
## 🐞 Bug description
'Important' filter has default value is 'Yes' and the result of filter is all items. It should only show the result which match with default 

**Update from Ian:** the `important` boolean filter should have a third `Empty` state, which does not try to match anything.  When added, the filter component will be in this state.  The user then selects a boolean state, and the filter is applied.

## 🔢 Steps to Reproduce
2. Click on channel 'Core Mapping'
3. Add filter with 'Important'
4. Notice the default filter 'Important' is yes
5. Notice 'Items' panel shows all items instead of it shows the items which are matched with 'Important' = Yes

image

View below screencast for more details
## ⚙️ Expected behavior
'Important' filter has default value is 'Yes' and the result of filter is all items. It should only show the result which match with default 

## 🖥️ Screencasts

https://github.com/serge-web/serge/assets/107697044/7258d86a-7aba-4865-9438-ca9277c3e139

## 💻 System information

 - Browser : Chrome

lebaphi:
Fixes #2847


IanMayo (approved):
Tested and working

PR #2862, opened 2024-02-18
Cannot load CoreMapping in Story Book

### LINKED ISSUE #2856: Cannot load CoreMapping in Story Book
## 🐞 Bug description
Get error when load CoreMapping in Story Book

**Update from Ian:** I _expect_ this is because there are no delta messages in the storybook. Here is the trace:
image

When I added the type for deltaMessages (see highlight below) VS-Code showed some compiler issues that would probably solve the above issue:
image

## 🔢 Steps to Reproduce
1. Go to 'Story Book' page
2. Click 'CoreMapping' > select 'Default'
3. Notice screen show the error page instead of the core mapping

image

## ⚙️ Expected behavior
It should show the 'CoreMapping' page instead without error

## 💻 System information
- Browser : Chrome

lebaphi:
Fixes #2856


Tristina1788:
Checking on '2856-core-mapping-storybook' branch, now the CoreMapping on Story Book can show as expected

https://github.com/serge-web/serge/assets/107697044/0a6163c0-e6b4-4ecc-8629-ffad68b804b0

IanMayo (approved):

IanMayo (changes requested):
Hello @lebaphi - this fixes the issue, but the underlying problem remains.  I _think_ the code is falling over when there are zero delta messages.  This PR resolves the problem by providing a delta message - but I'm pretty sure we need to fix it so that it doesn't fall over with zero delta messages.

IanMayo:
@lebaphi - wouldn't it be neater to only go down the `isAppliedPatch` and `applyPatch` route if the list of delta-messages is non-empty?

image

lebaphi:
> @lebaphi - wouldn't it be neater to only go down the `isAppliedPatch` and `applyPatch` route if the list of delta-messages is non-empty?
> 
> image

it's fine @IanMayo, let me update

IanMayo (approved):

PR #2890, opened 2024-02-26
Buttons to collapse map bar panels

### LINKED ISSUE #2875: Buttons to collapse map bar panels
The mapping bar has 3 panels, that can be dragged to resize.

I user was on a monitor with a fairly small resolution, and was frequently concentrating on just one or two panels.  It would have been easier to maintain the UI if the 3 panels had `collapse` buttons that kept the panel title bar, but hid the rest of that panel.

I _guess_ the rotating arrow marker is the standard expand/collapse icon.

lebaphi:
Fixes #2875

### DIFF
--- a/client/src/Components/local/atoms/mapping-panel/helpers/resize-handler.tsx
+++ b/client/src/Components/local/atoms/mapping-panel/helpers/resize-handler.tsx
@@ -6,10 +6,11 @@ import cx from 'classnames'
 type ResizeHandleProps = {
   direction?: 'vertical' | 'horizontal'
   className?: string
+  disabled?: boolean
 }
 
-const ResizeHandle: React.FC<ResizeHandleProps> = ({ direction = 'vertical', className }): React.ReactElement => {
-  return <PanelResizeHandle className={styles['resize-handle-outer']}>
+const ResizeHandle: React.FC<ResizeHandleProps> = ({ direction = 'vertical', className, disabled }): React.ReactElement => {
+  return <PanelResizeHandle className={cx({ [styles['resize-handle-outer']]: true, [styles.disabled]: disabled })}>
     <div className={cx(styles['resize-handle-inner'], className)}>
       <svg className={cx({ [styles['icon']]: true, [styles['resize-handle-horizontal']]: direction !== 'horizontal' })} viewBox="0 0 24 24">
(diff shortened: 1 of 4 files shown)


IanMayo:
I like the UI/UX, but there seem to be some collapse issues @lebaphi 

It looks like collapsed panels are shown as open when another one is collapsed.

screencast

lebaphi:
> I like the UI/UX, but there seem to be some collapse issues @lebaphi
> 
> It looks like collapsed panels are shown as open when another one is collapsed.
> 
> screencast [ ![screencast]

I see, it seems the method collapse (https://github.com/bvaughn/react-resizable-panels/tree/main/packages/react-resizable-panels#panel) does not work as expected, I will take a look

IanMayo:
Still not quite right @lebaphi 

https://i.gyazo.com/b033cfcac0a4c9719987a1036f0e448e.gif

Tristina1788:
@IanMayo , @lebaphi  Tested on '2875-collapse-button' branch, now collapse button show correct


https://github.com/serge-web/serge/assets/107697044/0fdded2f-cb41-4744-824a-2e98d3cb57c4

But I think it's better if collapse 'Propertive' should be on original position (top of this panel) instead of the bottom of panel.

image

Expected : 

image

IanMayo:
Yes @Tristina1788 - I agree.  When all 3 panels are collapsed, I think they should be in a stack at the top of the column.  The `properties` panel currently collapses down, but I think it should collapse upwards.

(I hope that makes sense @lebaphi )

lebaphi:
thanks @IanMayo @Tristina1788 , actually, this function is not supported as default, we can only implement it to a certain level, it is complex to implement it perfectly. So what is your expectation? 😄

IanMayo:
Actually @lebaphi - I realise that this is part of my issue.  I thought the "props" panel wasn't opening and closing.  It was working, but since nothing was selected, when open it didn't show anything.

If nothing is selected, could we still show (disabled) `Cancel` and `Save` buttons in the `Props` panel?

I'm happy to accept the Props panel collapsing down - that's just the bar giving as much v-space as possible to the `Items` panel.

Tristina1788:
Tested on '2875-collapse-button' , now when no item is selected, we can see  Cancel and Save buttons (disable)  in the Props panel .

image

It works as expected. (More details in this below video) 

https://github.com/serge-web/serge/assets/107697044/474a0fe1-988a-46e8-9963-32548bfbe07b

IanMayo (approved):
Working perfectly :-D

PR #2908, opened 2024-04-04
2867 Core Mapping - filter by shape type

### LINKED ISSUE #2867: Core Mapping - filter by shape type
It would be useful to be able to filter shapes by the feature type:
- point (symbol)
- linestring
- line / arrow
- polygon
- circle?

This may allow an umpire to delete a bunch of `route` data at the end of adjudication.

The filter entries will need to be updated on `features` change - so it contains a unique list of the shape types on the map.

lebaphi:
Fixes #2867


Tristina1788:
@lebaphi , @IanMayo Shapetype filter should not show as text input when I don't create more shape item. And after creating more Shape items, 2 filter ShapeType are on filter lists, it should shows only one.


https://github.com/serge-web/serge/assets/107697044/6c2fbe6d-65a2-4485-83b4-2467c5573228

I think it's better for filter ShapeType is dropdown list with all shape type from beginning as below

image

lebaphi:
Hi @Tristina1788 , @IanMayo 
about shapeType_should_alway_shows_as_dropdown_list.mp4 issue, I can explain a bit of logic here. When creating a new feature, we assign a **shapeType** in the feature properties and save it to the db (line 219)
image
So when adding a ShapeType filter, we load the dropdown options by the **shapeType** feature property (line 411)
image

In your case, the original features don't have a **shapeType** in the property, so that's why it's showing Input (choices.length === 0)
image

So I think Ian can update the shapeType property for the original features in the db, then will it be fixed.

Tristina1788:
> Hi @Tristina1788 , @IanMayo about shapeType_should_alway_shows_as_dropdown_list.mp4 issue, I can explain a bit of logic here. When creating a new feature, we assign a **shapeType** in the feature properties and save it to the db (line 219) image So when adding a ShapeType filter, we load the dropdown options by the **shapeType** feature property (line 411) image
> 
> In your case, the original features don't have a **shapeType** in the property, so that's why it's showing Input (choices.length === 0) image
> 
> So I think Ian can update the shapeType property for the original features in the db, then will it be fixed.

@lebaphi   Thanks for explaination. @IanMayo Could you help to update shapeType property for the original features in the db? So I can check again.

IanMayo:
Hello @philb89 - I've just jumped in to fix the missing data.  

But, I realise we don't need a new attribute. The shape-type is part of the GeoJSON standard, it's in the Geometry:
image

I realise that our other filters only apply to `properties` - so I guess we'll need to move it in/out of the properties dictionary when we're changing the selected feature.

I know it's a pain supporting something that isn't in the properties @philb89 - but it's of great benefit that we're providing search for an attribute that should always be present in all GeoJSON data structures.

Tristina1788:
@lebaphi @IanMayo This branch still have some conflict. Could you help to resolve it first?

IanMayo:
Hello @lebaphi  - I see there is some outstanding work in supporting the GeoJSON `geometry.type` field, please.

lebaphi:
okay @IanMayo, I will take a look

IanMayo:
Thanks for the update @lebaphi .

@Tristina1788 - do you understand the requirement here, are you able to test?  (I can give more explanation if necessary)

Tristina1788:
> Thanks for the update @lebaphi .
> 
> @Tristina1788 - do you understand the requirement here, are you able to test? (I can give more explanation if necessary)

Thanks @IanMayo , I can test it. Now we have filter 'GeoType' with 3 values (Polygon and Point, LineString) and we can filter it as expected


https://github.com/serge-web/serge/assets/107697044/6cf935bc-151b-4131-a1ed-67dff75a0112

Tristina1788 (dismissed):
It works as expected.

IanMayo:
Thanks @lebaphi 

But, in the property selector dialog and the filters panel,  could we call it `Shape`, instead of `GeoType` please?

IanMayo (approved):
Working perfectly thanks @lebaphi

PR #2911, opened 2024-04-10
Should not allow user update 'Health' with invalid value

### LINKED ISSUE #2909: Should not allow user update 'Health' with invalid value
## 🐞 Bug description
Should not allow user update 'Health' with invalid value

From @IanMayo .  Ok, @lebaphi the `health` property is defined in the database as a `NumberProperty` property - so when we have this type of property, we should be able to verify it is a number - please.

## 🔢 Steps to Reproduce
2. Click on 'Core Mapping' channel
3. Click on MilSymbol item
4. Update 'Health' with invalid value (example 'abc')
5. Notice I can save for this properties and health bar is not showed

image

## ⚙️ Expected behavior
Should not allow user update 'Health' with invalid value

## 💻 System information
- Browser : Chrome

lebaphi:
Fixes #2909


Tristina1788:
@lebaphi  I can input the number < 0 and > 100 . I think we should not.  @IanMayo How do you think?

image

image

IanMayo:
> @lebaphi I can input the number < 0 and > 100 . I think we should not. @IanMayo How do you think?
> 
Nice observation @Tristina1788 , but we don't actually specify limits for number-properties - I just included it in the description as a hint. We _may_ extend to include that in the future, but it's not a current requirement.

Here is what we specify:
image

Tristina1788:
> > @lebaphi I can input the number < 0 and > 100 . I think we should not. @IanMayo How do you think?
> 
> Nice observation @Tristina1788 , but we don't actually specify limits for number-properties - I just included it in the description as a hint. We _may_ extend to include that in the future, but it's not a current requirement.
> 
> Here is what we specify: image

Now I can see the hint for 'Health' as expected.

image

PR #2933, opened 2024-05-09
Define mapping participants, include sample data

### LINKED ISSUE #2931: Design mapping permissions model
We (Ian) will need to design the permissions data model for editing mapping data. Roles will have some combination of [Add/Delete], [Move/Resize], [Edit all props], [Edit user editable props]. Umpire permissions apply to shapes of all forces, other forces only control their own.

IanMayo:
Fixes #2931 

**Note:** still need to change the permissions into per-force


Tristina1788:
@IanMayo  I see some shape has force 'f-Red' but the color is pink. I think it's better if it has red color

image


- When I update propertiy 'Force' of the shape to 'f-green', the color of shape is pink. => It should green color.

image

IanMayo:
> @IanMayo I see some shape has force 'f-Red' but the color is pink. I think it's better if it has red color

Thanks @Tristina1788 - but that doesn't seem related to this PR.  Could you create it as a fresh issue please?

PR #2897, opened 2024-03-06
2892_Edit_templates_in_admin_pages

lilitkarapetyan:
**NOTE:** Converted to draft, since we can only merge it once we've fully moved to the new template model.

Fixes #2892 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations

- [ ] I have chosen reviewers for my PR.
- [ ] I have assigned myself to this PR.
- [ ] I have chosen an appropriate label for the PR.
- [ ] I have completed the mandatory sections of this document.
- [ ] I have deleted any unused sections.
- [ ] I confirm that I have checked for required README updates and acted accordingly.

PR #2964, opened 2024-05-30
update automation to work with develop branch and template branch

Tristina1788:
Currently PR #2897 is using another template with develop code. So we need to update automation for passing on both branch.


Tristina1788:
I can't revert DB file although I don't change it. So close this PR.

PR #2965, opened 2024-05-30
Update test automation for working template and develop

Tristina1788:
Currently PR https://github.com/serge-web/serge/pull/2897 is using another template with develop code. So we need to update automation for passing on both branch.


IanMayo (approved):
Great that tests now work :-D

PR #2981, opened 2024-06-12
Re-use string consts, to ensure consistent case in phase naming

### LINKED ISSUE #2980: Inconsistent naming of phases
We have string consts to represent phase, and we have enumerations to represent phase.

These mistakenly use different capitalisation.

Consistently use the lower case version, fixing the case for neat rendering as appropriate.

IanMayo:
Fixes #2980


IanMayo:
Close. There wasn't actually an issue.

Issue #2979, opened 2024-06-12
Control which phases a collab participant can interact


IanMayo:
For most of the time, the collab channel participation will be during the Planning Phase.

Where collab channel messages are generated by adjudications  during the adjudication phase, we should be able to prevent players from editing/interacting with them.

Introduce new attribute to the Collab Participant record type, which lists the phases that this participant has the `action` buttons visible.  For backwards compatibility (and ease of use), if the fields is empty then the participant can interact in all phases.

* [x] Update data definitions
* [x] Reflect data definitions in Collab Channel UI for player
* [x] Verify implementation via StoryBook
* [x] Update admin page to reflect new attribute

**Update:** ok, I've done the early steps for this.  We now just need the  `admin` page for `collab channels` to allow zero or more phases to be specified for an participant.

Here is the current UI. This will require a new multi-select drop-down containing `planning` and `adjudication` (to the right of `See live updates` .  Zero or more can be selected.
![image](https://github.com/serge-web/serge/assets/1108513/19096802-9614-4218-93e9-a7446ca3357a)

Here is the data change that this control relates to:
![image](https://github.com/serge-web/serge/assets/1108513/cc249a86-4557-49b0-80e7-36f5250fc51f)

The PR for this issue contains a branch that can be used to complete development.

lebaphi:
should we close it @IanMayo ?

IanMayo:
Now complete.

PR #2982, opened 2024-06-12
2979 collab channel phases

IanMayo:
Supports #2979


IanMayo:
Hello @Tristina1788 - this PR won't be ready for review until the admin pages have been updated. Then, you should be able to check Serge matches the behaviour specified in the issue.

IanMayo:
It looks good to me, I look forward to @Tristina1788 verifying that when phases are specified, they are reflected in which controls are offered to players in which phases.

Tristina1788:
@IanMayo  @lebaphi The header should be left of column (as 'Force' header).

image


image

- User still can't interact message  when user have permission on specific phase. The user only can interract message when has permission for 'All Phase' => User should have permission on any define phase.

Ex : setting collab Channel with user Blue - 3rd Party have permission on 'Planning' phase. But playing game with this user, no action this user can do for message : https://app.screencast.com/fYQ8u101YL6lU

Tristina1788:
@IanMayo @lebaphi 

- Now the header shows as expected

image

- 'User still can interact message when user have permission on specific phase as expected.

Everything is fine now. I think we can close it.

PR #2983, opened 2024-06-12
As a player I wish the map to pan to a particular feature

### LINKED ISSUE #2977: As a player I wish the map to pan to a particular feature
In PFT-9, when we selected an asset the map panned to that asset.

I'd like to try something different in our new implementation.

I'd like to add a `pan` button to the features list in the Core Mapping panel.  For each feature we will show an icon, that, when clicked pans the map to either the location of a point feature, or the centre of a shape feature.  The icon should be `magnifying-glass-location`:
image

The icon will sit immediately to the left of the feature icon.
image

**Update**: I've implemented most of this, the only part I couldn't handle was the actual `map.flyTo` operation.

@lebaphi - could you finish that off for me please?

We just need some map magic around this line:
https://github.com/serge-web/serge/blob/7d9d6928b02a9c607b751469fb9a6d702ce0735c/client/src/Components/local/atoms/core-mapping/index.tsx#L48

Note: when I was experimenting, it looked like `useMap` needed to be in a decendent of `MapContainer`.  _Maybe_ `MappingControls` meets that requirement, I'm not sure.

Note 2: I mistakenly made the above changes in _develop_, but please make your chances in a new branch for this issue.

Note 3:  I've put the new clickable icon in the UI, but it's centered :-( . It would be better if it was moved to the right, alongside the icon.
image

lebaphi:
Fixes #2977


IanMayo (approved):
That's great, thanks @lebaphi

PR #2997, opened 2024-07-16
Separate MapControls AddRemove property into Add and Remove

### LINKED ISSUE #2996: Separate MapControls `AddRemove` property into `Add` and `Remove`
We currently manage which controls to provide the player depending on the value of `canAddRemove`.

But, the value of this property is updated when an item is selected.

That logic is slightly wrong. The ability to `remove` is sensitive to the selected item.  But, the ability to `add` should not be.  That is probably why @Tristina1788 is finding she cannot add new `green` items, [here](https://github.com/serge-web/serge/pull/2970#issuecomment-2230622794)

So, we should separate the `MapControls` prop titled `canAddRemove` into `canAdd` and `canRemove`.  The first just checks if this player has permission to `AddRemove` for their own force.  The second is updated on a new `selectedFeature` - as in the existing logic.

This work should be done in a branch off `core-mapping-admin`

lebaphi:
Fixes #2996

### DIFF
--- a/client/src/Components/local/atoms/mapping-panel/helpers/has-mapping-permission.tsx
+++ b/client/src/Components/local/atoms/mapping-panel/helpers/has-mapping-permission.tsx
@@ -44,4 +44,4 @@ export const canAddRemove = (feature: Feature<Geometry, any>, permissions: Parti
   return hasMappingPermissions(feature, [MappingPermissions.AddRemove], permissions)
 }
 
-export const permissionError = (): void => alert('You do not have permission on this action')
+export const permissionError = (): void => console.log('You do not have permission on this action')
--- a/client/src/Components/local/atoms/mapping-panel/helpers/has-mapping-permission.tsx
+++ b/client/src/Components/local/atoms/mapping-panel/helpers/has-mapping-permission.tsx
@@ -43,3 +43,5 @@ export const canMoveResize = (feature: Feature<Geometry, any>, permissions: Part
 export const canAddRemove = (feature: Feature<Geometry, any>, permissions: ParticipantMapping[]): boolean => {
   return hasMappingPermissions(feature, [MappingPermissions.AddRemove], permissions)
 }
+
+export const permissionError = (): void => alert('You do not have permission on this action')
--- a/client/src/Components/local/atoms/core-mapping/helper/map-controls.tsx
+++ b/client/src/Components/local/atoms/core-mapping/helper/map-controls.tsx
@@ -114,8 +114,10 @@ const MapControls: React.FC<GeomanControlProps> = ({ onCreate, onShowLabels, can
     })
     
     // handle onCreate listener for geoman items
-    map.on('pm:create', (e: LeafletEvent) => {
-      switch (e['shape']) {
+    // note: the shape string parameter was added to the event object to help determine the type of shape being created,
+    // and avoid a compiler issue.
+    map.on('pm:create', (e: LeafletEvent & { shape?: string }) => {
+      switch (e.shape) {
         case 'Marker':
         case 'Polygon':
         case 'Line':
@@ -124,7 +126,7 @@ const MapControls: React.FC<GeomanControlProps> = ({ onCreate, onShowLabels, can
           onCreate(e as unknown as PM.ChangeEventHandler)
           break
         default:
-          console.log('OnCreate Unimplemented !!!', e['shape'])
+          console.log('OnCreate Unimplemented !!!', e.shape)
       }
     })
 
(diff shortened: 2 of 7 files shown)


IanMayo:
Thanks @lebaphi 

I see you haven't had to split `canAddRemove` into `canAdd` and `canRemove`.  That's fine, as long as it meets the UI requirement :-)

lebaphi:
> Thanks @lebaphi
> 
> I see you haven't had to split `canAddRemove` into `canAdd` and `canRemove`. That's fine, as long as it meets the UI requirement :-)

Yep @IanMayo, I tried with a different approach, you are welcome to test this 🤣

client/src/Components/local/atoms/mapping-panel/helpers/has-mapping-permission.tsx
```diff
@@ -43,3 +43,5 @@ export const canMoveResize = (feature: Feature<Geometry, any>, permissions: Part
 export const canAddRemove = (feature: Feature<Geometry, any>, permissions: ParticipantMapping[]): boolean => {
   return hasMappingPermissions(feature, [MappingPermissions.AddRemove], permissions)
 }
+
+export const permissionError = (): void => alert('You do not have permission on this action')
```
IanMayo: Hello @lebaphi - right now, I think this can just be a message to the console.  I can't quite think of a tidy UI for when the user tries to drag something they don't have permission for.

Tristina1788:
@IanMayo @lebaphi I can see the log for user don't have permission on item

image

- And I also see screen is blink when move some items, some items don't make screen blink (Move black shapes make screen blink, but red shapes don't)

https://github.com/user-attachments/assets/ee5fda03-f7aa-4a53-a491-d9279cadad23

- Should not show 'Cancel' button because user can't update properties so 'Cancel' button is not useful.

image

- The coremapping tool is lost after try to remove item which the user don't have permission. => This way will make user confuse.


https://github.com/user-attachments/assets/b084fdc0-ecda-4c03-abbb-f088d450f36c

lebaphi:
@Tristina1788 
> And I also see screen is blink when move some items, some items don't make screen blink (Move black shapes make screen blink, but red shapes don't)

It's because the map is re-render after changes, I will check if I can improve that

> Should not show 'Cancel' button because user can't update properties so 'Cancel' button is not useful.

I think it's still useful if the user want to unselect the current selected item instead of clicking on the map.

> The coremapping tool is lost after try to remove item which the user don't have permission. => This way will make user confuse.

That's because you don't have the "Remove" permission on Red, so when you click on Red => the toolbar updates its status.

Tristina1788:
@lebaphi @IanMayo 
`It's because the map is re-render after changes, I will check if I can improve that` : Now I still don't see the improvement on 'screen is blink when move some items'

`I think it's still useful if the user want to unselect the current selected item instead of clicking on the map.` : So It's ok for this button.

`That's because you don't have the "Remove" permission on Red, so when you click on Red => the toolbar updates its status.` : Yes, I understand that, but I'm afraid that user which play game and he's not an admin, so it's quite confuse when some button is lost when click some items. I think we can add more steps on the guide for tutorial, so user can know the way it works. How do you think? @IanMayo

lebaphi:
> It's because the map is re-render after changes, I will check if I can improve that : Now I still don't see the improvement on 'screen is blink when move some items'

Yep, I will try to reduce the rendering fever as much as possible, but the flickering phenomenon is due to React's behavior, it is what it is, I cannot interfere.

lebaphi:
If there are no more issues then we can merge this PR? @Tristina1788 @IanMayo

IanMayo (approved):

Tristina1788:
> Yep, I will try to reduce the rendering fever as much as possible, but the flickering phenomenon is due to React's behavior, it is what it is, I cannot interfere.

If the reason is React's behavior, why the performance is different from items. (Some items are blink when moving, some items are not) . @IanMayo How do you think?

https://github.com/user-attachments/assets/02ef082c-cfb2-41e4-bc41-bcad7f2be349

lebaphi:
I think it's simple @Tristina1788 
- If you have permission to drag the item => it will update the new position after dragging => re-render => flicker
- If you don't have the permission => you can still drag item on the map it but it does nothing

Looks like your browser is caching, can you check again in an incognito browser?

Tristina1788:
> I think it's simple @Tristina1788
> 
> * If you have permission to drag the item => it will update the new position after dragging => re-render => flicker
> * If you don't have the permission => you can still drag item on the map it but it does nothing
> 
> Looks like your browser is caching, can you check again in an incognito browser?

 Try on browser no cache, and seem the shape which user don't have permission move,  the item will not blink (maybe because it doesn't need to update the new position). So it's ok.

PR #2998, opened 2024-07-16
2845 pan to selection

### LINKED ISSUE #2845: Mapping - pan to selected feature
When we select an item in the left-hand mapping bar, we should pan the map so that the feature is in the centre.

**Improvement:** when an item is double-clicked in the mapping bar pan to it.  Don't pan if it is just being selected.

Since the `onselect` method in the mapping component only receives and id (and not the element itself) I _guess_ we could add an optional `point` to the `onSelect` method in the Mapping component.

The left-hand bar would create a shape for the feature, ask `turf` to provide the centre of that shape, and pass that.

IanMayo:
Fixes #2845


IanMayo:
Drop this, it was already fixed in #2983

